PR #14, opened 2019-08-10
Docker introduction

kelson42:


docker/README
```diff
@@ -0,0 +1,7 @@
+# Wikipedia 1.0 bot
+
+WP1 bot has been dockerized in a few images:
+- "update" image is responsible for gathering the WP1 evaluation data to the DB.
+- "upload" image is responsible for update Wikipedia pages to
```
audiodude: There is no separate upload image. When a project is enqueued, it will be for both update and upload.

PR #31, opened 2019-08-14
Undo the changes in PR #29

audiodude:


audiodude:
This change undoes opening the database connection prefork, because pymysql database connections are not thread safe and cannot be shared. Undoes #29.

PR #38, opened 2019-08-14
File hierarchy reorganisation

kelson42:


audiodude:
You need to change the path `lucky/requirements.txt` to just `requirements.txt` in the travis.yml

PR #39, opened 2019-08-14
Tests for the tables.py file

audiodude:
This is the beginning of the unit tests for tables.py. There are still several tests that need to be added.


audiodude:
Maybe I will merge this once I have better coverage for tables.py

PR #41, opened 2019-08-15
Small code style improvements

kelson42:


wp1/logic/project_test.py
```diff
@@ -22,10 +22,7 @@ def _get_first_category(wp10db):
   with wp10db.cursor() as cursor:
     cursor.execute('SELECT * FROM ' + Category.table_name + ' LIMIT 1')
     db_category = cursor.fetchone()
-    if db_category:
-      return Category(**db_category)
-    else:
-      return None
+    return db_category ? Category(**db_category) : None
```
audiodude: There is no such operator in python. You might be thinking of:

`return Category(**db_category) if db_category else None`

kelson42: @audiodude thx for the hint

wp1/logic/api/page.py
```diff
@@ -25,10 +25,9 @@ def get_redirect(title_with_ns):
     logger.warn('Error contacting API, returning None')
     return None
 
-  if 'query' not in res or 'redirects' not in res['query']:
-    return None
-  redir = res['query']['redirects'][0]['to']
-  if len(redir) == 0:
+  if ('query' not in res or
+      'redirects' not in res['query'] or
+      not res['query']['redirects'][0]['to']):
```
audiodude: len(res[...]) > 0

instead of not res[...]

Issue #23, opened 2019-08-13
Consider opening db connection and configuring logging outside of worker function


audiodude:
We might use https://python-rq.org/docs/workers/#custom-worker-classes to do this.

audiodude:
Logging into the API too frequently was a definite problem, which was addressed in https://github.com/openzim/wikimedia_wp1_bot/commit/042022cce8ceecae7f303a39175b22a28485777a with a custom worker script. This script could also configure logging and get a db connection.

However, the db connection would have to be made into a global object again, as opposed to using the _connect() functions which were introduced recently.

audiodude:
Not sure why this was assigned to me, I closed it last night: won't fix.

PR #80, opened 2019-08-24
New strategy for dealing with failed page edits

### LINKED ISSUE #69: API login retry logic not working
The retry logic that was added in https://github.com/openzim/wikimedia_wp1_bot/commit/d5769dd3c25888cd8a42c4ef3c8e4786c932bb2d is not working.

Here is the relevant snippet from the logs:
```
INFO:wp1.tables:Uploading wikicode to Wikipedia: Adele
WARNING:wp1.api:Got login exception, retrying login
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): en.wikipedia.org:443
DEBUG:urllib3.connectionpool:https://en.wikipedia.org:443 "GET /w/api.php?meta=siteinfo%7Cuserinfo%7Cuserinfo&siprop=general%7Cnamespaces&uiprop=groups%7Crights%7Cblockinfo%7Chasmsg&continue=&action=query&format=json HTTP/1.1" 200 2220
DEBUG:urllib3.connectionpool:https://en.wikipedia.org:443 "GET /w/api.php?meta=tokens&type=login&action=query&format=json HTTP/1.1" 200 110
DEBUG:urllib3.connectionpool:https://en.wikipedia.org:443 "POST /w/api.php HTTP/1.1" 200 88
DEBUG:urllib3.connectionpool:https://en.wikipedia.org:443 "GET /w/api.php?meta=userinfo%7Cuserinfo&uiprop=groups%7Crights%7Cblockinfo%7Chasmsg&continue=&action=query&format=json HTTP/1.1" 200 280
00:00:24 mwclient.errors.AssertUserFailedError: By default, mwclient protects you from accidentally editing without being logged in. If you actually want to edit without logging in, you can set force_login on the Site object to False.
Traceback (most recent call last):
  File "/usr/src/app/wp1/api.py", line 48, in save_page
    page.save(wikicode, msg)
  File "/usr/local/lib/python3.7/site-packages/mwclient/page.py", line 182, in save
    raise mwclient.errors.AssertUserFailedError()
mwclient.errors.AssertUserFailedError: By default, mwclient protects you from accidentally editing without being logged in. If you actually want to edit without logging in, you can set force_login on the Site object to False.
```

My current best guess, from looking at the mwclient code, is that the `page` object keeps a reference to the site it was pulled from, so that logging in again on the global site object perhaps doesn't do anything.

The other problem is that this is hard to test on Wikipedia, since the login time is presumably pretty long.

audiodude:
As stated in the comment in the code, since the site is tied to the page itself, get a new page object from the API when we re-login. Fixes #69.

### DIFF
--- a/wp1/api.py
+++ b/wp1/api.py
@@ -45,11 +45,12 @@ def save_page(page, wikicode, msg):
   try:
     page.save(wikicode, msg)
   except mwclient.errors.AssertUserFailedError as e:
-    logger.warning('Got login exception, retrying login')
+    logger.warning('Got login exception, creating new site object')
     login()
-    # Retrieve a token from the api, which will cause the userinfo to refresh and site.logged_in
-    # to have the correct value. This is a necessary workaround for this bug:
+
+    # Get a new copy of the page, since page.save is tied to page.site and that refers to the old
+    # site from before we did the re-login. All of this is a workaround for:
     # https://github.com/mwclient/mwclient/issues/231
-    page.get_token('edit', force=True)
+    page = get_page(page.name)
     page.save(wikicode, msg)
   return True
--- a/wp1/api_test.py
+++ b/wp1/api_test.py
@@ -39,12 +39,13 @@ class ApiTest(unittest.TestCase):
 
   @patch('wp1.api.site')
   @patch('wp1.api.login')
-  def test_save_page_tries_login_on_exception(self, patched_login, patched_site):
+  @patch('wp1.api.get_page')
+  def test_save_page_retries_on_exception(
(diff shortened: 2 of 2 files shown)


audiodude:
The problem with this approach is that although the workers logged in to Wikipedia "pre fork" and shared a single login (good, see #23), when they retry and clear the site object, that only happens on a per job basis.

So once the global site object gets logged out, the individual re-logins will happen in each worker job, which will likely result in the "too many login requests" rate limiting we originally saw.

A better idea might be #81.

audiodude:
I still think this is a good idea though, as a last ditch effort rather than failing completely.

PR #85, opened 2019-08-24
Format all code in the wp1 directory with yapf

audiodude:
See https://github.com/google/yapf

I've configured it slightly for Google style (80 width) but with 2 spaces indent, which helps save width.

@kelson42 let me know what you think! We can add a pre-commit git hook for this if we like it. Personally, I think it's pretty good, except for the SQL queries, which probably need some love anyways.

### DIFF
--- a/.style.yapf
+++ b/.style.yapf
@@ -1,0 +1,3 @@
+[style]
+based_on_style = google
+indent_width = 2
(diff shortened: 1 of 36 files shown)


audiodude:
Okay apparently the pre-commit hook has to be installed per repository, but it's available here: https://raw.githubusercontent.com/google/yapf/master/plugins/pre-commit.sh

PR #87, opened 2019-08-25
Create a config for the 'pre-commit' program to help keep pre-commits straight

audiodude:
We should document this in README.md, but the basic gist is:

1. Install the requirements in your virtual env and you should get the `pre-commit` binary
1. Run `pre-commit install` to install the hook in your repo
1. When you try to commit with changes that would be modified by the pre-commit hook you get:

```
(venv) tbriggs-imac:wikimedia_wp1_bot tmoney$ git commit -am 'Test commit'
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
yapf.....................................................................Failed
hookid: yapf
```

And you have to look at the files and then re-add them, then re-commit.


audiodude:
I've added the required information to README.md.

Issue #89, opened 2019-08-26
Logs for some WikiProjects show articles getting reassessed every day


audiodude:
https://en.wikipedia.org/wiki/Wikipedia:Version_1.0_Editorial_Team/American_animation_articles_by_quality_log

The log shows the same articles, which appear to probably be every article in the project, being reassessed every day. Investigate this.

audiodude:
https://en.wikipedia.org/wiki/Wikipedia:Version_1.0_Editorial_Team/American_Civil_War_articles_by_quality_log

American Civil War articles seem to have the same problem.

audiodude:
Similar problem here:

https://en.wikipedia.org/wiki/Wikipedia:Version_1.0_Editorial_Team/Adelaide_articles_by_quality_log

And this is just the A's. Looks pretty widespread.

kelson42:
Another one https://en.wikipedia.org/w/index.php?title=Wikipedia:Version_1.0_Editorial_Team/Saints_articles_by_quality_log&oldid=913442816

audiodude:
@walkerma if you look at the first example of this I showed, it's actually a problem with Unknown-Class it seems: https://en.wikipedia.org/wiki/Wikipedia:Version_1.0_Editorial_Team/American_animation_articles_by_quality_log

This example also looks like a lot of problems with Unknown-Class:
https://en.wikipedia.org/wiki/Wikipedia:Version_1.0_Editorial_Team/Adelaide_articles_by_quality_log

kelson42:
Should be fixed now

PR #103, opened 2019-09-06
Avoid recording duplicate assessments

audiodude:
This PR refactors the way article assessments are inserted/updated into the database. Instead of inserting them as they are found, a list of ratings for an article is made and then later that list of ratings is sorted based on the "score" value for that rating. This means that if an article is rated both Start-Class and A-Class, only the A-Class is taken and recorded.

By deferring the decision to update a rating until all "candidate" ratings have been collected, we also avoid rating thrashing, where logs for ratings are recorded with the article going back and forth between ratings, as seen in #89.


audiodude:
@kelson42 Can you take a look at this and make sure it seems sane?

kelson42:
@audiodude I'm sorry, I don't have really enough understanding over the whole process, but everything I have read seems to make sense to me.

PR #110, opened 2019-09-11
FIrst "hello world" version of web app

audiodude:
Replaces the rq dashboard at "/" with a hello page that simply lists the number of projects that we are serving. However, this serves as a good test for database connectivity, Flask being installed correctly, etc.

The rq dashboard is now on "/rq", with the same credentials as before, which are now taken from the environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,3 +12,4 @@ after_success:
   - codecov
 services:
   - mysql
+  - redis-server
--- a/docker/web/start.sh
+++ b/docker/web/start.sh
@@ -3,4 +3,4 @@ export RQ_USER=$(head -n 1 app/rq-credentials.txt)
 export RQ_PASS=$(tail -n 1 app/rq-credentials.txt)
 
 cd app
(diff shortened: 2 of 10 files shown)


audiodude:
I'm having trouble testing the Flask app, because I can't get it to properly connect to the test database. See SO question here: https://stackoverflow.com/questions/57897685/how-do-i-get-flask-to-connect-to-my-test-database-if-im-storing-the-db-connecti

kelson42:
@rgaudin Maybe you could have a quick look?

audiodude:
Added working tests! The key was to not pass in the same connection to the web app that we use to setup/teardown the db, but to use different connections for that.

kelson42:
@audiodude Looks like we are hitting here a UI testing problematic. Code coverage decreases.

audiodude:
Yes, it's not testing the branch where the RQ creds are provided and it sets up the RQ dashboard. I'll add tests for that. (FWIW I tried to see the uncovered lines before but I was either clicking the wrong link or the report wasn't ready)

PR #125, opened 2019-11-25
New web frontend written in vue.js

audiodude:
This is the first 'hello world' version of a web frontend written in Vue.

The current implementation is identical to what is running at https://wp1.openzim.org, except that the data is loaded asynchronously from the api (with the current web service/api being renamed to api.wp1.openzim.org)

Included as well is a docker file for the frontend, which simply installs Node dependencies and builds a static site, then serves it with nginx.


audiodude:
It's still just a placeholder API client, which only does one call, to get the project count. As mentioned, when this gets deployed the site will look basically the same as it does today except that the project count will be retrieved asynchronously from the API.

I just wanted to make sure that the infrastructure could support the basic deployment model that I had in mind. Thanks!

Issue #126, opened 2019-11-28
wp1.openzim.org does not use HTTP compression


kelson42:
* https://wp1.openzim.org https://www.whatsmyip.org/http-compression-test/?url=aHR0cHM6Ly93cDEub3BlbnppbS5vcmc=
* https://api.wp1.openzim.org https://www.whatsmyip.org/http-compression-test/?url=aHR0cHM6Ly9hcGkud3AxLm9wZW56aW0ub3Jn

kelson42:
See also https://github.com/openzim/zimfarm/issues/332 (for the solution)

audiodude:
Check the URLs again:

https://www.whatsmyip.org/http-compression-test/?url=aHR0cHM6Ly93cDEub3BlbnppbS5vcmc=

https://www.whatsmyip.org/http-compression-test/?url=aHR0cHM6Ly9hcGkud3AxLm9wZW56aW0ub3Jn

Compression now verified.

PR #128, opened 2019-11-30
Enable gzip compression for the Flask API server.

audiodude:
This includes installing the minimal flask-gzip plugin and turning it on. Additionally, some fluff has been added to the index page text so that it is large enough to qualify for gzip compression, for verification purposes.


audiodude:
This is the rest of the fix of #126

PR #131, opened 2020-01-26
Add first Vue component, with no-op Storybook test

audiodude:
This is the first attempt at adding a Vue component. It is the Autocomplete component, which offers suggestions of project names based on what is typed in an input box.

In a follow up, the project that is selected can be used to navigate to a page that shows a summary of the project. Or this component could be used in searching for articles, instead of the error-prone full text matching that is done on the old website.

This PR also includes adding [Prettier](https://prettier.io/) formatting to pre-commit, which is the HTML/JS/Vue equivalent of yapf (the Python formatting tool we use).

Unfortunately, I had some trouble with Storybook. While I could get the component to render in its basic state, I couldn't figure out how to add keystrokes to it so that it would display the various autocomplete states.

TODO: Add instructions on how to run a local API server, a local frontend, and the storybook server.


kelson42:
@audiodude Great news. @rgaudin has just finished the version 2.0 of the zimfarm which - like you know - use the same stack as the wp1 engine. I would like to ask him to review this PR, he is far more skilled for that.

audiodude:
Thanks for the review!

I've filed #133 to capture updating the READMEs all around.

PR #142, opened 2020-03-05
Create a docker image and docker compose for a database for development

audiodude:
Create a dev database with 30 MB of sample data (included in the repo). This can be loaded as a docker image using the new docker-compose-dev.yml, then used to test data collection (when connected to the enwiki_p replica database through an SSH tunnel) or used to power a dev version of the web frontend.


kelson42:
@audiodude Would that be possible to have this 30MB of data outside the git repo? It is really big! We could host it on download.openzim.org for example.

audiodude:
This branch got messed up, reverting.

PR #143, opened 2020-03-05
Create a docker image for a dev database

audiodude:
Create a dev database with 30 MB of sample data (included in the repo). This can be loaded as a docker image using the new docker-compose-dev.yml, then used to test data collection (when connected to the enwiki_p replica database through an SSH tunnel) or used to power a dev version of the web frontend.

### DIFF
--- a/dev-database.Dockerfile
+++ b/dev-database.Dockerfile
@@ -1,0 +1,4 @@
+FROM mariadb:10.1
+
+ENV MYSQL_ROOT_PASSWORD wikipedia
+COPY enwp10_dev.sample.sql /docker-entrypoint-initdb.d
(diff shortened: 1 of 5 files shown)


audiodude:
@kelson42 What do you think of adding 30MB of sample data to the repo, so that we can have a dev database for developing the frontend? This is what I have been working on.

I also considered making the dev database it's own repo, which we could do, but it might be harder to remember where it lives and update it when necessary.

audiodude:
From @kelson42 's comment elsewhere:

> @audiodude Would that be possible to have this 30MB of data outside the git repo? It is really big! We could host it on download.openzim.org for example.

Yes I think I would lean towards just having a separate repo for creating the docker image for the dev database. It doesn't actually have any dependencies on this code.

audiodude:
Created https://github.com/openzim/wp1-dev-database

audiodude:
Also created https://hub.docker.com/repository/docker/openzim/wp1-dev-database

audiodude:
Okay, the other changes in this PR can be addressed when we start integrating the dev database. Closing.

Issue #169, opened 2020-05-14
Upload links in wiki code to point to wp1.openzim.org


kelson42:
An stop pointing to the old WP1 engine. This is a measure to prepare the phaseout of old Perl based WP1 engine.

audiodude:
This is already done.

PR #191, opened 2020-06-08
Link to v2 of the Web frontend in project tables

audiodude:
This is optionally controlled by a constant, link_to_v2, which means that the old linking behavior is preserved and can be turned back on at any time.


audiodude:
Fixes #169

Issue #186, opened 2020-06-07
Don't allow user to start a new manual update if one is in progress, even if it's been over an hour


audiodude:
Currently there is a "lock" on manual updates, which is the next_update_time, which is set to 60 minutes after the initial update is requested. However, if the rq job is still being processed at that time, we should not allow a new manual update because it will just queue up behind the currently executing one.

audiodude:
Oops, wrong issue.

mahakporwal02:
@audiodude Can I work on this issue?

audiodude:
Yes, thank you!

PR #192, opened 2020-06-08
Guard against pages loading with non-existant projects and other frontend cleanups

audiodude:
Fixes some alignment issues with paragraphs on a few pages. Also removes the Autocomplete box from the Article listing page to make it less busy.

If the main page is loaded with a non-existant project: (eg [http://wp1.openzim.org/#/project/Paris](http://wp1.openzim.org/#/project/Paris)), the page is simply loaded in it's default, empty state.

The same is true of the manual update page (eg [http://wp1.openzim.org/#/update/Paris](http://wp1.openzim.org/#/update/Paris)), the page simply loads as if no Project was chosen.

These two default states make sense, since the only way the page can be reached with an invalid project is through URL hacking anyways.

The article listing page also refuses to list the results for non-existant projects, and this time displays a "not found" message, in order to distinguish between existing projects with no results and non-existant projects: (see [http://wp1.openzim.org/#/project/Paris/articles?quality=Start-Class&importance=Mid-Class](http://wp1.openzim.org/#/project/Paris/articles?quality=Start-Class&importance=Mid-Class))

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -28,6 +28,8 @@ paths:
                 type: array
                 items:
                   $ref: '#/components/schemas/Project'
+        404:
+          description: 'No project with that projectId was found'
   /projects/count:
     get:
       summary: 'Count of the total number of projects'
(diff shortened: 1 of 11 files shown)


audiodude:
Fixes #186

PR #242, opened 2020-10-01
Add Cypress end-to-end tests for frontend

audiodude:
So far there is only two tests: That the page loads, and that you can autocomplete search for 'Water'.

I want to see if this works in Travis CI, so I've modified the .travis.yml file as well. It probably needs a change to the environment variables to add a "TEST" environment, so that the frontend knows where to look for the backend.


audiodude:
I have no idea how to resolve those CodeFactor issues, unfortunately. Merging anyway.

Issue #259, opened 2020-12-06
Create a way for conf.json to be variable


audiodude:
Although the file conf.json in theory provides a way to make the labels and categories that wp1 bot uses variable, in practice there is a single copy of the file that is checked into the repo.

We need to possibly make a `conf/` directory with `en/`, `ar/`, etc subdirectories, each with a conf.json file in them. Then we can map the proper directory at Docker Compose time based on which language is being deployed.

PR #268, opened 2020-12-06
Move conf to language specific sub-directory

audiodude:
The file conf.json specifies which identifiers are used for parsing out "by quality" and "by importance" categories for Wikipedia. It also specifies the labels for quality and importance and their default weights. This file exists solely to allow for localization of the project, but until now it was hardcoded in the top level directory.

Now there are various language specific subdirectories for conf.json files, with the language code being specified in credentials.py (which is already necessarily different between different language instances of the project).


audiodude:
Fixes #259

Issue #271, opened 2020-12-09
Setup around configuration


kelson42:
A few questions/remarks to secure we don't leak the configuration:
* The configuration file template should be different from the real configuration
* The configuration file(s) should be in the `.gitignore` (with all the files which might appear for temporary saving, like the one with `~` for emacs)
* Secure the file is outside the served directory.

audiodude:
> The configuration file template should be different from the real configuration  ✅

> The configuration file(s) should be in the .gitignore (with all the files which might appear for temporary saving, like the one with ~ for emacs)  ✅

> Secure the file is outside the served directory.  ✅

PR #276, opened 2021-01-22
Update .gitignore for alternate credentials file

audiodude:
Sometimes I need to run the tool with real-ish credentials, and sometimes (for nosetests) I need to run it with no credentials. The credential leak we experienced in December was because I moved my credentials "temporarily" to credentials.py.bak, but then ended up accidentally committing this file.

I have added that path to the .gitignore, so this should help in the future.


audiodude:
Fixes #271

PR #283, opened 2021-03-16
new manual update not starting until previous one in progress

### LINKED ISSUE #186: Don't allow user to start a new manual update if one is in progress, even if it's been over an hour

mahakporwal02:
Button does not show up on the page until the previous one is in progress even after completing an hour.
This resolves #186

### DIFF
--- a/wp1-frontend/src/components/UpdatePage.vue
+++ b/wp1-frontend/src/components/UpdatePage.vue
@@ -11,7 +11,7 @@
     </div>
     <div class="row">
       <div class="col-xl-4">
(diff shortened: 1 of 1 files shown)


mahakporwal02:
Sry wrong code

mahakporwal02:
Sry wrong code

PR #284, opened 2021-03-17
Manual update changes

### LINKED ISSUE #186: Don't allow user to start a new manual update if one is in progress, even if it's been over an hour

mahakporwal02:
new manual update not starting until previous update in progress.
This resolves #186

### DIFF
--- a/wp1-frontend/src/components/Autocomplete.vue
+++ b/wp1-frontend/src/components/Autocomplete.vue
@@ -10,7 +10,7 @@
         the project you wish to update.
       </p>
       <p>
-        <b>Note: This tool can only perform manual updates once per hour.</b>
+        <b>Note: This tool can only perform manual updates once per hour only after current update is completed.</b>
       </p>
     </div>
     <div class="input-group">
--- a/wp1-frontend/src/components/Autocomplete.vue
+++ b/wp1-frontend/src/components/Autocomplete.vue
@@ -10,7 +10,7 @@
         the project you wish to update.
       </p>
       <p>
-        <b>Note: This tool can only perform manual updates once per hour only after current update is completed.</b>
+        <b>Note: This tool can only perform manual updates once per hour at most. It also cannot perform an update until all pending updates are complete.</b>
       </p>
     </div>
     <div class="input-group">
--- a/wp1-frontend/src/components/UpdatePage.vue
+++ b/wp1-frontend/src/components/UpdatePage.vue
@@ -30,17 +30,12 @@
     <div class="row">
       <div class="col-xl-6">
         <div v-if="currentProject && jobScheduled && !jobComplete">
-          <p v-if="updateTime">
+          <p>
             Manual update of <b>{{ this.$route.params.projectName }}</b> has
             been scheduled. It can take anywhere from 2 - 200 minutes, depending
-            on project size. The next update can be performed at
-            <b>{{ updateTime }}</b
-            >.
-          </p>
-          <p v-else>
-            Manual update of <b>{{ this.$route.params.projectName }}</b> has
-            been scheduled. It can take anywhere from 2 - 200 minutes, depending
-            on project size. The next update can be performed after completion of the current update.
+            on project size. The next update can be performed
+            <span v-if="updateTime">at <b>{{ updateTime }}</b>.</span>
+            <span v-else>when the current update completes.</span>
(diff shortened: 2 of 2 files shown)


mahakporwal02:
@audiodude Sorry, there was issue with my local repository therefore I had to force-push the code multiple times.
I have added conditions to `manual update button` so that it would not show while the update is ongoing.
Also after one hour and when update is not yet completed, the `update time`  becomes `null` so I have added the extra paragraph.

wp1-frontend/src/components/UpdatePage.vue
```diff
@@ -29,14 +29,19 @@
...
     <div class="row">
       <div class="col-xl-6">
-        <div v-if="currentProject && updateTime">
-          <p>
+        <div v-if="currentProject && jobScheduled && !jobComplete">
+          <p v-if="updateTime">
             Manual update of <b>{{ this.$route.params.projectName }}</b> has
             been scheduled. It can take anywhere from 2 - 200 minutes, depending
             on project size. The next update can be performed at
             <b>{{ updateTime }}</b
```
audiodude, line 37: I would prefer if this was all one paragraph. Something like "The next update can be performed at {{ updateTime }} or when the current update completes, whichever is later."

wp1-frontend/src/components/UpdatePage.vue
```diff
@@ -11,7 +11,7 @@
     </div>
     <div class="row">
       <div class="col-xl-4">
-        <div v-if="currentProject && !updateTime">
+        <div v-if="currentProject && !updateTime && (!jobScheduled || jobComplete)">
```
audiodude, line 14: Doesn't the status enum contain an "in_progress" value or something similar that we can check for. This is a roundabout way of saying that.

wp1-frontend/src/components/Autocomplete.vue
```diff
@@ -10,7 +10,7 @@
         the project you wish to update.
       </p>
       <p>
-        <b>Note: This tool can only perform manual updates once per hour.</b>
+        <b>Note: This tool can only perform manual updates once per hour only after current update is completed.</b>
```
audiodude, line 13: "...once per hour at most. It also cannot perform an update until all pending updates are complete."

audiodude (changes requested):

wp1-frontend/src/components/UpdatePage.vue
mahakporwal02: If the job is not completed but update time has become null (its been over an hour) so it would show something like this:
(I have made some changes in front-end for this screenshot.)
manual update

wp1-frontend/src/components/UpdatePage.vue
mahakporwal02: @audiodude In the original code same format is used to show in_progress status please have a look at line 51.
manual update progress

audiodude: Fair enough.

wp1-frontend/src/components/UpdatePage.vue
audiodude: Right, but what I don't like is two paragraphs with almost identical text. You should either strive to have one explanation that works in both cases, or minimize your v-if block so that it covers only the part of the text that is changing.

wp1-frontend/src/components/Autocomplete.vue
audiodude: Can you change the text to what I specified?

mahakporwal02: Yes sir, already done
Thanks

wp1-frontend/src/components/UpdatePage.vue
mahakporwal02: Done, pls have a look at new changes.

audiodude (approved):
LGTM, thanks!

audiodude:
I'll try to remember to deploy this and the other change tomorrow morning.

mahakporwal02:
OK, thanks

Issue #280, opened 2021-03-14
Write more Cypress integration tests


audiodude:
Currently we have only [one integration test](https://github.com/openzim/wp1/blob/main/wp1-frontend/cypress/integration/home_page.spec.js), and it just tests that the autocomplete dropdown opens when a search is entered.

We should have more test cases for the basic usage of the app, such as:

- When the user selects an item from the dropdown, the appropriate table is displayed
- That table has the appropriate row and column labels
- Clicking a number in the table brings the user to the appropriate article detail list page

And for the [article table](http://wp1.openzim.org/#/project/Water/articles?quality=Start-Class&importance=High-Class) page:

- It displays the right name for the page that it is (ie "Water articles - High importance / Start quality")
- Custom Pagination works, with the start page and number of items being respected
- Filtering by article name works

I'm not sure we can meaningfully test the manual update page, which is the last major page in the app right now, because in the integration environment we don't have access to Redis or the worker processes.

This is not an exhaustive list and more tests might be useful or necessary. Also if you accept this issue, the tests do not (and likey *should not*) be made all in one giant PR, but rather split up.

audiodude:
Forgot to mention, but since the dev database is static and well known, we don't have to "guess" at any of this. For example, the existing integration test searches for "Water" because we know there is a WikiProject Water in the dev database. We also know the exact number of articles in every class for that project, etc.

mahakporwal02:
@audiodude @kelson42 I have made cypress for
'When the user selects an item from the dropdown, the appropriate table is displayed'
Please have a look at it so that I can make rest of the tests after your approval.

mahakporwal02:
@audiodude Are there more tests that you need me to add?

audiodude:
@mahakporwal02 Any tests that you can think of that I might have missed would be helpful.

mahakporwal02:
@audiodude I think all important tests have been covered.

audiodude:
What about tests for this widget? 
<img width="782" alt="Screen Shot 2021-03-26 at 11 09 21 AM" src="https://user-images.githubusercontent.com/57832/112674948-c5e43900-8e23-11eb-9b4a-be783aab6b82.png">

Also can we/should we test that clicking the article name brings you to the right article on Wikipedia? Or that clicking on the revision or talk links bring you to the right places?

audiodude:
We have no tests for the manual update page. Do we need some?

audiodude:
Closing, we have a good suite of first tests.

PR #286, opened 2021-03-21
cypress test for project-table

mahakporwal02:
This resolves a part of #280   
 @audiodude @kelson42 I have made cypress for
'When the user selects an item from the dropdown, the appropriate table is displayed'
Please have a look at it so that I can make rest of the tests after your approval.

And what should be done of this CodeFactor issue?

### DIFF
--- a/wp1-frontend/cypress/integration/home_page.spec.js
+++ b/wp1-frontend/cypress/integration/home_page.spec.js
@@ -16,4 +16,20 @@ describe('the home page', () => {
       .eq(1)
       .should('contain.text', 'Water');
   });
+
+  it('project-table for Alien displayed', () => {
+    cy.visit('/');
+
+    cy.get('.search').type('Alien');
+
+    cy.get('.results').should('be.visible');
+    cy.get('.results')
+      .children('li')
+      .eq(0)
+      .should('contain.text', 'Alien')
+      .click();
+
+    cy.get('table').should('be.visible')
+      .should('contain.text','Alien articles by quality and importance');
+  });
 });


wp1-frontend/cypress/integration/home_page.spec.js
```diff
@@ -16,4 +16,21 @@ describe('the home page', () => {
       .eq(1)
       .should('contain.text', 'Water');
   });
-});
+
+  it('project-table for Alien displayed', () => {
```
audiodude, line 20: I would put this test case inside of the describe block for "the home page" because it's testing something on the home page. Generally, all the tests in one spec.js file will be in a master describe block, with sub-describe blocks for individual areas.

audiodude (changes requested):
For the codefactor issues, I would try implementing this SO answer: https://stackoverflow.com/a/60154342/41060

Your test also isn't passing on CI: https://travis-ci.com/github/openzim/wp1/jobs/492621989#L395

Were you able to run it locally?

wp1-frontend/cypress/integration/home_page.spec.js
audiodude: Okay I just patched your PR and tried the tests locally and they pass locally. So it must be timing out because the docker server is too slow in CI.

audiodude:
> Were you able to run it locally?

Okay I just patched your PR and tried the tests locally and they pass locally. So it must be timing out because the docker server is too slow in CI.

wp1-frontend/cypress/integration/home_page.spec.js
mahakporwal02: @audiodude Do you mean I should make a new file and add this code in it?

audiodude: No, just that it should be part of the `describe` block above. So like the other `it` declarations in this file, it should be part of the function passed to `describe('the home page', ...`

kelson42:
Two small remarks:
* We migrate all our ptojects from Travis CI to Github Actions. One of the advantages is that it is more reliable
* We try to use more and more the Github Docker repository because it does not have such limitations like Docker hub.

wp1-frontend/cypress/integration/home_page.spec.js
mahakporwal02: I think it is part of describe('the home page') block like other it declarations.

audiodude: Yes you're right.

audiodude:
@kelson42 So should we hold off on this PR until we can migrate to Github Actions? What limitations of Docker hub are you referring to? Docker hub doesn't seem to be the issue here.

audiodude:
@mahakporwal02  Can you push a new commit here and see if the new CI with GIthub Actions picks it up and if it passes? Thanks!

mahakporwal02:
@audiodude I don't get why cypress test is failing, They are running fine locally. Please have a look.
Screenshot from 2021-03-25 13-28-18

audiodude:
I'm not sure why it's failing either, my initial guess was that it was timing out somehow in Travis CI, but we're now seeing the same (mis)behavior in Github actions.

I've submitted #297 to upload the videos and screenshots from Cypress runs. Can you try merging with master and pushing the code again to see what artifacts we get from Cypress?

mahakporwal02:
> I'm not sure why it's failing either, my initial guess was that it was timing out somehow in Travis CI, but we're now seeing the same (mis)behavior in Github actions.
> 
> I've submitted #297 to upload the videos and screenshots from Cypress runs. Can you try merging with master and pushing the code again to see what artifacts we get from Cypress?
> 
> EDIT: Actually I was able to re-run CI myself, let's see if it picks up my upload changes.
> EDIT 2: No it didn't pick up the upload changes, can you try merging master like I suggested?

Done, please check it.

audiodude:
@mahakporwal02 Can you try merging master again? I made a change to the workflow so that it always uploads the artifacts even if the test fails (#300). Currently it's bailing on the upload because the previous step failed.

mahakporwal02:
> @mahakporwal02 Can you try merging main again? I made a change to the workflow so that it always uploads the artifacts even if the test fails (#300). Currently it's bailing on the upload because the previous step failed.

Done.

kelson42:
> @kelson42 So should we hold off on this PR until we can migrate to Github Actions? What limitations of Docker hub are you referring to? Docker hub doesn't seem to be the issue here.

I let you decide this. Does not seem to be a bad approach to me. The problem with Docker hub are:
* Its relative lack of stability
* Its newly limitation/quota in terme of pull

Perfectly possible you are not impacted here.

audiodude:
So by looking at the video and screenshot, it seems clear that the call is simply timing out.

I was going to suggest adding a `cy.wait()` command here but [the documentation](https://docs.cypress.io/guides/references/best-practices#Unnecessary-wait-for-cy-get) strongly recommends against it. So we should instead use the pattern that they suggest, intercepting the call to /table and waiting on that.

audiodude:
I'm going to cherrypick the Codefactor fix into the main branch so that all of these PRs don't suffer from it. Thanks for that!

mahakporwal02:
> So by looking at the video and screenshot, it seems clear that the call is simply timing out.
> 
> I was going to suggest adding a `cy.wait()` command here but [the documentation](https://docs.cypress.io/guides/references/best-practices#Unnecessary-wait-for-cy-get) strongly recommends against it. So we should instead use the pattern that they suggest, intercepting the call to /table and waiting on that.

cy.intercept doesn't run in current version of cypress.Should I update it then?

audiodude:
> cy.intercept doesn't run in current version of cypress.Should I update it then?

Yes let's update it, but as a separate issue and PR. I think the cy.intercept() changes are the main differences between Cypress 5 (which we're on) and 6 (which we'll go to). And since we don't use it yet, it should be a smooth upgrade.

mahakporwal02:
@audiodude I am stuck on cy.wait() part. Pls have a look.
Screenshot from 2021-03-27 19-46-50
Screenshot from 2021-03-27 19-47-02

audiodude:
@mahakporwal02 You want to intercept the project table call. Intercept is used for spying on or mocking XHR calls that the page makes. So here you would want to spy on the call to the backend, `/v1/projects/Alien/table`:

Screen Shot 2021-03-27 at 12 17 47 PM

You should probably match on the string `'/v1/projects/Alien/table'` instead of the full url, so that the test doesn't break if the backend port changes for any reason.

Hope this helps.

mahakporwal02:
@audiodude I have tried adding intercept as suggested but it still doesn't pass the test.

audiodude:
@mahakporwal02 Does it pass locally still?

mahakporwal02:
> @mahakporwal02 Does it pass locally still?

Yes, it passes locally.

audiodude:
@mahakporwal02 Okay, so I think I found the problem and fixed it in #309. Can you revert the test to the version without cy.intercept and try again?

.github/workflows/workflow.yml
```diff
@@ -89,7 +89,7 @@ jobs:
           python -m flask run &
           cd wp1-frontend
           $(yarn bin)/wait-on http://localhost:5000/
-          curl http://localhost:5000/v1/projects/
+          curl http://localhost:5000/v1/projects/Alien/table
```
audiodude, line 92: Could you revert this back to just /v1/projects please?

audiodude (changes requested):

mahakporwal02:
@audiodude Done.

audiodude (approved):

audiodude:
Thanks for your dedication to getting this one through. The rest should be easy now. 🤞

PR #287, opened 2021-03-24
cypress test for project table containing proper row and column labels

mahakporwal02:
@audiodude Added cypress test for appropriate row and column labels in project-table.
This PR resolves a part of #280.
Codefactor issue is present which I have already resolved in #286.


wp1-frontend/cypress/integration/home_page.spec.js
```diff
@@ -16,4 +16,24 @@ describe('the home page', () => {
       .eq(1)
       .should('contain.text', 'Water');
   });
+
+  it('project-table contains proper row and column labels', () => {
+    cy.visit('#/project/Aesthetics');
+    cy.request(`${Cypress.env('VUE_APP_API_URL')}/projects/Aesthetics/table`).then((resp) => {
```
audiodude, line 22: This seems to be testing that the API response contains labels that match itself. I'm more interested in checking that the actual table on the screen has specific column and row labels that we expect. So your test setup should be similar to the previous tests (type in a query, click a result), then it should expect against the actual column and row labels in the `table` that is displayed. It's okay to use constants to check against because like I mentioned, the response data is constant.

audiodude (changes requested):

audiodude:
The error message in CI for this one is a bit of a clue: https://travis-ci.com/github/openzim/wp1/jobs/493461547#L403

It seems the web API backend is not getting started properly?

audiodude (approved):
LGTM in the revised version. We can't check this in unless we can get it passing on Travis (or until we switch to Github actions) though.

audiodude:
@mahakporwal02 Can you push this again so that it runs against the new Github Actions?

wp1-frontend/cypress/integration/home_page.spec.js
```diff
@@ -16,4 +16,21 @@ describe('the home page', () => {
       .eq(1)
       .should('contain.text', 'Water');
   });
+
+  it('project-table contains proper row and column labels', () => {
+    cy.visit('/');
```
audiodude, line 21: I'm actually thinking now that instead of making this a "home page" test and clicking through the results, that we should actually move this to the proposed "project page" test spec. Then we can just `cy.visit('/#/project/Aesthetics')` and go from there. What do you think?

audiodude (changes requested):
Had second thoughts about this one, let me know what you think.

wp1-frontend/cypress/integration/home_page.spec.js
audiodude: Oh and of course, now that I notice it one place I notice it everywhere: `it` statements should read like sentences.

audiodude (approved):

Issue #294, opened 2021-03-25
Github Actions: Write a workflow that uses Cypress


audiodude:
One potential drawback of the Travis CI setup for Cypress tests was that it used the latest docker hub image for the python web code. So if you were making a change to that code and adding a test for it in the same PR, that wouldn't work.

That's something that could probably be improved with the Github Actions based approach.

PR #296, opened 2021-03-25
Rework Github actions workflow to include Cypress e2e tests

audiodude:
Also re-organize the file so the jobs run in parallel, and delete the .travis.yml configuration file so that Travis CI no longer runs on the project.


audiodude:
Fixes #294

PR #298, opened 2021-03-25
added cypress test for article detail table

mahakporwal02:
@audiodude Added cypress test for "Clicking a number in the table brings the user to the appropriate article detail list page"
Edit: I think it tests "It displays the right name for the page that it is (ie "Water articles - High importance / Start quality)" also.
This resolves a part of #280


wp1-frontend/cypress/integration/home_page.spec.js
```diff
@@ -16,4 +16,20 @@ describe('the home page', () => {
       .eq(1)
       .should('contain.text', 'Water');
   });
-});
+
+  it('article detail-list properly displayed', () => {
```
audiodude, line 20: The string in these it statements should read as a proper sentence. So it would be better as:

`it('displays the article detail-list')`

You also don't need to put  'properly' because we won't normally be testing that something is done improperly.

wp1-frontend/cypress/integration/home_page.spec.js
```diff
@@ -16,4 +16,20 @@ describe('the home page', () => {
...
+    cy.visit('/#/project/Alien');
+    
+    cy.get('tr').eq(3).find('td').eq(0).invoke('text').then((text) => {
+      cy.get('tr').eq(3).find('td').eq(0).contains(/^\d+$/).click();
```
audiodude, line 24: This line doesn't use the `text` variable that you invoked. Is there another way of writing this? Can you assign the result of `cy.get('tr').eq(3).find('td').eq(0)` to a variable?

wp1-frontend/cypress/integration/home_page.spec.js
```diff
@@ -16,4 +16,20 @@ describe('the home page', () => {
       .eq(1)
       .should('contain.text', 'Water');
   });
-});
+
+  it('article detail-list properly displayed', () => {
+    cy.visit('/#/project/Alien');
```
audiodude, line 21: If you're visiting a page other than the home page, this should probably go in its own spec. I would call it the `article_page`.

audiodude (changes requested):

mahakporwal02:
@audiodude cypress commands are asynchronous so it may create an issue when I assign `cy.get('tr').eq(3).find('td').eq(0)` to a variable as it might return a promise.

audiodude:
> @audiodude cypress commands are asynchronous so it may create an issue when I assign `cy.get('tr').eq(3).find('td').eq(0)` to a variable as it might return a promise.

Yup it even tells you that in codefactor: https://www.codefactor.io/repository/github/openzim/wp1/pull/298

mahakporwal02:
@audiodude So should just I revert these changes to previous ones?

wp1-frontend/cypress/integration/home_page.spec.js
audiodude: Looks like assigning variables is not the way to do things in Cypress 🤷

audiodude:
> @audiodude So should I just revert these changes to previous ones?

I'm not sure you want to revert all of your changes, but you want to go back to a version that doesn't assign a variable yes.

audiodude (approved):

PR #301, opened 2021-03-26
Cypress test for filter by article name works

mahakporwal02:
@audiodude Added cypress test for "Filtering by article name works".
This resolves a part of #280

### DIFF
--- a/wp1-frontend/cypress/integration/article_page.spec.js
+++ b/wp1-frontend/cypress/integration/article_page.spec.js
@@ -10,7 +10,7 @@ describe('the article page', () => {
         cy.get('button').eq(3).click();
 
         cy.get('table').find('tr').each(($el) => {
-            cy.wrap($el).contains('Predator');
+            cy.wrap($el).should('contain.text', 'Predator');
         });
     });
 });
--- a/wp1-frontend/cypress/integration/article_page.spec.js
+++ b/wp1-frontend/cypress/integration/article_page.spec.js
@@ -1,0 +1,16 @@
+/// <reference types="Cypress" />
+
+describe('the article page', () => {
+    it('filters by article name', () => {
+        cy.visit('/#/project/Alien/articles');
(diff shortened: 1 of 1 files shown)


wp1-frontend/cypress/integration/home_page.spec.js
```diff
@@ -16,4 +16,17 @@ describe('the home page', () => {
       .eq(1)
       .should('contain.text', 'Water');
   });
-});
+
+  it('filter by article name', () => {
+    cy.visit('/#/project/Alien/articles');
```
audiodude, line 21: Same thing, this is not the home page.

wp1-frontend/cypress/integration/home_page.spec.js
```diff
@@ -16,4 +16,17 @@ describe('the home page', () => {
       .eq(1)
       .should('contain.text', 'Water');
   });
-});
+
+  it('filter by article name', () => {
```
audiodude, line 20: Same thing again, the it should be a sentence. Here you can probably get by with just `it('filters by article name')`.

audiodude (changes requested):

wp1-frontend/cypress/integration/article_page.spec.js
```diff
@@ -4,7 +4,7 @@ describe('the article page', () => {
     it('filters by article name', () => {
         cy.visit('/#/project/Alien/articles');
 
-        cy.contains('Filter by article name').click();
+        cy.get('a').eq(8).should('contain.text', 'Filter by article name').click();
```
audiodude, line 7: I think for table rows and columns, getting them by index is fine.

For this, if we ever add another link to the page, this will break. Do you need the eq(8) at all?

audiodude (changes requested):

audiodude (approved):

PR #302, opened 2021-03-26
cypress test for custom pagination

mahakporwal02:
@audiodude Added cypress test for custom pagination.
This resolves a part of #280

### DIFF
--- a/wp1-frontend/cypress/integration/article_page.spec.js
+++ b/wp1-frontend/cypress/integration/article_page.spec.js
@@ -24,22 +24,35 @@ describe('the article page', () => {
         cy.wrap($el).should('contain.text', 'Predator');
       });
   });
-        cy.get('table').find('tr').each(($el) => {
-            cy.wrap($el).should('contain.text', 'Predator');
-        });
-    });
 });
 
 describe('custom pagination', () => {
     it('shows 50 rows in article-table', () => {
         cy.visit('/#/project/Alien/articles');
+        cy.contains('Custom pagination')
+          .click();
+
+        cy.get('input')
+          .eq(0)
+          .clear()
+          .type('50');
+
+        cy.get('input')
+          .eq(1)
+          .clear()
+          .type('2');
+
+        cy.get('button')
+          .eq(1)
+          .click();
     
-        cy.contains('Custom pagination').click();
-        cy.get('input').eq(0).clear().type('50');
-        cy.get('input').eq(1).clear().type('2');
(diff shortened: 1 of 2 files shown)


wp1-frontend/cypress/integration/home_page.spec.js
```diff
@@ -16,4 +16,16 @@ describe('the home page', () => {
...
+
+  it('custom pagination works', () => {
+    cy.visit('/#/project/Alien/articles');
+
+    cy.contains('Custom pagination').click();
+    cy.get('input').eq(0).clear().type('50');
+    cy.get('input').eq(1).clear().type('2');
+    cy.get('button').eq(1).click();
+
+    cy.get('tr').eq(0).find('td').eq(0).should('have.text', '51');
```
audiodude, line 28: What is the difference with `have.text` versus `contain.text` that we use elsewhere. Does have match exactly?

wp1-frontend/cypress/integration/home_page.spec.js
```diff
@@ -16,4 +16,16 @@ describe('the home page', () => {
       .eq(1)
       .should('contain.text', 'Water');
   });
+
+  it('custom pagination works', () => {
+    cy.visit('/#/project/Alien/articles');
```
audiodude, line 21: You're not visiting the home page, so this doesn't belong in `home_page.spec.js`. Perhaps you can create a file for the article page, `article_page.spec.js`?

audiodude (changes requested):

wp1-frontend/cypress/integration/home_page.spec.js
mahakporwal02: Yes, have.text matches exactly.

wp1-frontend/cypress/integration/article_page.spec.js
```diff
@@ -25,3 +25,34 @@ describe('the article page', () => {
       });
   });
 });
+
+describe('custom pagination', () => {
```
audiodude, line 29: Put this under the `describe('article page')` block, because it's describing the custom pagination on that page. In general, a spec.js file should have one master describe block with other describes nested.

audiodude (changes requested):

audiodude (approved):
One nitpick: You shouldn't have to force push the new changes unless you've edited the commits: re-ordered, deleted some, or squashed the. If you haven't done that, please just do a normal push because it will make the code review easier to follow.

PR #303, opened 2021-03-26
Fix Codefactor for cypress related changes

audiodude:
In my original Cypress PR, I ignored the Codefactor warnings for undefined symbols because I didn't know how to fix them. These have now been fixed in this PR.


audiodude:
The Cypress failures on this PR are concerning. Reading through [a related issue](https://github.com/cypress-io/cypress/issues/2938) on the Cypress repo it does mention that upgrading Cypress might fix it. Let's do #304 first then come back to this one.

PR #309, opened 2021-03-27
Add Redis to Cypress CI job

audiodude:
We were already using a Redis service in the `python-nose` job, but it's also required in the `frontend-cypress` job.


audiodude:
This should fix the failures we're seeing in #286

PR #316, opened 2021-03-31
Development versions of manual update API server code

audiodude:
This PR introduces a `dev_projects` blueprint. If the code is running where ENV == Environment.DEVELOPMENT (as is the case in the example credentials, and is NOT the case in production), the Flask application setup will include dev_projects first, and then the regular projects blueprint.

The effect of this is that endpoints defined in `dev_projects` are used instead of/before the regular ones. This allows us to mock out the manual update process in development, which is what it is immediately used for.

As provided in this PR, the manual update is API compatible with the normal path as far as the frontend is concerned. However it has the following features:

- The "lockout" time between updates is 5 minutes instead of 1 hour.
- All updates take 60 seconds, with the progress being 1/60, 2/60 etc as the seconds pass.
- If the update is for the project 'Water', it will fail after 10 seconds.

Future work would be to make the update duration and wait time configurable. Once that is done, this could be used in CI  to test the manual update page.

### DIFF
--- a/wp1/web/dev/projects.py
+++ b/wp1/web/dev/projects.py
@@ -1,3 +1,13 @@
+'''Development overrides for the /v1/projects endpoints.
+
+The functions provided in this file are API compatible with the normal
+path as far as the frontend is concerned. However this blueprint has the following features:
+
+- The "lockout" time between updates is 5 minutes instead of 1 hour.
+- All updates take 60 seconds, with the progress being 1/60, 2/60 etc as the seconds pass.
+- If the update is for the project 'Water', it will fail after 10 seconds.
+'''
+
 from datetime import datetime, timedelta
 
 import flask
--- a/wp1/web/dev/projects.py
+++ b/wp1/web/dev/projects.py
@@ -70,7 +70,7 @@ def get_project_progress(redis, project_name):
   dt = datetime.strptime(ts, '%Y-%m-%d %H:%M:%S UTC')
   progress = _progress_secs(dt)
 
-  if project_name == b'Water' and progress > 10:
+  if project_name == b'Water' and progress >= 10:
     return 10, 60
   return progress, 60
 
@@ -83,11 +83,16 @@ def get_project_queue_status(redis, project_name):
   if progress < 5:
     return {'status': 'queued'}
 
-  if progress > 10 and project_name == b'Water':
+  if progress >= 10 and project_name == b'Water':
     return {'status': 'failed'}
 
   if progress > 50:
-    return {'status': 'finished', 'ended_at': utcnow()}
+    ts = next_update_time(redis, project_name)
+    if ts is None:
+      end = utcnow()
+
+    dt = datetime.strptime(ts, '%Y-%m-%d %H:%M:%S UTC')
+    return {'status': 'finished', 'ended_at': dt - timedelta(minutes=4)}
 
   return {'status': 'started'}
 
--- a/wp1/web/dev/projects.py
+++ b/wp1/web/dev/projects.py
@@ -86,7 +86,7 @@ def get_project_queue_status(redis, project_name):
     return {'status': 'failed'}
 
   if progress > 50:
-    return {'status': 'finished', 'ended_at': datetime.now()}
+    return {'status': 'finished', 'ended_at': utcnow()}
 
   return {'status': 'started'}
 
--- a/wp1/web/dev/projects.py
+++ b/wp1/web/dev/projects.py
@@ -80,15 +80,15 @@ def get_project_queue_status(redis, project_name):
     return None
 
   if progress < 5:
-    return 'queued'
+    return {'status': 'queued'}
 
   if progress > 10 and project_name == b'Water':
-    return 'failed'
+    return {'status': 'failed'}
 
   if progress > 50:
-    return 'finished'
+    return {'status': 'finished'}
 
-  return 'started'
+  return {'status': 'started'}
 
 
 @dev_projects.route('/<project_name>/update', methods=['POST'])
--- a/wp1/web/dev/projects.py
+++ b/wp1/web/dev/projects.py
@@ -39,8 +39,9 @@ def _project_progress_key(project_name):
 
 def mark_project_manual_update_time(redis, project_name):
   key = _manual_key(project_name)
-  ts = (utcnow() +
-        timedelta(minutes=UPDATE_DURATION_MINS)).strftime('%Y-%m-%d %H:%M UTC')
+  ts = (
+      utcnow() +
+      timedelta(minutes=UPDATE_DURATION_MINS)).strftime('%Y-%m-%d %H:%M:%S UTC')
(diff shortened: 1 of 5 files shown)


audiodude:
@mahakporwal02 Would you mind doing a code review?

audiodude:
I'm inclined to submit this without code coverage, because all of the code is just development/test code.

mahakporwal02:
@audiodude I'll have a look at it once I make required changes in previous PRs.

wp1/web/dev/projects.py
```diff
@@ -0,0 +1,130 @@
...
+UPDATE_DURATION_MINS = 5
+
+
+def clear_project_progress(redis, project_name):
+  key = _project_progress_key(project_name)
+  redis.delete(key)
+
+
+def enqueue_single_project(redis, project_name):
+  clear_project_progress(redis, project_name)
```
mahakporwal02, line 29: Why to add an extra function? Can't we directly call `clear_project_progress` or are we trying to display the same procedure as the production manual update?

wp1/web/dev/projects.py
```diff
@@ -0,0 +1,130 @@
...
+  ts = next_update_time(redis, project_name)
+  if ts is None:
+    return None, None
+
+  dt = datetime.strptime(ts, '%Y-%m-%d %H:%M UTC')
+  progress = _progress_secs(dt)
+
+  if project_name == b'Water' and progress > 10:
+    return 10, 60
+  return progress, 60
```
mahakporwal02, line 74: In this condition do you mean to show the scenario in #186, where the `progress` is not complete but the `update time` is over?

mahakporwal02 (changes requested):
@audiodude Shouldn't we need to update `readme.md` in case the developer needs to test the backend of the manual update.About how to switch blueprint to `projects`?

wp1/web/dev/projects.py
```diff
@@ -0,0 +1,130 @@
...
+  if progress < 5:
+    return 'queued'
+
+  if progress > 10 and project_name == b'Water':
+    return 'failed'
+
+  if progress > 50:
+    return 'finished'
+
+  return 'started'
```
mahakporwal02, line 91: Return statement should be of the form `{'status':'started'}`, and for finished case `{'status': 'finished', 'ended_at': ending time}`

wp1/web/dev/projects.py
```diff
@@ -0,0 +1,130 @@
...
+def get_project_queue_status(redis, project_name):
+  progress, work = get_project_progress(redis, project_name)
+  if not progress:
+    return None
+
+  if progress < 5:
+    return 'queued'
+
+  if progress > 10 and project_name == b'Water':
+    return 'failed'
```
mahakporwal02, line 86: Also this won't work as progress returned would always be less than or equal to 10 sec.

wp1/web/dev/projects.py
```diff
@@ -0,0 +1,130 @@
...
+    return 60
+  return 60 - secs.seconds
+
+
+def get_project_progress(redis, project_name):
+  ts = next_update_time(redis, project_name)
+  if ts is None:
+    return None, None
+
+  dt = datetime.strptime(ts, '%Y-%m-%d %H:%M UTC')
```
mahakporwal02, line 69: This should also include seconds to get correct value of `secs`.

audiodude:
> @audiodude Shouldn't we need to update `readme.md` in case the developer needs to test the backend of the manual update.About how to switch blueprint to `projects`?

Good idea. Done.

wp1/web/dev/projects.py
audiodude: Right, the way I developed this was to stub out all of the production functions. It's probably easier to follow this way, if you know how the production functions work. It makes more sense for the API endpoint to call `enqueue_single_project` than to call `clear_project_progress`.

wp1/web/dev/projects.py
audiodude: Yes, I should add that scenario. I should also add the scenario where the update is "in progress" but the progress is greater than the "total".

wp1/web/dev/projects.py
audiodude: Good catch, thanks! Done.

wp1/web/dev/projects.py
audiodude: Good point! Done.

wp1/web/dev/projects.py
audiodude: Do you mean because I wasn't producing the seconds on the timestamp? I'm not sure I understand.

audiodude: I get it now, just tested it. Fixing.

audiodude:
@mahakporwal02 Okay I've completed some major changes to this, incorporated your feedback (thanks!) and tried to make it configurable so that it can run in CI. PTAL.

I'm a bit afraid of how complex it's gotten, but I think it's better than not being able to test the manual update page.

wp1/web/dev/projects.py
mahakporwal02: OK, got it.

wp1/web/dev/projects.py
audiodude: Added those for "Aesthetics" and "Basic Income"

wp1/web/dev/projects.py
mahakporwal02: OK, got it.

wp1/web/dev/projects.py
mahakporwal02: > @mahakporwal02 Okay I've completed some major changes to this, incorporated your feedback (thanks!) and tried to make it configurable so that it can run in CI. PTAL.
> 
> I'm a bit afraid of how complex it's gotten, but I think it's better than not being able to test the manual update page.

It has been very helpful to me while I was testing my last PR.

mahakporwal02:
@audiodude It has been very helpful while I was testing my last PR. I'll see if I could help you more with these changes.

audiodude:
> @audiodude It has been very helpful while I was testing my last PR. I'll see if I could help you more with these changes.

Awesome, good to hear. I think if we can get a first version in, even if it's too complex, we can iterate from there.

mahakporwal02:
> > @audiodude It has been very helpful while I was testing my last PR. I'll see if I could help you more with these changes.
> 
> Awesome, good to hear. I think if we can get a first version in, even if it's too complex, we can iterate from there.

Seems good.

audiodude:
Merging with codecov failing, this is development code I don't think it needs to be tested.

The next step is to write some Cypress tests for this page.

mahakporwal02:
> Merging with codecov failing, this is development code I don't think it needs to be tested.
> 
> The next step is to write some Cypress tests for this page.

Should I work on cypress tests?

audiodude:
Yes, feel free to file an issue and write some tests! I'll work on fixing #317

Issue #317, opened 2021-03-31
Cypress tests broken/flaky for article page


audiodude:
Looks like the test added in #301 is either broken or flaky.

https://github.com/openzim/wp1/actions/runs/706346926

Reading https://www.cypress.io/blog/2020/07/22/do-not-get-too-detached/ makes me think that we are grabbing all of the `<tr>`s before the table is updated, and then when we go to check them (after the update), the list is stale and some of them have become detached.

audiodude:
@mahakporwal02 would you be able to take a look at this?

audiodude:
This is still happening: https://github.com/openzim/wp1/pull/316/checks?check_run_id=2265489194

PR #322, opened 2021-04-05
Cypress test to display articles with selected quality and importance

mahakporwal02:
@audiodude Added cypress test for widget you mentioned in #280.
This resolves a part of #280.


mahakporwal02:
@audiodude I think problem similar to #317 will occur in this one too.
So what should be done now?

audiodude:
> @audiodude I think problem similar to #317 will occur in this one too.
> So what should be done now?

@mahakporwal02 We should figure out how to fix it and not submit this until we have a fix. Thanks for pointing that out.

PR #345, opened 2021-05-30
User login with OAuth via mwoauth library

mahakporwal02:
This PR adds a new `oauth` blueprint to the backend API server that initiates and completes the OAuth login flow with the Meta Wikipedia server. The end result is that our app receives a username and user id for the user, and they are "logged in" to our app, via a session cookie which stores the same.

### DIFF
--- a/wp1-frontend/src/App.vue
+++ b/wp1-frontend/src/App.vue
@@ -56,6 +56,15 @@
             >
           </li>
         </ul>
+        <div>
+          <div v-if="this.token">
+            <button @click="logout">Logout</button>
+            <span>Username</span>
+          </div>
+          <a v-else href="http://127.0.0.1:5000/v1/oauth/initiate"
+            ><button>Login</button></a
+          >
+        </div>
       </div>
     </nav>
     <div id="app" class="container">
@@ -66,7 +75,38 @@
 
 <script>
 export default {
-  name: 'app'
+  name: 'app',
+  data: function() {
+    return {
+      token: null
+    };
+  },
+  methods: {
+    logout: async function() {
+      localStorage.removeItem('token');
+      window.location.href = 'http://localhost:3000/#/';
+      await fetch(`${process.env.VUE_APP_API_URL}/oauth/logout`, {
+        method: 'POST',
+        body: JSON.stringify(this.token)
+      });
+      this.token = null;
+      localStorage.removeItem('token', this.token);
+    }
+  },
+  created: async function() {
+    if (!window.location.hash.split('?')[1]) {
+      return;
+    }
+    this.token = window.location.hash.split('?')[1];
+    console.log(this.token);
+    const response = await fetch(`${process.env.VUE_APP_API_URL}/oauth/token`, {
+      credentials: 'include',
+      method: 'GET',
+      body: JSON.stringify(this.token)
+    });
+    this.token = await response.json();
+    localStorage.setItem('token', this.token);
+  }
 };
 </script>
 
--- a/wp1-frontend/src/App.vue
+++ b/wp1-frontend/src/App.vue
@@ -57,13 +57,11 @@
           </li>
         </ul>
         <div>
-          <div v-if="this.token">
+          <div v-if="this.username">
             <button @click="logout">Logout</button>
-            <span>Username</span>
+            <span> {{ this.username }} </span>
           </div>
-          <a v-else href="http://127.0.0.1:5000/v1/oauth/initiate"
-            ><button>Login</button></a
-          >
+          <a v-else @click="login"><button>Login</button> </a>
         </div>
       </div>
     </nav>
@@ -78,34 +76,45 @@ export default {
   name: 'app',
   data: function() {
     return {
-      token: null
+      username: null
     };
   },
   methods: {
     logout: async function() {
-      localStorage.removeItem('token');
-      window.location.href = 'http://localhost:3000/#/';
-      await fetch(`${process.env.VUE_APP_API_URL}/oauth/logout`, {
-        method: 'POST',
-        body: JSON.stringify(this.token)
+      await fetch(`${process.env.VUE_APP_API_URL}/oauth/logout`);
+      this.username = null;
+    },
+    login: async function() {
+      await fetch(`${process.env.VUE_APP_API_URL}/oauth/initiate`, {
+        credentials: 'include'
       });
-      this.token = null;
-      localStorage.removeItem('token', this.token);
+      this.identify();
+    },
+    identify: async function() {
+      if (this.username) {
+        return;
+      }
+      const response = await fetch(
+        `${process.env.VUE_APP_API_URL}/oauth/identify`
+      )
+        .then(response => {
+          if (response.status === 200) {
+            return response.json();
+          } else {
+            throw new Error(response.status);
+          }
+        })
+        .then(data => {
+          return data;
+        })
+        .catch(err => {
+          console.error(err);
+        });
+      this.username = response;
     }
   },
-  created: async function() {
-    if (!window.location.hash.split('?')[1]) {
-      return;
-    }
-    this.token = window.location.hash.split('?')[1];
-    console.log(this.token);
-    const response = await fetch(`${process.env.VUE_APP_API_URL}/oauth/token`, {
-      credentials: 'include',
-      method: 'GET',
-      body: JSON.stringify(this.token)
-    });
-    this.token = await response.json();
-    localStorage.setItem('token', this.token);
+  created: function() {
+    this.identify();
   }
 };
 </script>
--- a/wp1-frontend/src/App.vue
+++ b/wp1-frontend/src/App.vue
@@ -59,9 +59,10 @@
         <div>
           <div v-if="this.username">
             <button @click="logout">Logout</button>
+            <span style="font-size:20px"> | </span>
             <span> {{ this.username }} </span>
           </div>
-          <a v-else @click="login"><button>Login</button> </a>
+          <a v-else :href="this.loginInitiateUrl"><button>Login</button> </a>
         </div>
       </div>
     </nav>
@@ -76,41 +77,40 @@ export default {
   name: 'app',
   data: function() {
     return {
-      username: null
+      username: null,
+      loginInitiateUrl: `${process.env.VUE_APP_API_URL}/oauth/initiate`
     };
   },
   methods: {
     logout: async function() {
-      await fetch(`${process.env.VUE_APP_API_URL}/oauth/logout`);
-      this.username = null;
-    },
-    login: async function() {
-      await fetch(`${process.env.VUE_APP_API_URL}/oauth/initiate`, {
+      await fetch(`${process.env.VUE_APP_API_URL}/oauth/logout`, {
         credentials: 'include'
       });
-      this.identify();
+      this.username = null;
     },
     identify: async function() {
       if (this.username) {
         return;
       }
-      const response = await fetch(
-        `${process.env.VUE_APP_API_URL}/oauth/identify`
+      const data = await fetch(
+        `${process.env.VUE_APP_API_URL}/oauth/identify`,
+        {
+          credentials: 'include'
+        }
       )
         .then(response => {
-          if (response.status === 200) {
+          if (response.ok) {
             return response.json();
           } else {
             throw new Error(response.status);
           }
         })
-        .then(data => {
-          return data;
-        })
         .catch(err => {
           console.error(err);
         });
-      this.username = response;
+      if (data) {
+        this.username = data.username;
+      }
     }
   },
   created: function() {
--- a/wp1/base_db_test.py
+++ b/wp1/base_db_test.py
@@ -120,15 +120,6 @@ def get_test_connect_creds():
               'host': 'localhost',
               'port': 55555
           },
-          'API': {},
-          'MWOAUTH': {
-              'consumer_key': 'consumer_key',
-              'consumer_secret': 'consumer_secret'
-          },
-          'SESSION': {
-              'secret_key': 'wp1_secret'
-          },
-          'CLIENT_URL': 'http://localhost:3000/#/'
       },
       Environment.PRODUCTION: {}
   }
--- a/wp1/base_db_test.py
+++ b/wp1/base_db_test.py
@@ -120,6 +120,15 @@ def get_test_connect_creds():
               'host': 'localhost',
               'port': 55555
           },
+          'API': {},
+          'MWOAUTH': {
+              'consumer_key': 'consumer_key',
+              'consumer_secret': 'consumer_secret'
+          },
+          'SESSION': {
+              'secret_key': 'wp1_secret'
+          },
+          'CLIENT_URL': 'http://localhost:3000/#/'
       },
       Environment.PRODUCTION: {}
   }
--- a/wp1/credentials.py.example
+++ b/wp1/credentials.py.example
@@ -52,12 +52,16 @@ CREDENTIALS = {
         'API': {}
 
         # Credentials for authentication through mwoauth.
+        # To register the app go to 'https://meta.wikimedia.org/wiki/Special:OAuthConsumerRegistration/propose'
         'MWOAUTH': {
             'consumer_key': '',
             'consumer_secret': ''
         },
 
         # Secret key for session.
+        # To get your random secret key for production use the following script
+        # python3
+        # import os
         'SESSION': {
             'secret_key': 'WP1_secret_key'
         },
--- a/wp1/credentials.py.example
+++ b/wp1/credentials.py.example
@@ -50,15 +50,16 @@ CREDENTIALS = {
         # WMF wiki OAuth credentials. Used to edit English Wikipedia.
         # These don't exist in development.
         'API': {}
+
+        # Credentials for authentication through mwoauth
         'MWOAUTH': {
             'consumer_key': '',
             'consumer_secret': ''
         },
+        # secret key for session
         'SESSION': {
-            'secret_key': 'wp1',
-            'domain': 'api.wp1.openzim.org'
-        },
-        'CLIENT_URL': 'http://localhost:3000/#/'
+            'secret_key': 'WP1_secret_key'
+        }
     },
 
     # EDIT: Remove the next line after you've provided actual credentials.
--- a/wp1/credentials.py.example
+++ b/wp1/credentials.py.example
@@ -51,15 +51,19 @@ CREDENTIALS = {
         # These don't exist in development.
         'API': {}
 
-        # Credentials for authentication through mwoauth
+        # Credentials for authentication through mwoauth.
         'MWOAUTH': {
             'consumer_key': '',
             'consumer_secret': ''
         },
-        # secret key for session
+
+        # Secret key for session.
         'SESSION': {
             'secret_key': 'WP1_secret_key'
-        }
+        },
+
+        # Client side url used for redirection in login process.
+        'CLIENT_URL': 'http://localhost:3000/#/'
     },
 
     # EDIT: Remove the next line after you've provided actual credentials.
--- a/wp1/credentials.py.example
+++ b/wp1/credentials.py.example
@@ -62,6 +62,7 @@ CREDENTIALS = {
         # To get your random secret key for production use the following script
         # python3
         # import os
+        # os.urandom(24)
         'SESSION': {
             'secret_key': 'WP1_secret_key'
         },
--- a/wp1/credentials.py.example
+++ b/wp1/credentials.py.example
@@ -50,6 +50,15 @@ CREDENTIALS = {
         # WMF wiki OAuth credentials. Used to edit English Wikipedia.
         # These don't exist in development.
         'API': {}
+        'MWOAUTH': {
+            'consumer_key': '',
+            'consumer_secret': ''
+        },
+        'SESSION': {
+            'secret_key': 'wp1',
+            'domain': 'api.wp1.openzim.org'
+        },
+        'CLIENT_URL': 'http://localhost:3000/#/'
     },
 
     # EDIT: Remove the next line after you've provided actual credentials.
--- a/wp1/web/app.py
+++ b/wp1/web/app.py
@@ -112,12 +112,12 @@ def create_app():
   app.register_blueprint(projects, url_prefix='/v1/projects')
   app.register_blueprint(articles, url_prefix='/v1/articles')
 
-  if CREDENTIALS is None or CREDENTIALS[ENV].get(
-      'MWOAUTH') is None or CREDENTIALS[ENV]['MWOAUTH'].get(
-          'consumer_key') is None or CREDENTIALS[ENV]['MWOAUTH'].get(
-              'consumer_secret') is None:
-    print('No MWOAUTH credentials, not attaching OAuth endpoints')
-  else:
+  missing_credentials = CREDENTIALS is None or ENV is None
+  if not missing_credentials:
+    mwoauth = CREDENTIALS.get(ENV, {}).get('MWOAUTH', {})
+  if not (missing_credentials or mwoauth.get('consumer_key') is None or
+          mwoauth.get('consumer_secret') is None):
     from wp1.web.oauth import oauth
     app.register_blueprint(oauth, url_prefix='/v1/oauth')
+
   return app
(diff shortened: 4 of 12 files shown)


wp1-frontend/src/App.vue
```diff
@@ -66,7 +75,38 @@
...
+    }
+    this.token = window.location.hash.split('?')[1];
+    console.log(this.token);
+    const response = await fetch(`${process.env.VUE_APP_API_URL}/oauth/token`, {
+      credentials: 'include',
+      method: 'GET',
+      body: JSON.stringify(this.token)
+    });
+    this.token = await response.json();
+    localStorage.setItem('token', this.token);
```
audiodude, line 108: We should not fetch the token from the backend or store it in Local Storage.

wp1/web/app.py
```diff
@@ -1,23 +1,25 @@
...
 from wp1.web.db import get_db, has_db
 from wp1.web.articles import articles
+from wp1.web.oauth import oauth
 from wp1.web.projects import projects
 from wp1.web.dev.projects import dev_projects
 
 try:
   # The credentials module isn't checked in and may be missing
-  from wp1.credentials import ENV
+  from wp1.credentials import ENV, CREDENTIALS
```
audiodude, line 22: Code factor is right that we should have one import for ENV and CREDENTIALS. However, if we don't do the import in get_redis_creds, then in cases where credentials.py doesn't exist, it will be a different error (either a KeyError or a NameError with CREDENTIALS being undefined). Either way, that would cause the server to error out with a cryptic error message.

We want to gracefully handle the case where wp1.credentials doesn't exist.

Perhaps in the except block here you can set ENV = None and CREDENTIALS = None, then check for those values in get_redis_creds() and print an appropriate message if they don't exist.

wp1/web/app.py
```diff
@@ -69,6 +71,14 @@ def create_app():
   else:
     print('No RQ_USER/RQ_PASS found in env. RQ dashboard not created.')
 
+  app.config['SESSION_TYPE'] = 'redis'
+  app.config['SESSION_REDIS'] = redis.from_url('redis://{}:{}'.format(
+      redis_creds['host'], redis_creds['port']))
+  app.config['SECRET_KEY'] = CREDENTIALS[ENV]['SESSION']['secret_key']
```
audiodude, line 77: Similar to the above, if credentials.py doesn't exist, this will probably blow up with a NameError.

Instead, create a get_session_settings() method that works similarly to get_redis_creds() (and warns the user if credentials.py was undefined).

wp1/web/oauth.py
```diff
@@ -0,0 +1,52 @@
+import flask
+import jwt
+from flask import request, jsonify, session
+from mwoauth import ConsumerToken, Handshaker, RequestToken
+from wp1.credentials import ENV, CREDENTIALS
```
audiodude, line 5: Just like in app.py, we need to protect for the case that wp1.credentials doesn't exist, with a try and an `except ImportError`.

wp1/web/oauth.py
```diff
@@ -0,0 +1,52 @@
...
+
+oauth = flask.Blueprint('oauth', __name__)
+
+consumer_token = ConsumerToken(CREDENTIALS[ENV]['MWOAUTH']['consumer_key'],
```
audiodude, line 9: If the credentials in fact don't exist, think about what the behavior of this module should be.

wp1/credentials.py.example
```diff
@@ -50,6 +50,15 @@ CREDENTIALS = {
         # WMF wiki OAuth credentials. Used to edit English Wikipedia.
         # These don't exist in development.
         'API': {}
+        'MWOAUTH': {
```
audiodude, line 53: Please write a comment about what each of these are.

wp1/web/oauth.py
```diff
@@ -0,0 +1,52 @@
...
+  access_token = handshaker.complete(RequestToken(**session[request_token_key]),
+                                     query_string)
+  session.pop(request_token_key)
+  session[request_token_key] = dict(zip(access_token._fields, access_token))
+  return flask.redirect('{}?{}'.format(CREDENTIALS[ENV]['CLIENT_URL'],
+                                       request_token_key))
+
+
+@oauth.route('/token')
+def token():
```
audiodude, line 36: This endpoint shouldn't exist

wp1/web/oauth.py
```diff
@@ -0,0 +1,52 @@
...
+                           algorithm="HS256")
+  return jsonify(encoded_jwt)
+
+
+# @oauth.route('/identify')
+
+
+@oauth.route('/logout', methods=['POST'])
+def logout():
+  request_token = request.get_json(force=True)
```
audiodude, line 50: The user shouldn't need the token to logout. However, you might be worried about CRSF for the logout method, which is a valid concern. In that case, we should integrate a Python CSRF plugin (but you can do that in a follow up PR).

audiodude (changes requested):
This is looking really nice and compact, just a few comments.

mahakporwal02:
@audiodude So the sessions in all three routes happen to be different.
 Please try to run the code.
 You might get what problem acually I am facing

mahakporwal02:
@audiodude The problem I am facing is that I am not actually making a request from frontend to backnend, I am redirecting from frontend to backend. So if I redirect how would I get cookies from response?

audiodude (changes requested):
Make sure to run `pip freeze > requirements.txt` so any new libraries you installed are included.

mahakporwal02:
> Make sure to run `pip freeze > requirements.txt` so any new libraries you installed are included.

Done.

wp1/web/app.py
```diff
@@ -69,6 +71,14 @@ def create_app():
...
+  app.config['SESSION_COOKIE_DOMAIN'] = CREDENTIALS[ENV]['SESSION']['domain']
```
audiodude, line 78: We should not set SESSION_COOKIE_DOMAIN, because we're always serving the cookies from the same domain they will be sent to. The browser knows "I got a cookie from api.wp1.openzim.org, so I will send it back to that server". And if we put this line, then the cookies don't work in localhost testing.

audiodude (changes requested):

wp1/web/oauth.py
mahakporwal02: @audiodude I have added error handling but what should be done when credentials.py doesnot exist?
Should we take input from user?

mahakporwal02:
@audiodude Sorry I forgot and force-pushed.
Will remeber it from next time.

wp1-frontend/src/App.vue
```diff
@@ -66,7 +73,49 @@
...
-  name: 'app'
+  name: 'app',
+  data: function() {
+    return {
+      username: null
+    };
+  },
+  methods: {
+    logout: async function() {
+      await fetch(`${process.env.VUE_APP_API_URL}/oauth/logout`);
```
audiodude, line 84: I think you need credentials: include here too, because otherwise how will the backend know which session to logout?

wp1-frontend/src/App.vue
```diff
@@ -66,7 +73,49 @@
...
+    },
+    identify: async function() {
+      if (this.username) {
+        return;
+      }
+      const response = await fetch(
+        `${process.env.VUE_APP_API_URL}/oauth/identify`
+      )
+        .then(response => {
+          if (response.status === 200) {
```
audiodude, line 101: https://developer.mozilla.org/en-US/docs/Web/API/Response/ok

wp1-frontend/src/App.vue
```diff
@@ -66,7 +73,49 @@
...
+      await fetch(`${process.env.VUE_APP_API_URL}/oauth/initiate`, {
+        credentials: 'include'
+      });
+      this.identify();
+    },
+    identify: async function() {
+      if (this.username) {
+        return;
+      }
+      const response = await fetch(
```
audiodude, line 97: I would call this `const data =` because you're returning the data, not the response object.

wp1-frontend/src/App.vue
```diff
@@ -66,7 +73,49 @@
...
+        `${process.env.VUE_APP_API_URL}/oauth/identify`
+      )
+        .then(response => {
+          if (response.status === 200) {
+            return response.json();
+          } else {
+            throw new Error(response.status);
+          }
+        })
+        .then(data => {
```
audiodude, line 107: This is superfluous. It takes data and just returns it.

wp1/credentials.py.example
```diff
@@ -50,6 +50,16 @@ CREDENTIALS = {
         # WMF wiki OAuth credentials. Used to edit English Wikipedia.
         # These don't exist in development.
         'API': {}
+
+        # Credentials for authentication through mwoauth
+        'MWOAUTH': {
+            'consumer_key': '',
+            'consumer_secret': ''
+        },
+        # secret key for session
```
audiodude, line 59: Please explain how this should be populated in production, eg tips on how to generate a good secret key.

wp1-frontend/src/App.vue
```diff
@@ -56,6 +56,13 @@
             >
           </li>
         </ul>
+        <div>
+          <div v-if="this.username">
+            <button @click="logout">Logout</button>
+            <span> {{ this.username }} </span>
```
audiodude, line 62: nit: put a "&nbsp;|&nbsp;" between the button and the span.

wp1/web/app.py
```diff
@@ -64,11 +67,24 @@ def create_app():
...
 
+  try:
+    from wp1.credentials import ENV, CREDENTIALS
+    app.config['SECRET_KEY'] = CREDENTIALS[ENV]['SESSION']['secret_key']
+  except ImportError:
+    print('No secret_key found, using defaults.')
+    app.config['SECRET_KEY'] = 'WP1'
+
+  app.config['SESSION_TYPE'] = 'redis'
+  app.config.from_object(__name__)
```
audiodude, line 85: What does this do? I was under the impression that it's a convenience method for setting config variables based on "floating" variables in the module namespace. Why do we need it here?

wp1/web/oauth.py
```diff
@@ -0,0 +1,54 @@
...
+try:
+  # The credentials module isn't checked in and may be missing
+  from wp1.credentials import ENV, CREDENTIALS
+  consumer_token = ConsumerToken(CREDENTIALS[ENV]['MWOAUTH']['consumer_key'],
+                                 CREDENTIALS[ENV]['MWOAUTH']['consumer_secret'])
+  handshaker = Handshaker("https://en.wikipedia.org/w/index.php",
+                          consumer_token)
+except ImportError:
+  print(
+      'No credentials.py file found, Please add your mwoauth credentials in credentials.py'
```
audiodude, line 16: nit: line length.

wp1/web/oauth.py
```diff
@@ -0,0 +1,54 @@
...
+    session['access_token'] = access_token
+    return {'status:': '200'}
+  except KeyError:
+    flask.abort(404, 'User does not exist')
+
+
+@oauth.route('/identify')
+def identify():
+  try:
+    username = handshaker.identify(session['access_token'])['username']
```
audiodude, line 42: Instead of doing a network request here (and every time a page is loaded), why don't we store the username in the session?

Then identify() looks like this:

```
def identify():
  username = session.get('username')
  if username is None:
     flask.abort(401, 'Unauthorized')
     return
  return jsonify(username)
```

wp1/web/oauth.py
```diff
@@ -0,0 +1,54 @@
...
+    return jsonify(username)
+  except KeyError:
+    flask.abort(401, 'Unauthorized')
+
+
+@oauth.route('/logout')
+def logout():
+  try:
+    session.pop('access_token')
```
audiodude, line 51: If you're going to use my suggestion about storing the username in the session, then you probably want to have a single key in the session for the access_token and username so that you can pop just one thing here. Maybe:

```
  session['user'] = {'access_token': access_token, 'username': username}
```

wp1/web/oauth.py
```diff
@@ -0,0 +1,54 @@
...
+@oauth.route('/initiate')
+def initiate():
+  redirect, request_token = handshaker.initiate()
+  session['request_token'] = request_token
+  return flask.redirect(redirect)
+
+
+@oauth.route('/complete')
+def complete():
+  try:
```
audiodude, line 29: In general, you should wrap a try/except around the smallest possible snippet of code. Here, I think you're expecting the KeyError from line 32, so really that's all that should be in the try/except.

In honesty, you should really be doing:

```
if 'request_token' not in session:
  flask.abort(404, 'User does not exist')
....rest of method....
```

wp1/web/oauth.py
```diff
@@ -0,0 +1,54 @@
...
+    query_string = str(flask.request.query_string.decode('utf-8'))
+    access_token = handshaker.complete(session['request_token'], query_string)
+    session.pop('request_token')
+    session['access_token'] = access_token
+    return {'status:': '200'}
```
audiodude, line 34: Use `204 NO CONTENT` for responses that are successful but have no body.

wp1/web/oauth.py
```diff
@@ -0,0 +1,54 @@
...
+                          consumer_token)
+except ImportError:
+  print(
+      'No credentials.py file found, Please add your mwoauth credentials in credentials.py'
+  )
+
+
+@oauth.route('/initiate')
+def initiate():
+  redirect, request_token = handshaker.initiate()
```
audiodude, line 22: What if the user is already logged in? Should we really send them back to Wikipedia?

wp1/web/oauth.py
```diff
@@ -0,0 +1,54 @@
+import flask
```
audiodude, line 1: Overall, the pattern of:

```
try:
  do 
  a whole
  bunch
  of things
except OneSpecificError:
  flask.abort(404)
```

Is an anti-pattern. As I mention in other comments, try/except blocks should be limited in scope to the exact lines that can throw the exception. Additionally, there is no need to pop from a dictionary like object if you know the key might not be in it, then wait for an exception to occur. Instead, just use tests like `in` or access like `.get(key_name)`. If those "fail" then you can send your error statuses.

wp1/web/oauth.py
```diff
@@ -0,0 +1,54 @@
...
+    return {'status:': '200'}
+  except KeyError:
+    flask.abort(404, 'User does not exist')
+
+
+@oauth.route('/identify')
+def identify():
+  try:
+    username = handshaker.identify(session['access_token'])['username']
+    return jsonify(username)
```
audiodude, line 43: We might someday need to return other information about the user. Instead of returning a "naked" string here, try using a simple dictionary, like:

```
jsonify({'username': username})
```

That way, if the user data changes, the code that retrieves the username from the backend won't have to. (You'll also need to update the Vue app of course.)

audiodude (changes requested):
Looking good, just a few things to clean up. Please remember not to force push next time, thanks!

audiodude:
Please also make sure tests are passing, and I think you need to write some tests for your OAuth endpoints.

wp1/credentials.py.example
mahakporwal02: Should I add these comments
```
To get your random secret key for production use the following script
python3
import os
os. urandom(24)
```

wp1-frontend/src/App.vue
mahakporwal02: @audiodude Can I do all the UI changes in seperate PR?

mahakporwal02:
> Please also make sure tests are passing, and I think you need to write some tests for your OAuth endpoints.

@audiodude I will make two separate commits 
first one involving ui changes and adding authorisation on manual update second involving tests for the new endpoints.
The tests are failing in the main branch, we need to fix that as well.

wp1/credentials.py.example
audiodude: That would work.

audiodude:
> > Please also make sure tests are passing, and I think you need to write some tests for your OAuth endpoints.
> 
> @audiodude I will make two separate commits
> first one involving ui changes and adding authorisation on manual update second involving tests for the new endpoints.

I agree that requiring login for manual update should be its own PR. However the general rule should be to add tests for code in the same PR that adds the code.

> The tests are failing in the main branch, we need to fix that as well.
I checked, and I believe the tests are passing on main: https://github.com/openzim/wp1/actions/runs/886353197

That's the last CI that ran on main, when I merged my dev-db branch.

audiodude:
It looks like your test errors are related to the ENV variable not being defined in certain places: https://github.com/openzim/wp1/pull/345/checks?check_run_id=2732274725#step:8:724

wp1/web/oauth_test.py
```diff
@@ -1,47 +1,53 @@
 from wp1.web.app import create_app
-from wp1.web.oauth import identify
 from wp1.web.base_web_testcase import BaseWebTestcase
-from unittest import mock
+import unittest
 
-from unittest.mock import patch
+try:
+  # The credentials module isn't checked in and may be missing
```
audiodude, line 6: Actually, in unit tests the credentials module is _always_ missing. You should design your tests so that they work without them, or in some cases with the module mocked out.

wp1/web/oauth_test.py
```diff
@@ -1,47 +1,53 @@
...
 class IdentifyTest(BaseWebTestcase):
 
   def test_identify(self):
     self.app = create_app()
     self.app.testing = True
-    with self.override_db(self.app), self.app.test_client() as c:
-      with c.session_transaction() as sess:
+    with self.app.test_client() as client:
+      rv = client.get('/v1/oauth/identify')
+      self.assertEqual('401 UNAUTHORIZED', rv.status)
```
audiodude, line 25: These should be two separate tests:

1. When there is nothing in the session, it returns 401
2. When the session has a 'user' key it returns the correct username.

wp1-frontend/src/App.vue
audiodude: What other UI changes are there?

wp1/credentials.py.example
```diff
@@ -50,6 +50,24 @@ CREDENTIALS = {
...
+        # To register the app go to 'https://meta.wikimedia.org/wiki/Special:OAuthConsumerRegistration/propose'
+        'MWOAUTH': {
+            'consumer_key': '',
+            'consumer_secret': ''
+        },
+
+        # Secret key for session.
+        # To get your random secret key for production use the following script
+        # python3
+        # import os
```
audiodude, line 64: I think you're missing a step here.

wp1/web/app.py
```diff
@@ -17,10 +18,10 @@
 
 try:
   # The credentials module isn't checked in and may be missing
-  from wp1.credentials import ENV
+  from wp1.credentials import ENV, CREDENTIALS
 except ImportError:
-  print(
-      'No credentials.py file found, Development overlay will not be enabled.')
+  print('''No credentials.py file found, Development overlay
```
audiodude, line 23: nit: Instead of a triple quoted string, use implicit string joining to avoid line length limits. Like this:

```
print('No credentials.py file found. Development overlay will '
         'not be enabled.')
```

wp1/web/app.py
```diff
@@ -93,4 +109,13 @@ def swagger_api_docs_yml():
 
   app.register_blueprint(projects, url_prefix='/v1/projects')
   app.register_blueprint(articles, url_prefix='/v1/articles')
+
+  if CREDENTIALS is None or CREDENTIALS[ENV].get(
```
audiodude, line 113:
```
missing_credentials = CREDENTIALS is None or ENV is None
if not missing_credentials:
  mwoauth = CREDENTIALS.get(ENV, {}).get(MWOAUTH, {})
if missing_credentials or mwoauth.get('consumer_key') is None or mwoauth.get('consumer_secret') is None:
  ....
```

wp1/web/oauth_test.py
```diff
@@ -0,0 +1,50 @@
...
+except ImportError:
+  ENV = None
+  CREDENTIALS = None
+
+skip = bool(CREDENTIALS is None or CREDENTIALS[ENV].get('MWOAUTH') is None or
+            CREDENTIALS[ENV]['MWOAUTH'].get('consumer_key') is None or
+            CREDENTIALS[ENV]['MWOAUTH'].get('consumer_secret') is None)
+
+
+@unittest.skipIf(skip, 'Cannot find Credentials.py')
```
audiodude, line 17: Like mentioned previously, this is incorrect because credentials.py will never be checked in for unit tests.

audiodude (changes requested):
The tests look good, but you need to handle the case where CREDENTIALS are missing. I think there are some db tests that do this by mocking the CREDENTIALS variable in the module under test.

wp1/credentials.py.example
mahakporwal02: Regarding UI changes, do you want me to change anything else?

mahakporwal02:
> What other UI changes are there?
Do you have any changes in mind?

wp1/base_db_test.py
```diff
@@ -120,6 +120,15 @@ def get_test_connect_creds():
               'host': 'localhost',
               'port': 55555
           },
+          'API': {},
```
audiodude, line 123: This is the right idea, but the oauth creds don't really belong in the base_db_test file.

wp1/web/oauth_test.py
```diff
@@ -0,0 +1,54 @@
+import unittest
+from wp1.web.app import create_app
+from wp1.environment import Environment
+from wp1.base_db_test import get_test_connect_creds
+from unittest.mock import patch
+
+
+class IdentifyTest(unittest.TestCase):
+
+  @patch('wp1.web.app.CREDENTIALS', get_test_connect_creds())
```
audiodude, line 10: Instead, just put the oauth credentials in a constant object (TEST_OAUTH_CREDS) and use that constant instead of get_test_connect_creds().

audiodude (changes requested):
You've definitely got the idea now, good work. I think you still need additional test code though, because I see a lot of Codecov warnings that lines are not tested.

mahakporwal02:
> You've definitely got the idea now, good work. I think you still need additional test code though, because I see a lot of Codecov warnings that lines are not tested.

I have to mock out the oauth parts and I think Codecov warnings are related to that?

audiodude:
> there

There will be changes but none that are applicable to this PR

audiodude:
> > What other UI changes are there?
> 
> Do you have any changes in mind?

There will be other changes of course, but none that I can think of as part of this PR

audiodude:
> > You've definitely got the idea now, good work. I think you still need additional test code though, because I see a lot of Codecov warnings that lines are not tested.
> 
> I have to mock out the oauth parts and I think Codecov warnings are related to that?

Right, the codecov warnings are telling you which parts of the code are untested. I'm glad you have plans to test those parts.

mahakporwal02:
So should I just change code for the above two comments (regarding the test_ oauth_creds object) for this PR?

audiodude:
> So should I just change code for the above two comments (regarding the test_ oauth_creds object) for this PR?

Yes. If I make a comment with 'requested changes' as the review, it means you should make those changes

wp1/web/oauth.py
```diff
@@ -0,0 +1,59 @@
+import flask
+from flask import jsonify, session
+from mwoauth import ConsumerToken, Handshaker
+
+oauth = flask.Blueprint('oauth', __name__)
+
+try:
```
audiodude, line 7: Do you still need this try/except block? If the credentials are missing, won't this module never get imported?

mahakporwal02: Yes we will need that. 
In case of tests (mock) it will give error in import.

wp1/web/oauth.py
```diff
@@ -13,11 +13,12 @@
                           consumer_token)
   client_url = CREDENTIALS[ENV]['CLIENT_URL']
 except ImportError:
-  print(
-      'No credentials.py file found, Please add your mwoauth credentials in credentials.py'
-  )
+  print('''No credentials.py file found,
```
audiodude, line 16: I believe we talked about this before, but prefer implicit string concatenation over triple quoted strings, for line length purposes:

```
print('No credentials.py file found, Please add your '
         'mwoauth credentials in credentials.py')

audiodude (changes requested):
Just one nit on the triple quoted string, but otherwise the new tests look really good!

wp1/web/oauth.py
mahakporwal02: Should I squash the commits?

audiodude: No, I don't think so. I personally believe that squashing is time consuming and error prone. We can merge just like this.

audiodude (approved):
LGTM

PR #353, opened 2021-06-12
added authorisation on manual updates

### LINKED ISSUE #350: Make manual updates require login
Since the manual update can be expensive, it should require at the least a valid Wikipedia account.

Also, putting this behind login will let us develop and test out methods for redirecting to login from a page and redirecting back.

mahakporwal02:
@audiodude This fixes #350


wp1/web/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+  #  The credentials module isn't checked in and may be missing
+  from wp1.credentials import ENV, CREDENTIALS
+except ImportError:
+  print('No credentials.py file found, Please add your '
+        'mwoauth credentials in credentials.py')
+  ENV = None
+  CREDENTIALS = None
+
+
+def authenticate(f):
```
audiodude, line 15: This is a great helper decorator.

audiodude (changes requested):
Great work on the backend, but I think the frontend needs some changes. We should not hide the manual update feature from non logged in users, and I'm not sure we need Vuex.

wp1-frontend/src/components/UpdatePage.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <div v-if="isLoggedIn">
```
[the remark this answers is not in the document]
audiodude: Additionally, we need a mechanism so that once you log in you are redirected back to the manual update page. Some kind of `&next=/#/update` parameter that gets stored in the session.

wp1-frontend/src/components/UpdatePage.vue
```diff
@@ -61,6 +61,17 @@
       </div>
     </div>
   </div>
+  <div v-else>
+    <h2 class="content-style" style="padding-top: 250px">
```
audiodude, line 65: Instead of an explicit padding, use the bootstrap "p-1" "p-2" etc classes. Here and below.

wp1-frontend/src/components/UpdatePage.vue
```diff
@@ -79,7 +90,8 @@ export default {
       pollingId: 0,
       progressCurrent: null,
       progressTotal: null,
-      jobStatusEnum: null
+      jobStatusEnum: null,
+      loginInitiateUrl: `${process.env.VUE_APP_API_URL}/oauth/initiate?next=update`
```
audiodude, line 94: Let's go with just `loginUrl`.

wp1-frontend/src/components/UpdatePage.vue
```diff
@@ -193,4 +211,9 @@ export default {
 };
 </script>
 
-<style></style>
+<style scoped>
+.content-style {
+  display: flex;
```
audiodude, line 216: You shouldn't have to manually `display: flex` anything, you should be able to use Bootstrap rows and columns to get the desired effect.

wp1/web/oauth.py
```diff
@@ -47,6 +50,8 @@ def complete():
           'sub': identity['sub']
       }
   }
+  if session.get('next_path'):
```
audiodude, line 53: You should use `session.pop` here, because we want to consume the next path value and not have it "hang around" potentially for the next login.

wp1/web/oauth_test.py
```diff
@@ -109,6 +109,7 @@ def test_logout_authorized_user(self):
     with self.app.test_client() as client:
       with client.session_transaction() as sess:
         sess['user'] = self.USER
+        sess['next_path'] = '/'
```
audiodude, line 112: There should be a new test for when the user requests initiate and has a `next=` param. There should also be a test for when the user hits `complete` and there is a `next_path` in the session.

audiodude (changes requested):
A few styling changes: we really want to take advantage of Bootstrap and not have custom styles. Also I think you need a few more tests, but besides that this looks really good!

wp1-frontend/src/components/UpdatePage.vue
```diff
@@ -61,6 +61,17 @@
...
     </div>
   </div>
+  <div v-else>
+    <h2 class="row justify-content-center pt-5">
+      Please Log In To Continue
+    </h2>
+    <a
+      class="row justify-content-center"
+      :href="this.loginUrl"
+      style="padding-top: 10px;"
```
audiodude, line 71: Same thing here, use pt-* instead

wp1/web/oauth.py
```diff
@@ -47,6 +50,10 @@ def complete():
           'sub': identity['sub']
       }
   }
+  next_path = session.get('next_path')
+  session.pop('next_path')
```
audiodude, line 54: Doesn't pop return the value? So shouldn't you just be able to do `next_path = session.pop('next_path')`?

audiodude (changes requested):
Looking good, just a few small issues.

wp1-frontend/src/components/UpdatePage.vue
```diff
@@ -61,6 +61,17 @@
...
+      ><button>Login</button></a
```
audiodude, line 72: Also, make sure to use the Bootstrap button classes: `type="button"` and `class="btn btn-primary"`.

https://getbootstrap.com/docs/4.0/components/buttons/

mahakporwal02: should I add these classes to other login button also?

audiodude: Yes, the login button should have `btn-primary` but the logout button should be `btn-secondary`.

mahakporwal02:
> Remove print statement

Sorry for that

audiodude (approved):
LGTM

PR #358, opened 2021-06-14
Skeleton of selection pages, with Secondary Nav Bar

### LINKED ISSUE #357: Create Selection page skeletons with secondary navbar
There will eventually be a link in the main navbar for "Selections" which will go to wp1.openzim.org/#/selections. We should start building the pages at that location but not link it to the main navbar for now.

Currently there are two pages under /selections:

1. /selections/user  --  A list of the users selections
2. /selections/simple/new -- A place to create a new "simple list" as described in #336.

For the purposes of completing this issue, each of these pages can have a simple:

```
<h1>Coming soon -- User selections</h1>
or
<h1>Coming soon -- New simple list creation</h1>
```

as their content. The important thing is that there is a secondary navbar that displays which page the user is on, and that clicking on the navbar takes the user to the correct page. Also there should be new .vue files for each of these pages, obviously, that will be filled in later.

mahakporwal02:
@audiodude This fixes #357

### DIFF
--- a/wp1-frontend/src/App.vue
+++ b/wp1-frontend/src/App.vue
@@ -70,17 +70,14 @@
         </div>
       </div>
     </nav>
-    <SecondaryNav></SecondaryNav>
-    <div id="app" class="container">
+    <div id="app">
       <router-view></router-view>
     </div>
   </div>
 </template>
 
 <script>
-import SecondaryNav from './components/SecondaryNav.vue';
 export default {
-  components: { SecondaryNav },
   name: 'app',
   data: function() {
     return {
--- a/wp1-frontend/src/App.vue
+++ b/wp1-frontend/src/App.vue
@@ -70,6 +70,7 @@
         </div>
       </div>
     </nav>
+    <SecondaryNav></SecondaryNav>
     <div id="app" class="container">
       <router-view></router-view>
     </div>
@@ -77,7 +78,9 @@
 </template>
 
 <script>
+import SecondaryNav from './components/SecondaryNav.vue';
 export default {
+  components: { SecondaryNav },
   name: 'app',
   data: function() {
     return {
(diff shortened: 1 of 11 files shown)


wp1-frontend/src/main.js
```diff
@@ -79,6 +82,27 @@ const routes = [
         ' and ' +
         route.params.projectNameB
     }
+  },
+  {
+    path: '/selection/',
+    component: SecondaryNav,
```
audiodude, line 88: This isn't right. This will just display the secondary nav component as the main component when the URL is `/selection`. I think we don't actually need a base `/selection` route right now.

wp1-frontend/src/App.vue
```diff
@@ -70,14 +70,17 @@
         </div>
       </div>
     </nav>
+    <SecondaryNav></SecondaryNav>
```
audiodude, line 73: The secondary nav should not display on all pages, only the selection pages.

audiodude (changes requested):
You've got the basic idea and the parts are correct, but the secondary nav should not display on all pages.

wp1-frontend/src/components/SecondaryNav.vue
```diff
@@ -0,0 +1,55 @@
...
+        <ul class="navbar-nav mr-auto">
+          <li
+            :class="
+              'nav-item ' +
+                (this.$route.path.startsWith('/selection/myLists')
+                  ? 'active'
+                  : '')
+            "
+          >
+            <router-link class="nav-link" to="/selection/myLists"
```
audiodude, line 27: The URL should be `/selection/user` not `/selection/myLists`

wp1-frontend/src/components/SecondaryNav.vue
```diff
@@ -0,0 +1,55 @@
...
+              >My Lists</router-link
+            >
+          </li>
+          <li
+            :class="
+              'nav-item ' +
+                (this.$route.path.startsWith('/selection/createSimpleList')
```
audiodude, line 34: The URL should be `/selection/lists/simple/new`

audiodude (approved):
LGTM

PR #360, opened 2021-06-17
Add database migrations using YoYo framework

### LINKED ISSUE #340: Create users table in WP1 database
For user login (see #339), we will need to store the user's wikipedia user id in a database table, so that we can associate it with ownership of whatever selection lists the user makes. We probably want to also store the user's username, in case we need to display it when they are not logged in.

### LINKED ISSUE #341: Create method for migrating WP1 database
We should try integrating [this library](https://ollycope.com/software/yoyo/latest/) for database migrations.

It will provide a nice way of keeping track of the state of the databases in both development and production.

audiodude:
Migrations are simple .sql or .py files that specify how to apply and how to rollback a change to the database schema. Instructions on how to run the Development database migrations have been added to the main README.md. There are additional instructions on how to migration the production database as well.

More information at https://ollycope.com/software/yoyo/latest/

Additionally, the first migration, adding the `users` table, is supplied.

Fixes #341. Fixes #340.


db/README.md
```diff
@@ -0,0 +1,15 @@
...
+lives on the production box at `/data/wp1bot/db/yoyo.ini` (because it contains sensitive
+database credentials).
+
+All migrations are in the `migrations` folder. See the YoYo docs for more details on how
+to add a new migration.
+
+See the sections in the main README on:
+
+1. Migrating the dev database.
+1. Updating production (section on running database migrations).
```
mahakporwal02, line 15: Just a small typo.
Should be `2.`

mahakporwal02 (approved):
Just a small typo else everything LGTM.

db/README.md
audiodude: Actually in markdown you use:

1. Numbered lists
1. Like this
1. So if you have to insert or delete an item
1. You don't need to renumber anything:

Screen Shot 2021-06-17 at 2 58 30 PM

PR #364, opened 2021-06-18
store users in database

### LINKED ISSUE #362: Store users in the database when they log in
When a user logs in using OAuth, we should store an entry for them in the database, so that later selections can be linked to their user id (`sub`). #360 added a schema for a `users` table, which is simply id:INTEGER, username:VARCHAR(255). We should fill this out on OAuth logins, potentially with an `INSERT...ON DUPLICATE KEY UPDATE` statement, so that we don't have to first query the database.

mahakporwal02:
This fixes #362

### DIFF
--- a/wp1/web/oauth.py
+++ b/wp1/web/oauth.py
@@ -1,6 +1,7 @@
 import flask
 from flask import jsonify, session
 from mwoauth import ConsumerToken, Handshaker
+from wp1.web.db import get_db
 
 oauth = flask.Blueprint('oauth', __name__)
 
@@ -50,6 +51,14 @@ def complete():
           'sub': identity['sub']
       }
   }
+  wp10db = get_db('wp10db')
+  with wp10db.cursor() as cursor:
+    cursor.execute('''INSERT INTO users (u_id,u_username)
+                    VALUES ({id},"{username}")
+                    ON DUPLICATE KEY UPDATE u_id={id}, u_username="{username}";'''
+                   .format(username=session['user']['identity'].get('username'),
+                           id=session['user']['identity'].get('sub')))
+    wp10db.commit()
   next_path = session.pop('next_path')
   if next_path:
     return flask.redirect(f"{homepage_url}{str(next_path)}")
--- a/wp1/web/oauth.py
+++ b/wp1/web/oauth.py
@@ -58,7 +58,7 @@ def complete():
     cursor.execute(
         '''INSERT INTO users (u_id, u_username)
                       VALUES (%(u_id)s, %(u_username)s)
-                      ON DUPLICATE KEY UPDATE u_id= %(u_id)s, u_username= %(u_username)s;''',
+                      ON DUPLICATE KEY UPDATE u_id= %(u_id)s, u_username= %(u_username)s''',
         {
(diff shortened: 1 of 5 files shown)


wp1/web/oauth.py
```diff
@@ -50,6 +51,14 @@ def complete():
           'sub': identity['sub']
       }
   }
+  wp10db = get_db('wp10db')
+  with wp10db.cursor() as cursor:
+    cursor.execute('''INSERT INTO users (u_id,u_username)
+                    VALUES ({id},"{username}")
+                    ON DUPLICATE KEY UPDATE u_id={id}, u_username="{username}";'''
+                   .format(username=session['user']['identity'].get('username'),
```
audiodude, line 59: This is not the right way to do the SQL execution, and is vulnerable to SQL injection attacks. It does not properly escape the data coming in from the session.

Look at https://github.com/openzim/wp1/blob/main/wp1/logic/rating.py#L211 for a better example. The second argument to the execute method is the data to bind. If the second argument is a tuple, you can use `%s` markers in your SQL. If the second argument is a dict, you can use %{name}s markers.

audiodude (changes requested):
This looks good, but you need to execute your SQL properly.

wp1/web/oauth.py
audiodude: Note that CodeFactor is warning you about the same thing.

audiodude: Sorry the syntax for dictionaries is `%(dictionary_key)s` with parens not brackets.

wp1/web/oauth.py
```diff
@@ -50,6 +53,16 @@ def complete():
           'sub': identity['sub']
       }
   }
+  wp10db = get_db('wp10db')
+  with wp10db.cursor() as cursor:
+    cursor.execute(
+        '''INSERT INTO users (u_id, u_username)
+                      VALUES (%(u_id)s, "%(u_username)s")
```
audiodude, line 60: You don't want quotes around `$(u_username)s`. By properly binding the data using a dictionary, the SQL library will automatically add the quotes. Both here and in the UPDATE clause.

audiodude (changes requested):
Good job adding a User model, a few more comments on how to insert the data.

wp1/web/oauth.py
```diff
@@ -57,11 +57,12 @@ def complete():
   with wp10db.cursor() as cursor:
     cursor.execute(
         '''INSERT INTO users (u_id, u_username)
-                      VALUES (%(u_id)s, "%(u_username)s")
-                      ON DUPLICATE KEY UPDATE u_id= %(u_id)s, u_username= "%(u_username)s";''',
-        attr.asdict(
-            User(u_id=session['user']['identity'].get('sub'),
-                 u_username=session['user']['identity'].get('username'))))
+                      VALUES (%(u_id)s, %(u_username)s)
+                      ON DUPLICATE KEY UPDATE u_id= %(u_id)s, u_username= %(u_username)s;''',
```
audiodude, line 61: minor nit: you also don't need the semicolon in the statement.

audiodude (approved):
LGTM

PR #369, opened 2021-06-24
added UI for creating list

### LINKED ISSUE #367: Build simple list creation UI
Here is a mockup of the simple list creation UI, for #336.

Screen Shot 2021-06-20 at 2 59 21 PM

The project list is populated from the endpoint specified in #366. A project must be selected, but should default to en.wikipedia.org.

The list name is required.

The list of articles can contain as few as one entry. The backend will be responsible for validating this list otherwise.

mahakporwal02:
This fixes #367 
Shouldn't this be under the `Create Simple List` page (currently in mock-up it was under My Lists)?

### DIFF
--- a/wp1-frontend/cypress/integration/createList_page.spec.js
+++ b/wp1-frontend/cypress/integration/createList_page.spec.js
@@ -1,8 +1,11 @@
 /// <reference types="Cypress" />
 
 describe('the createList page', () => {
-  it('successfully loads', () => {
+  beforeEach(() => {
     cy.intercept('v1/sites/', { fixture: 'sites.json' });
+  });
+
+  it('successfully loads', () => {
     cy.visit('/#/selection/lists/simple/new');
   });
 
@@ -12,20 +15,28 @@ describe('the createList page', () => {
     cy.get('select').contains('en.wikipedia.org');
   });
 
-  it('displays list name', () => {
+  it('list name validation', () => {
     cy.get('#saveListButton').click();
-    cy.on('window:alert', txt => {
-      expect(txt).to.contains('Please fill in this field');
-    });
+    cy.get('#listName').contains('Please provide a valid List Name.');
     cy.get('#listName')
       .click()
       .type('My List');
   });
 
-  it('displays textbox', () => {
+  it('textbox item validation', () => {
     cy.get('#saveListButton').click();
-    cy.on('window:alert', txt => {
-      expect(txt).to.contains('Please fill in this field');
-    });
+    cy.get('#items').contains('Please provide valid items.');
+  });
+
+  it('list name validation on blur', () => {
+    cy.visit('/#/selection/lists/simple/new');
+    cy.get('#listName > .form-control').click();
+    cy.get('#listName').contains('Please provide a valid List Name.');
+  });
+
+  it('textbox item validation on blur', () => {
+    cy.visit('/#/selection/lists/simple/new');
+    cy.get('#items > .form-control').click();
+    cy.get('#items').contains('Please provide valid items.');
   });
 });
--- a/wp1-frontend/cypress/integration/createList_page.spec.js
+++ b/wp1-frontend/cypress/integration/createList_page.spec.js
@@ -18,9 +18,6 @@ describe('the createList page', () => {
   it('list name validation', () => {
     cy.get('#saveListButton').click();
     cy.get('#listName').contains('Please provide a valid List Name.');
-    cy.get('#listName')
-      .click()
-      .type('My List');
   });
 
   it('textbox item validation', () => {
--- a/wp1-frontend/cypress/integration/createList_page.spec.js
+++ b/wp1-frontend/cypress/integration/createList_page.spec.js
@@ -1,0 +1,25 @@
+/// <reference types="Cypress" />
+
+describe('the createList page', () => {
+  it('successfully loads', () => {
+    cy.visit('/#/selection/lists/simple/new');
+    cy.get('nav').contains('Create Simple List');
+  });
+
+  it('displays wiki projects', () => {
+    cy.get('select').select('ab.wikipedia.org');
+  });
+
+  it('displays List Name', () => {
+    cy.get('#listName')
+      .click()
+      .type('My List');
+  });
+
+  it('displays textbox', () => {
+    cy.get('textarea')
+      .click()
+      .type('Items');
+    cy.get('#saveListButton').click();
+  });
+});
--- a/wp1-frontend/cypress/integration/createList_page.spec.js
+++ b/wp1-frontend/cypress/integration/createList_page.spec.js
@@ -2,24 +2,30 @@
 
 describe('the createList page', () => {
   it('successfully loads', () => {
+    cy.intercept('v1/sites/', { fixture: 'sites.json' });
     cy.visit('/#/selection/lists/simple/new');
-    cy.get('nav').contains('Create Simple List');
   });
 
   it('displays wiki projects', () => {
-    cy.get('select').select('ab.wikipedia.org');
+    cy.get('select').contains('aa.wikipedia.org');
+    cy.get('select').contains('en.wiktionary.org');
+    cy.get('select').contains('en.wikipedia.org');
   });
 
-  it('displays List Name', () => {
+  it('displays list name', () => {
+    cy.get('#saveListButton').click();
+    cy.on('window:alert', txt => {
+      expect(txt).to.contains('Please fill in this field');
+    });
     cy.get('#listName')
       .click()
       .type('My List');
   });
 
   it('displays textbox', () => {
-    cy.get('textarea')
-      .click()
-      .type('Items');
     cy.get('#saveListButton').click();
+    cy.on('window:alert', txt => {
+      expect(txt).to.contains('Please fill in this field');
+    });
   });
 });
--- a/wp1-frontend/src/components/CreateSimpleList.vue
+++ b/wp1-frontend/src/components/CreateSimpleList.vue
@@ -4,50 +4,61 @@
       <SecondaryNav></SecondaryNav>
     </div>
     <div class="row">
-      <form>
-        <div class="col-lg-6 col-md-9 m-4">
-          <div>
-            <div class="m-4" style="font-weight: 550;">
-              Use this tool to create an article selection list for the
-              Wikipedia project of your choice. Your selection will be saved in
-              public cloud storage and can be accessed through URLs that will be
-              provided once it has been saved.
-            </div>
-            <div class="m-4">
-              <label for="Project">Project</label>
-              <select class="custom-select my-list">
-                <option selected>en.wikipedia.org</option>
-                <option v-for="item in wikiProjects" v-bind:key="item">
-                  {{ item }}
-                </option>
-              </select>
-            </div>
-            <div class="m-4">
-              <label for="listName">List Name</label>
-              <input
-                id="listName"
-                placeholder="My List"
-                class="form-control my-list"
-                required="true"
-              />
-            </div>
-            <div class="form-group m-4">
-              <label for="Items">Items</label>
-              <textarea
-                :placeholder="
-                  'Eiffel_Tower\nStatue_of_Liberty\nFreedom_Monument_(Baghdad)\nGeorge-Étienne_Cartier_Monument'
-                "
-                class="form-control my-list"
-                rows="13"
-                required="true"
-              ></textarea>
-            </div>
-            <button id="saveListButton" type="submit" class="btn-primary ml-4">
-              Save List
-            </button>
-          </div>
+      <div class="col-lg-6 col-md-9 m-4">
+        <div class="m-4" style="font-weight: 550;">
+          Use this tool to create an article selection list for the Wikipedia
+          project of your choice. Your selection will be saved in public cloud
+          storage and can be accessed through URLs that will be provided once it
+          has been saved.
         </div>
-      </form>
+        <form class="needs-validation" novalidate>
+          <div class="m-4">
+            <label for="Project">Project</label>
+            <select class="custom-select my-list">
+              <option selected>en.wikipedia.org</option>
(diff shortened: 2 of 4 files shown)


audiodude (changes requested):
Yes, it should be under CreateSimpleList, that was a bug in the mockup.

Perhaps this wasn't clear, but you should include the following:

1. The logic for fetching the sitematrix that you added in the previous PR and populating the `<select>` element. The initial list on page load can just be a hardcoded `['en.wikipedia.org']`
2. Validation logic for the form, as described in the mock issue.
3. Cypress tests.

Thanks!

mahakporwal02:
> Yes, it should be under CreateSimpleList, that was a bug in the mockup.
> 
> Perhaps this wasn't clear, but you should include the following:
> 
> 1. The logic for fetching the sitematrix that you added in the previous PR and populating the `<select>` element. The initial list on page load can just be a hardcoded `['en.wikipedia.org']`
> 2. Validation logic for the form, as described in the mock issue.
> 3. Cypress tests.
> 
> Thanks!

Sorry I didn't see you have already reviewed the PR thats why I force pushed.

audiodude:
> Sorry I didn't see you have already reviewed the PR thats why I force pushed.

Actually you should definitely force push, once you've transferred the changes to CreateSimpleList.vue. We don't need a commit that makes the changes to MyLists.vue.

mahakporwal02:
> > Sorry I didn't see you have already reviewed the PR thats why I force pushed.
> 
> Actually you should definitely force push, once you've transferred the changes to CreateSimpleList.vue. We don't need a commit that makes the changes to MyLists.vue.

Also, can you please elaborate what should be done in validation logic?

audiodude:
> Also, can you please elaborate what should be done in validation logic?
> Should I just add checks for spaces?

The rigorous validation of the list will be handled by the backend, because it will potentially involve checking that the articles actually exist in the specified project. Otherwise, the validation is as described in #367: The name cannot be empty and the list cannot be empty.

kelson42:
The article existence check on the backend is an expensive operation to run and not totally trivial to implement efficiently. I don't know if its implementation is part of this ticket, but I would differentiate it from a pure syntaxic check and recommend to open a dedicated ticket for it.

wp1-frontend/cypress/integration/createList_page.spec.js
```diff
@@ -0,0 +1,25 @@
+/// <reference types="Cypress" />
+
+describe('the createList page', () => {
+  it('successfully loads', () => {
+    cy.visit('/#/selection/lists/simple/new');
+    cy.get('nav').contains('Create Simple List');
```
audiodude, line 6: This will be true of several pages besides Create Simple List

wp1-frontend/cypress/integration/createList_page.spec.js
```diff
@@ -0,0 +1,25 @@
...
+  });
+
+  it('displays wiki projects', () => {
+    cy.get('select').select('ab.wikipedia.org');
```
audiodude, line 10: We need to mock out the request for this, because we don't want to call the actual Mediawiki API in CI.

wp1-frontend/cypress/integration/createList_page.spec.js
```diff
@@ -0,0 +1,25 @@
...
+  });
+
+  it('displays List Name', () => {
```
audiodude, line 13: I think what you actually want is the opposite of this: that when you don't have a list name, you get an error. Also in general, a test that doesn't make any assertions about the state of the page is not a good test.

wp1-frontend/cypress/integration/createList_page.spec.js
```diff
@@ -0,0 +1,25 @@
...
+    cy.get('#listName')
+      .click()
+      .type('My List');
+  });
+
+  it('displays textbox', () => {
```
audiodude, line 19: Same thing, this doesn't test anything.

wp1-frontend/src/components/CreateSimpleList.vue
```diff
@@ -1,8 +1,51 @@
...
+              Wikipedia project of your choice. Your selection will be saved in
+              public cloud storage and can be accessed through URLs that will be
+              provided once it has been saved.
+            </div>
+            <div class="m-4">
+              <label for="Project">Project</label>
+              <select class="custom-select my-list">
+                <option selected>en.wikipedia.org</option>
+                <option v-for="item in wikiProjects" v-bind:key="item">
+                  {{ item.replace('https://', '') }}
```
audiodude, line 21: You should perform this transformation in the data fetching function so that the select can simply display the items. Also, if you're going to hardcode en.wikipedia.org, you should remove it from the list so it doesn't appear twice.

wp1-frontend/src/components/CreateSimpleList.vue
```diff
@@ -1,8 +1,51 @@
...
+                id="listName"
+                placeholder="My List"
+                class="form-control my-list"
+                required="true"
+              />
+            </div>
+            <div class="form-group m-4">
+              <label for="Items">Items</label>
+              <textarea
+                placeholder="Article names with either 'Spaces between words' or 'Underscores_between_words'"
```
audiodude, line 37: Change this placeholder to be an appropriate list of items of some sort:

```
placeholder="Eiffel_Tower
Statue_of_Liberty
Freedom_Monument_(Baghdad)
George-Étienne_Cartier_Monument"
```

Note that this list contains both parenthesis and non-ASCII characters

wp1-frontend/src/components/CreateSimpleList.vue
```diff
@@ -11,8 +54,27 @@
 import SecondaryNav from './SecondaryNav.vue';
 export default {
   components: { SecondaryNav },
-  name: 'CreateSimpleList'
+  name: 'MyLists',
+  data: function() {
+    return {
+      wikiProjects: ['en.wikipedia.org']
```
audiodude, line 60: If you've encoded this here, you don't need to hardcode it in the <select> construction.

audiodude (changes requested):
I'm going to download this PR and run it locally to make sure everything looks good. Otherwise, I think your tests need improvement, but the page itself looks pretty good.

wp1-frontend/cypress/integration/createList_page.spec.js
mahakporwal02: So should I remove this for now?

wp1-frontend/src/components/CreateSimpleList.vue
mahakporwal02: I am removing it from the list.
I have to hardcode here because it should be seen as default option in select.

audiodude: This array can be empty, because your HTML contains a literal `<option>` for en.wikipedia.org. That will render no matter what is in the list.

wp1-frontend/cypress/integration/createList_page.spec.js
```diff
@@ -0,0 +1,31 @@
+/// <reference types="Cypress" />
+
+describe('the createList page', () => {
+  it('successfully loads', () => {
+    cy.intercept('v1/sites/', { fixture: 'sites.json' });
```
audiodude, line 5: You need to do this in a `beforeEach` in this describe block. Otherwise, you're not doing it for every test.

wp1-frontend/cypress/integration/createList_page.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    cy.visit('/#/selection/lists/simple/new');
+  });
+
+  it('displays wiki projects', () => {
+    cy.get('select').contains('aa.wikipedia.org');
+    cy.get('select').contains('en.wiktionary.org');
+    cy.get('select').contains('en.wikipedia.org');
+  });
+
+  it('displays list name', () => {
```
audiodude, line 15: The name of this test should be updated, because it's now testing that the validation works.

wp1-frontend/cypress/integration/createList_page.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    cy.get('#saveListButton').click();
+    cy.on('window:alert', txt => {
+      expect(txt).to.contains('Please fill in this field');
+    });
+    cy.get('#listName')
```
audiodude, line 20: Why do you need to type into the text input?

wp1-frontend/src/components/CreateSimpleList.vue
```diff
@@ -1,8 +1,53 @@
...
+                </option>
+              </select>
+            </div>
+            <div class="m-4">
+              <label for="listName">List Name</label>
+              <input
+                id="listName"
+                placeholder="My List"
+                class="form-control my-list"
+                required="true"
```
audiodude, line 31: Although this technically prevents the user from submitting an empty list name, the UI with alerts is not particularly user-friendly. What we would like is a validation function on this page that runs on the `<input>` onBlur, as well as when the save button is clicked. It should change the style of the inputs and display inline error messages, similar to: https://getbootstrap.com/docs/5.0/forms/validation/#custom-styles

Apologies that this wasn't in the mock.

wp1-frontend/src/components/CreateSimpleList.vue
```diff
@@ -11,8 +56,34 @@
...
+  methods: {
+    getWikiProjects: async function() {
+      const response = await fetch(`${process.env.VUE_APP_API_URL}/sites/`);
+      var data = await response.json();
+      this.wikiProjects = data.sites;
+      this.wikiProjects.forEach((element, index) => {
+        if (element === 'https://en.wikipedia.org') {
+          this.wikiProjects.splice(index, 1);
+          return;
+        }
```
audiodude, line 77: I know in Python you get bugs if you modify a list while you're iterating over it. The same is probably true in Javascript? Why not just:

```suggestion
      index = this.wikiProjects.indexOf('https://en.wikipedia.org')
      this.wikiprojects.splice(index, 1);
```

audiodude (changes requested):
Using HTML5 validation works, but we want custom validation with Bootstrap style colors and messages displayed inline. Please implement this and update the tests. Thanks!

wp1-frontend/src/components/CreateSimpleList.vue
audiodude: If you have this list populated on page load, as well as the hardcoded literal `<option>` you get the following:

Screen Shot 2021-06-26 at 10 12 06 AM

The array should be empty.

wp1-frontend/src/components/CreateSimpleList.vue
```diff
@@ -11,8 +67,55 @@
...
+      }
+    }
+  }
 };
 </script>
 
-<style scoped></style>
+<style scoped>
+.my-list {
+  border: 1px solid #000;
```
audiodude, line 119: I don't think this is necessary, let's just use the default Bootstrap styles.

audiodude (changes requested):
Just a couple of nits, but this looks really close!

wp1-frontend/src/components/CreateSimpleList.vue
```diff
@@ -1,8 +1,64 @@
...
   <div>
-    <SecondaryNav></SecondaryNav>
-    <div class="container">
-      <h1>Coming soon -- New simple list creation</h1>
+    <div>
+      <SecondaryNav></SecondaryNav>
+    </div>
+    <div class="row">
+      <div class="col-lg-6 col-md-9 m-4">
+        <div class="m-4" style="font-weight: 550;">
```
audiodude, line 8: This doesn't need to be bold, remove `font-weight` (and generally avoid inline styles).

mahakporwal02:
@audiodude 
Regarding "Still wondering this."
I have asserted that the div with id="listName" contains this text when field is empty.

audiodude:
> @audiodude
> Regarding "Still wondering this."
> I have asserted that the div with id=`listName` contains the text `Please provide a valid List Name.` when field is empty.

But there are no assertions after the typing.

mahakporwal02:
> > @audiodude
> > Regarding "Still wondering this."
> > I have asserted that the div with id=`listName` contains the text `Please provide a valid List Name.` when field is empty.
> 
> But there are no assertions after the typing.
I am a bit confused
So should I add 
`cy.get('#listName').should('contain.text', 'Please provide a valid List Name.')`; instead of
`cy.get('#listName').contains('Please provide a valid List Name.')`;

mahakporwal02:
> > @audiodude
> > Regarding "Still wondering this."
> > I have asserted that the div with id=`listName` contains the text `Please provide a valid List Name.` when field is empty.
> 
> But there are no assertions after the typing.

OK, I get it

audiodude:
> > > @audiodude
> > > Regarding "Still wondering this."
> > > I have asserted that the div with id=`listName` contains the text `Please provide a valid List Name.` when field is empty.
> > 
> > 
> > But there are no assertions after the typing.
> 
> I am a bit confused
> So should I add
> `cy.get('#listName').should('contain.text', 'Please provide a valid List Name.')`; instead of
> `cy.get('#listName').contains('Please provide a valid List Name.')`;

The test has three lines:

```
    cy.get('#listName').contains('Please provide a valid List Name.');
    cy.get('#listName')
      .click()
      .type('My List');
```

On the first line you click the save button. On the second line you make an assertion. Then, on the third line, you type 'My List' into the name box. But that's the final line, there are no assertions after that. So what I'm wondering is: what is the point of typing into the `#listName` box?

audiodude:
> > > @audiodude
> > > Regarding "Still wondering this."
> > > I have asserted that the div with id=`listName` contains the text `Please provide a valid List Name.` when field is empty.
> > 
> > 
> > But there are no assertions after the typing.
> 
> OK, I get it

Awesome! Also in the future, please try to keep discussion about specific code review items in the relevant conversation and off the main thread of the PR. Thanks!

audiodude (approved):
LGTM

PR #378, opened 2021-06-27
added function and tests for validating article names

### LINKED ISSUE #371: Simple list creation: validate list on backend, syntax only
As part of saving a simple list, there should be a backend function that validates/normalizes the items of the list.

This function should:

1. Replace spaces with underscores
2. Resolve percent encoding to UTF-8
3. Resolve URLs in various forms to just the article name
4. Reject article names which have any of the following properties:
    - Contain any [forbidden characters](https://en.wikipedia.org/wiki/Wikipedia:Naming_conventions_(technical_restrictions)#Forbidden_characters)
    - Are longer than 256 bytes long in UTF-8 representation
4. Return a tuple of:
    - a list of valid names, with order preserved
    - a list of invalid names, with order preserved

For the purposes of this function, article names are considered to be one on each line of the input. So `input.split('\n')`.

So, given the following list:

```
Eiffel_Tower
Statue of Liberty
George-%C3%89tienne_Cartier_Monument
https://en.wikipedia.org/wiki/Liberty_(personification)
https://en.wikipedia.org/w/index.php?title=Libertas
https://en.wikipedia.org/w/index.php?title=%E2%88%9E
```

The output should be:

```
Eiffel_Tower
Statue_of_Liberty
George-Étienne_Cartier_Monument
Liberty_(personification)
Libertas
∞
```

Or more literally:
`(['Eiffel_Tower', 'Statue_of_Liberty', ...], [])`

GIven this list:

```
Statue of Liberty
https://en.wikipedia.org/wiki/Liberty#Philosophy
Eiffel_Tower
This is supposedly an article name but it's way too long to actually be one because it contains more than 256 characters which is not allowed in article titles and it just runs on forever, it doesn't even have underscores and looks more like a long paragraph of text than an actual article name.
Not_an_<article_name>
```

The output would be:

`(['Statute_of_Liberty', 'Eiffel_Tower'], ['https://en.wikipedia.org/wiki/Liberty#Philosophy', 'This is supposedly an article name but it's way too long to actually be one because it contains more than 256 characters which is not allowed in article titles and it just runs on forever, it doesn't even have underscores and looks more like a long paragraph of text than an actual article name.', 'Not_an_<article_name>'])`

mahakporwal02:
This fixes #371

### DIFF
--- a/wp1/logic/selection.py
+++ b/wp1/logic/selection.py
@@ -6,21 +6,21 @@ def validate_list(items):
   invalid_article_names = []
   valid_article_names = []
   for item in item_list:
-    item_in_invalid_article_names = False
+    is_valid = False
     item = item.strip().replace(" ", "_")
     decoded_item = urllib.parse.unquote(item)
     len_item = len(decoded_item.encode("utf-8"))
     sets = {"#", "<", ">", "[", "]", "{", "}", "|"}
     if len_item > 256:
       invalid_article_names.append(decoded_item)
-      item_in_invalid_article_names = True
+      is_valid = True
       continue
     for forbiden_character in sets:
       if forbiden_character in decoded_item:
         invalid_article_names.append(decoded_item)
-        item_in_invalid_article_names = True
+        is_valid = True
         break
-    if not item_in_invalid_article_names:
+    if not is_valid:
       article_name = decoded_item.replace(
           "https://en.wikipedia.org/wiki/",
           "").replace("https://en.wikipedia.org/w/index.php?title=", "")
--- a/wp1/logic/selection.py
+++ b/wp1/logic/selection.py
@@ -6,21 +6,21 @@ def validate_list(items):
   invalid_article_names = []
   valid_article_names = []
   for item in item_list:
-    is_valid = False
+    is_valid = True
     item = item.strip().replace(" ", "_")
     decoded_item = urllib.parse.unquote(item)
     len_item = len(decoded_item.encode("utf-8"))
     sets = {"#", "<", ">", "[", "]", "{", "}", "|"}
     if len_item > 256:
       invalid_article_names.append(decoded_item)
-      is_valid = True
+      is_valid = False
       continue
     for forbiden_character in sets:
       if forbiden_character in decoded_item:
         invalid_article_names.append(decoded_item)
-        is_valid = True
+        is_valid = False
         break
-    if not is_valid:
+    if is_valid:
       article_name = decoded_item.replace(
           "https://en.wikipedia.org/wiki/",
           "").replace("https://en.wikipedia.org/w/index.php?title=", "")
--- a/wp1/logic/selection.py
+++ b/wp1/logic/selection.py
@@ -1,0 +1,28 @@
+import urllib.parse
+
+
+def validate_list(items):
+  item_list = items.split("\n")
+  invalid_article_names = []
+  valid_article_names = []
+  for item in item_list:
+    item_in_invalid_article_names = False
+    item = item.strip().replace(" ", "_")
+    decoded_item = urllib.parse.unquote(item)
+    len_item = len(decoded_item.encode("utf-8"))
+    sets = {"#", "<", ">", "[", "]", "{", "}", "|"}
+    if len_item > 256:
+      invalid_article_names.append(decoded_item)
+      item_in_invalid_article_names = True
+      continue
+    for forbiden_character in sets:
+      if forbiden_character in decoded_item:
+        invalid_article_names.append(decoded_item)
+        item_in_invalid_article_names = True
+        break
+    if not item_in_invalid_article_names:
+      article_name = decoded_item.replace(
+          "https://en.wikipedia.org/wiki/",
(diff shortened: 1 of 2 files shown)


wp1/logic/selection_test.py
```diff
@@ -0,0 +1,19 @@
+import unittest
+from wp1.logic import selection
+
+
+class ValidateArticleNameTest(unittest.TestCase):
+
+  items = "Eiffel_Tower \nStatue of Liberty\nThis is supposedly an article name but its way too long to actually be one because it contains more than 256 characters which is not allowed in article titles and it just runs on forever, it doesnt even have underscores and looks more like a long paragraph of text than an actual article name.\nNot_an_<article_name>\nhttps://en.wikipedia.org/wiki/Liberty#Philosophy\nhttps://en.wikipedia.org/wiki/Liberty_(personification)\nhttps://en.wikipedia.org/w/index.php?title=Libertas\nhttps://en.wikipedia.org/w/index.php?title=%E2%88%9E\nGeorge-%C3%89tienne_Cartier_Monument"
```
audiodude, line 7: I think this will be easier to read as a triple quoted string with literal line breaks.

wp1/logic/selection_test.py
```diff
@@ -0,0 +1,19 @@
...
+
+  def test_validate_items(self):
+    item_list = ([
```
audiodude, line 10: nit: You should call this `expected`.

wp1/logic/selection_test.py
```diff
@@ -0,0 +1,19 @@
...
+        'Eiffel_Tower', 'Statue_of_Liberty', 'Liberty_(personification)',
+        'Libertas', '∞', 'George-Étienne_Cartier_Monument'
+    ], [
+        'This_is_supposedly_an_article_name_but_its_way_too_long_to_actually_be_one_because_it_contains_more_than_256_characters_which_is_not_allowed_in_article_titles_and_it_just_runs_on_forever,_it_doesnt_even_have_underscores_and_looks_more_like_a_long_paragraph_of_text_than_an_actual_article_name.',
+        'Not_an_<article_name>',
+        'https://en.wikipedia.org/wiki/Liberty#Philosophy'
+    ])
+    article_names = selection.validate_list(self.items)
```
audiodude, line 18: nit: And call this `actual`

wp1/logic/selection.py
```diff
@@ -0,0 +1,28 @@
+import urllib.parse
+
+
+def validate_list(items):
+  item_list = items.split("\n")
+  invalid_article_names = []
+  valid_article_names = []
+  for item in item_list:
+    item_in_invalid_article_names = False
```
audiodude, line 9: maybe just `is_invalid`?

audiodude (changes requested):
This is really good, good job!

One last thing that I forgot, we should `.strip()` the article name before replacing spaces with underscores, so we don't end up with: `_Article_title___`. Add that to your test as well.

wp1/logic/selection_test.py
```diff
@@ -0,0 +1,32 @@
+import unittest
+from wp1.logic import selection
+
+
+class ValidateArticleNameTest(unittest.TestCase):
+
+  items = '''Eiffel_Tower     \nStatue of Liberty
```
mahakporwal02, line 7: I have added white spaces for testing and I have already added .strip() in the code.

wp1/logic/selection.py
```diff
@@ -6,21 +6,21 @@ def validate_list(items):
   invalid_article_names = []
   valid_article_names = []
   for item in item_list:
-    item_in_invalid_article_names = False
+    is_valid = False
```
audiodude, line 9: I think you inverted this. This will be true if the item "is invalid" right? Not if it "is valid"?

wp1/logic/selection_test.py
```diff
@@ -0,0 +1,32 @@
...
+https://en.wikipedia.org/w/index.php?title=Libertas
+https://en.wikipedia.org/w/index.php?title=%E2%88%9E
+George-%C3%89tienne_Cartier_Monument'''
+
+  def test_validate_items(self):
+    expected = ([
+        'Eiffel_Tower', 'Statue_of_Liberty', 'Liberty_(personification)',
+        'Libertas', '∞', 'George-Étienne_Cartier_Monument'
+    ], [
+        'This_is_supposedly_an_article_name_but_its_way_too_long_to_actually_be_one_because_it_contains_more_than_256_characters_which_is_not_allowed_in_article_titles_and_it_just_runs_on_forever,_it_doesnt_even_have_underscores_and_looks_more_like_a_long_paragraph_of_text_than_an_actual_article_name.',
```
audiodude, line 27: Can you split this up across multiple strings please? Not a triple quoted string but:

```
  ['This is the first part of the string, this is the '
  'second part of the string this is the third part '
  'of the string....']
```

audiodude (changes requested):
Just a few more changes, thanks!

audiodude (approved):
LGTM

PR #381, opened 2021-06-30
Create backend endpoint for saving simple lists

### LINKED ISSUE #379: Create backend endpoint for saving simple lists
For now, it should simply do the validation added in #378 and return the result to the frontend as an object:

```
{
  "success": true/false
  "items": {
    "valid": [],
    "invalid": []
  }
}
```

If success is true, the "items" object can be empty, indicating that the items were successfully saved.

We need to change the frontend page to:

1. Call this backend as a `fetch`
2. If the success field is false, display the valid elements in the main items `<textarea>` and the invalid list elements in a separate `<textarea>` with some error messaging ("The following items are not valid for selection lists") and allow the user to try again.
3. If the success field is true, redirect the user to the MyLists page (this will eventually be a redirect to the specific entry for this list, where the user can download the list as TSV or XLS).

Part of #336.

mahakporwal02:
This fixes #379

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -342,7 +342,7 @@ paths:
                     type: array
                     items:
                       type: string
-  /selection/:
+  /selection/simple:
     post:
       summary: 'Get response from sending acticle names'
       operationId: getResponse
@@ -361,29 +361,12 @@ paths:
                   type: string
       responses:
         200:
-          description: 'Successful operation'
+          description: 'If article names does not contain {"#", "<", ">", "[", "]", "{", "}", "|"}, success is true, list would be saved and the response would return an empty "items" object'
           content:
             application/json:
               schema:
                 type: object
                 properties:
-                  status:
-                    type: string
-                    default: '200'
-                  success:
-                    type: boolean
-                  items:
-                    type: object
-        403:
-          description: 'Unuccessful operation'
-          content:
-            application/json:
-              schema:
-                type: object
-                properties:
-                  status:
-                    type: string
-                    default: '403'
                   success:
                     type: boolean
                     default: false
--- a/openapi.yml
+++ b/openapi.yml
@@ -342,6 +342,66 @@ paths:
                     type: array
                     items:
                       type: string
+  /selection/:
+    post:
+      summary: 'Get response from sending acticle names'
+      operationId: getResponse
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                article_name:
+                  type: string
+                list_name:
+                  type: string
+                project_name:
+                  type: string
+      responses:
+        200:
+          description: 'Successful operation'
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  status:
+                    type: string
+                    default: '200'
+                  success:
+                    type: boolean
+                  items:
+                    type: object
+        403:
+          description: 'Unuccessful operation'
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  status:
+                    type: string
+                    default: '403'
+                  success:
+                    type: boolean
+                    default: false
+                  items:
+                    type: object
+                    properties:
+                      valid:
+                        type: array
+                        items:
+                          type: string
+                      invalid:
+                        type: array
+                        items:
+                          type: string
+                      forbidden:
+                        type: array
+                        items:
+                          type: string
 components:
   schemas:
     Project:
--- a/openapi.yml
+++ b/openapi.yml
@@ -353,15 +353,15 @@ paths:
             schema:
               type: object
               properties:
-                article_name:
+                articles:
                   type: string
-                list_name:
+                project:
                   type: string
                 project_name:
                   type: string
       responses:
         200:
-          description: 'If article names does not contain {"#", "<", ">", "[", "]", "{", "}", "|"}, success is true, list would be saved and the response would return an empty "items" object'
+          description: 'Successful operation'
           content:
             application/json:
               schema:
@@ -369,20 +369,24 @@ paths:
                 properties:
                   success:
                     type: boolean
-                    default: false
+                    description: True when article name doesnot contain forbidden chars {"#", "<", ">", "[", "]", "{", "}", "|"}, false otherwise.
                   items:
                     type: object
+                    description: Empty in case of success being true
                     properties:
                       valid:
                         type: array
+                        description: List of valid article names
                         items:
                           type: string
                       invalid:
                         type: array
+                        description: List of invalid article names
                         items:
                           type: string
                       forbidden:
                         type: array
+                        description: List of forbidden characters present in article names
                         items:
                           type: string
 components:
--- a/openapi.yml
+++ b/openapi.yml
@@ -1,8 +1,8 @@
 openapi: 3.0.0
 servers:
-  - url: http://localhost:5000/v1
+  - url: https://api.wp1.openzim.org/v1
     description: Production server HTTPS
-  - url: http://localhost:5000/v1
+  - url: http://api.wp1.openzim.org/v1
     description: Production server HTTP
 info:
   description: 'Wikipedia 1.0 is an editing project on Wikipedia. It includes a bot, called the WP 1.0 bot, and a frontend web interface for that bot which lives at https://wp1.openzim.org. This document describes the server that acts as the API backend for that frontend server. Generally speaking, it contains only features which are useful to that frontend. However it is documented here and made available without restriction in the hopes that it may be useful.'
--- a/openapi.yml
+++ b/openapi.yml
@@ -1,8 +1,8 @@
 openapi: 3.0.0
 servers:
-  - url: https://api.wp1.openzim.org/v1
+  - url: http://localhost:5000/v1
     description: Production server HTTPS
-  - url: http://api.wp1.openzim.org/v1
+  - url: http://localhost:5000/v1
     description: Production server HTTP
 info:
   description: 'Wikipedia 1.0 is an editing project on Wikipedia. It includes a bot, called the WP 1.0 bot, and a frontend web interface for that bot which lives at https://wp1.openzim.org. This document describes the server that acts as the API backend for that frontend server. Generally speaking, it contains only features which are useful to that frontend. However it is documented here and made available without restriction in the hopes that it may be useful.'
@@ -344,8 +344,8 @@ paths:
                       type: string
   /selection/simple:
     post:
-      summary: 'Get response from sending acticle names'
-      operationId: getResponse
+      summary: 'Save a new Simple list, validating it first and returning invalid items if they exist'
+      operationId: createSimpleList
       requestBody:
         required: true
         content:
@@ -369,7 +369,7 @@ paths:
                 properties:
                   success:
                     type: boolean
-                    description: True when article name doesnot contain forbidden chars {"#", "<", ">", "[", "]", "{", "}", "|"}, false otherwise.
+                    description: True when article name does not contain forbidden chars {"#", "<", ">", "[", "]", "{", "}", "|"}, false otherwise.
                   items:
                     type: object
                     description: Empty in case of success being true
@@ -389,6 +389,9 @@ paths:
                         description: List of forbidden characters present in article names
                         items:
                           type: string
+        400:
+          description: 'Request with either empty list_name or project or articles'
+
 components:
   schemas:
     Project:
--- a/openapi.yml
+++ b/openapi.yml
@@ -355,9 +355,9 @@ paths:
               properties:
                 articles:
                   type: string
-                project:
+                list_name:
                   type: string
-                project_name:
+                project:
                   type: string
       responses:
         200:
--- a/wp1-frontend/cypress/integration/createList_page.spec.js
+++ b/wp1-frontend/cypress/integration/createList_page.spec.js
@@ -36,4 +36,44 @@ describe('the createList page', () => {
     cy.get('#items > .form-control').click();
     cy.get('#items').contains('Please provide valid items.');
   });
+
+  it.only('unsucessful response', () => {
+    cy.visit('/#/selection/lists/simple/new');
+    cy.get('.needs-validation').then(form$ => {
+      form$.on('submit', e => {
+        e.preventDefault();
+      });
+    });
+    cy.get('#listName > .form-control')
+      .click()
+      .type('List Name');
+    cy.get('#items > .form-control')
+      .click()
+      .type('Eiffel_Tower\nStatue of Liberty#');
+    cy.intercept('v1/selection/').as('selection');
+    cy.get('#saveListButton').click();
+    cy.wait('@selection');
+    cy.get('#items > .form-control').should('have.value', 'Eiffel_Tower');
+    cy.get('#invalid_articles').contains(
+      'Following items are not valid for selection lists beacuse they have #'
+    );
+    cy.get('#invalid_articles > .form-control').should(
+      'have.value',
+      'Statue_of_Liberty#'
+    );
+  });
+
+  it('successful response', () => {
+    cy.visit('/#/selection/lists/simple/new');
+    cy.get('#listName > .form-control')
+      .click()
+      .type('List Name');
+    cy.get('#items > .form-control')
+      .click()
+      .type('Eiffel_Tower\nStatue of Liberty');
+    cy.intercept('v1/selection/').as('selection');
+    cy.get('#saveListButton').click();
+    cy.wait('@selection');
+    cy.url().should('eq', 'http://localhost:3000/?#/selection/user');
+  });
 });
--- a/wp1-frontend/cypress/integration/createList_page.spec.js
+++ b/wp1-frontend/cypress/integration/createList_page.spec.js
@@ -1,6 +1,6 @@
 /// <reference types="Cypress" />
 
-describe('the createList page', () => {
+describe('the CreateSimpleList page', () => {
   beforeEach(() => {
     cy.intercept('v1/sites/', { fixture: 'sites.json' });
   });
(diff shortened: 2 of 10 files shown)


mahakporwal02:
@audiodude Sorry for the late PR.

audiodude:
> @audiodude Sorry for the late PR.

I don't consider this late by any means, thanks!

audiodude (changes requested):
I think perhaps I didn't specify this clearly enough. The backend endpoint should eventually be the one we are going to use to save simple lists. As such, the URL should be `/selection/simple/`. The endpoint should not be called on validation, but on save, for this reason.

Also, since it is being called on save, it should return 200 for all valid requests. The 403 status generally means "NOT AUTHORIZED" and should be reserved for if the user attempts to save a list but is not logged in.

In the openapi.yml specification, you should describe in detail the meanings of the success boolean and what happens for values of true and false respectively. The response should not contain a redundant "status" field.

Also, the frontend should clear/hide the invalid items when submitting a second time. The idea is that I should be able to have a list of items, see which ones are invalid, and then click "Save" again and the list should save.

Sorry for any miscommunication!

audiodude:
Additionally, even though the frontend validates that the name and project are not empty, you should check these again on the API backend for "defense in depth" and the possibility that a request is being made from outside of our frontend.

openapi.yml
```diff
@@ -342,6 +342,49 @@ paths:
...
+        200:
+          description: 'If article names does not contain {"#", "<", ">", "[", "]", "{", "}", "|"}, success is true, list would be saved and the response would return an empty "items" object'
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  success:
+                    type: boolean
+                    default: false
```
audiodude, line 372: Remove. I don't think this is a valid use of the `default` indicator.

openapi.yml
```diff
@@ -342,6 +342,49 @@ paths:
...
+      summary: 'Get response from sending acticle names'
+      operationId: getResponse
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                article_name:
```
audiodude, line 356: Just call this `articles`.

openapi.yml
```diff
@@ -342,6 +342,49 @@ paths:
...
+                  type: string
+                list_name:
+                  type: string
+                project_name:
```
audiodude, line 360: Just call this `project`.

openapi.yml
```diff
@@ -342,6 +342,49 @@ paths:
...
+      responses:
+        200:
+          description: 'If article names does not contain {"#", "<", ">", "[", "]", "{", "}", "|"}, success is true, list would be saved and the response would return an empty "items" object'
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  success:
+                    type: boolean
```
audiodude, line 371: Can't you put a description or other annotation here to describe when it would be true or false? Same with other responses.

wp1-frontend/src/components/CreateSimpleList.vue
```diff
@@ -70,7 +86,11 @@ export default {
   name: 'MyLists',
```
audiodude, line 86: Why is the name still 'MyLists'? Didn't we fix that in the previous PR?

wp1-frontend/src/components/CreateSimpleList.vue
```diff
@@ -13,8 +13,8 @@
         </div>
         <form class="needs-validation" novalidate>
           <div class="m-4">
-            <label for="Project">Project</label>
-            <select class="custom-select my-list">
+            <label>Project</label>
+            <select ref="select" class="custom-select my-list">
```
audiodude, line 17: How about `ref="project"`?

wp1-frontend/src/components/CreateSimpleList.vue
```diff
@@ -43,14 +44,29 @@
               "
               class="form-control my-list"
               rows="13"
+              ref="article_name"
```
audiodude, line 47: How about `ref="articles"`?

wp1-frontend/src/components/CreateSimpleList.vue
```diff
@@ -87,21 +107,49 @@ export default {
...
-      var forms = document.querySelectorAll('.needs-validation');
-      Array.prototype.slice.call(forms).forEach(function(form) {
-        form.addEventListener(
-          'submit',
-          function(event) {
-            if (!form.checkValidity()) {
-              event.preventDefault();
-              event.stopPropagation();
-            }
+    saveForm: function() {
```
audiodude, line 110: nit: Just call this `save`, we know there's a form involved.

wp1-frontend/src/components/CreateSimpleList.vue
```diff
@@ -87,21 +107,49 @@ export default {
...
+      let parent = this;
+      const form = document.querySelectorAll('.needs-validation')[0];
+      form.addEventListener(
```
audiodude, line 113: You don't need to add an event listener manually. Look at the example on https://vuejs.org/v2/guide/events.html

`<form v-on:submit.prevent="onSubmit"></form>`

wp1/web/selection.py
```diff
@@ -0,0 +1,26 @@
+import flask
+from wp1.web import authenticate
+import wp1.logic.selection as selections
+
+selection = flask.Blueprint('selection', __name__)
+
+
+@selection.route('/simple', methods=['POST'])
+def get_article_names():
```
audiodude, line 9: `def create`

wp1/web/selection.py
```diff
@@ -0,0 +1,26 @@
...
+  list_name = flask.request.get_json()['list_name']
```
audiodude, line 10: Don't call .get_json() twice. Call it once and store the result in an intermediate variable.

wp1/web/selection_test.py
```diff
@@ -0,0 +1,43 @@
...
+      "items": {
+          'valid': ['Eiffel_Tower'],
+          'invalid': ['Statue_of#Liberty'],
+          'forbiden_chars': ['#']
+      }
+  }
+  valid_article_name = "Eiffel_Tower\nStatue of Liberty"
+  successful_response = {"status": "200", "success": True, "items": {}}
+
+  def test_get_unsuccessful_article_name(self):
```
audiodude, line 23: `def test_create_unsuccessful`

wp1/web/selection_test.py
```diff
@@ -0,0 +1,43 @@
...
+    with self.app.test_client() as client:
+      rv = client.post('/v1/selection/simple',
+                       json={
+                           'article_name': self.invalid_article_name,
+                           'list_name': 'my_list',
+                           'project_name': 'my_project'
+                       })
+      self.assertEqual(rv.get_json(), self.unsuccessful_response)
+
+  def test_get_successful_article_name(self):
```
audiodude, line 34: `def test_create_successful`

wp1/web/selection_test.py
```diff
@@ -0,0 +1,43 @@
+import unittest
+from wp1.web.app import create_app
+from wp1.web.base_web_testcase import BaseWebTestcase
+from wp1.logic import selection
+import json
+
+
+class ProjectTest(unittest.TestCase):
```
audiodude, line 8: You're missing tests for the 400 when the articles, list name, or project are missing. Should be three separate tests.

audiodude (changes requested):
It looks like you need to clean up the front end code a bit, but otherwise this is moving in the right direction.

wp1-frontend/src/components/CreateSimpleList.vue
mahakporwal02: We had changed the name from 'createSimpleLists' to 'MyLists'. Should I change it again?

openapi.yml
mahakporwal02: If I remove `default` the boolean value sets itself to `true`.But the `success` should be `false` if we are returning items

audiodude: You should remove it. I don't know what you mean by "sets itself to true".

openapi.yml
```diff
@@ -342,6 +342,53 @@ paths:
...
+        200:
+          description: 'Successful operation'
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  success:
+                    type: boolean
+                    description: True when article name doesnot contain forbidden chars {"#", "<", ">", "[", "]", "{", "}", "|"}, false otherwise.
```
audiodude, line 372: nit: s/doesnot/does not/

wp1-frontend/src/components/CreateSimpleList.vue
```diff
@@ -11,10 +11,10 @@
           storage and can be accessed through URLs that will be provided once it
           has been saved.
         </div>
-        <form class="needs-validation" novalidate>
+        <form v-on:submit.prevent class="needs-validation" novalidate>
```
audiodude, line 14: Shouldn't the value of this attribute be your save method? You want to call the save method on submit right?

wp1-frontend/src/components/CreateSimpleList.vue
```diff
@@ -87,21 +107,38 @@ export default {
...
-              event.stopPropagation();
-            }
-            form.classList.add('was-validated');
-          },
-          false
-        );
-      });
+    save: async function() {
+      let parent = this;
+      const form = document.querySelectorAll('.needs-validation')[0];
```
audiodude, line 112: Why not give the form a ref attribute so you can get it directly?

wp1/web/selection.py
```diff
@@ -0,0 +1,28 @@
...
+
+@selection.route('/simple', methods=['POST'])
+def create():
+  response = flask.request.get_json()
+  list_name = response['list_name']
+  items = response['article_name']
+  project = response['project_name']
+  if not items or not list_name or not project:
+    flask.abort(400)
+  valid_article_names, invalid_article_names, forbiden_chars = selections.validate_list(
```
audiodude, line 16: nit: line too long?

wp1/web/selection.py
```diff
@@ -0,0 +1,28 @@
...
+      items)
+  if invalid_article_names:
+    return {
+        "status": "200",
```
audiodude, line 20: You shouldn't return a "status" field, that's redundant with the HTTP status.

wp1/web/selection.py
```diff
@@ -0,0 +1,28 @@
...
+from wp1.web import authenticate
+import wp1.logic.selection as selections
+
+selection = flask.Blueprint('selection', __name__)
+
+
+@selection.route('/simple', methods=['POST'])
+def create():
+  response = flask.request.get_json()
+  list_name = response['list_name']
```
audiodude, line 11: You need to edit these parameter names to match what's in the openapi.yml. Then you'll need to update the tests as well.

wp1-frontend/cypress/integration/createList_page.spec.js
```diff
@@ -36,4 +36,49 @@ describe('the createList page', () => {
     cy.get('#items > .form-control').click();
     cy.get('#items').contains('Please provide valid items.');
   });
+
+  it('unsucessful response', () => {
+    cy.visit('/#/selection/lists/simple/new');
+    cy.get('.needs-validation').then(form$ => {
+      form$.on('submit', e => {
+        e.preventDefault();
```
audiodude, line 44: This shouldn't be necessary anymore.

wp1-frontend/cypress/integration/createList_page.spec.js
```diff
@@ -36,4 +36,49 @@ describe('the createList page', () => {
...
+      .type('List Name');
+    cy.get('#items > .form-control')
+      .click()
+      .type('Eiffel_Tower\nStatue of Liberty#');
+    cy.intercept('v1/selection/').as('selection');
+    cy.get('#saveListButton').click();
+    cy.wait('@selection');
+    cy.get('#items > .form-control').should('have.value', 'Eiffel_Tower');
+    cy.get('#invalid_articles').contains(
+      'Following items are not valid for selection lists beacuse they have #'
```
audiodude, line 58: s/beacuse/because/

You'll have to fix it in the actual form as well, too.

wp1-frontend/cypress/integration/createList_page.spec.js
```diff
@@ -36,4 +36,49 @@ describe('the createList page', () => {
     cy.get('#items > .form-control').click();
     cy.get('#items').contains('Please provide valid items.');
   });
+
+  it('unsucessful response', () => {
```
audiodude, line 40: The it clause should read like a sentence, and capture the behavior you're trying to test. So "It unsuccessful response" is not a logical sentence. Something like "It flams the whoobong when the blah blah blah" -> `it('flams the whoobong when the blah blah blah')`.

It looks like I didn't catch this for the initial tests on this page, but you should go back and edit those it blocks as well. It would be fine to do that in this same PR.

wp1-frontend/cypress/integration/createList_page.spec.js
```diff
@@ -36,4 +36,49 @@ describe('the createList page', () => {
...
+    cy.get('#invalid_articles').contains(
+      'Following items are not valid for selection lists beacuse they have #'
+    );
+    cy.get('#invalid_articles > .form-control').should(
+      'have.value',
+      'Statue_of_Liberty#'
+    );
+  });
+
+  it('successful response', () => {
```
audiodude, line 66: Same as above with regard to it calls reading like a sentence.

wp1-frontend/cypress/integration/createList_page.spec.js
```diff
@@ -36,4 +36,49 @@ describe('the createList page', () => {
...
+      form$.on('submit', e => {
+        e.preventDefault();
+      });
+    });
+    cy.get('#listName > .form-control')
+      .click()
+      .type('List Name');
+    cy.get('#items > .form-control')
+      .click()
+      .type('Eiffel_Tower\nStatue of Liberty#');
```
audiodude, line 52: You should make it more obvious which item is invalid. Even something like `.../nInvalid#Item` would be better

wp1-frontend/cypress/integration/createList_page.spec.js
```diff
@@ -36,4 +36,49 @@ describe('the createList page', () => {
...
+    cy.intercept('v1/selection/').as('selection');
+    cy.get('#saveListButton').click();
+    cy.wait('@selection');
+    cy.get('#items > .form-control').should('have.value', 'Eiffel_Tower');
+    cy.get('#invalid_articles').contains(
+      'Following items are not valid for selection lists beacuse they have #'
+    );
+    cy.get('#invalid_articles > .form-control').should(
+      'have.value',
+      'Statue_of_Liberty#'
```
audiodude, line 62: Do you think it's a bug that the user typed 'Statue of Liberty#' and we told them that 'Statue_of_Liberty#' (with underscores) is invalid? Maybe we can do better. Just something to think about maybe for a follow up PR.

wp1-frontend/src/components/CreateSimpleList.vue
```diff
@@ -87,21 +107,38 @@ export default {
         this.wikiProjects[index] = element.replace('https://', '');
```
audiodude, line 107: Do you think we should maybe do this in the selection list retrieval code (where we get the URLs from the meta wikipedia API)? Do we have any use for full URLs with https?

openapi.yml
```diff
@@ -342,6 +342,53 @@ paths:
                     type: array
                     items:
                       type: string
+  /selection/simple:
+    post:
+      summary: 'Get response from sending acticle names'
```
audiodude, line 347: "Get response" is not the best documentation. All HTTP endpoints return a response. Think about what the user is actually doing when they call this endpoint. Something like 'Save a new Simple list, validating it first and returning invalid items if they exist'. Okay I guess I wrote it for you ;)

openapi.yml
```diff
@@ -342,6 +342,53 @@ paths:
...
+      operationId: getResponse
```
audiodude, line 348: getResponse is way too generic. How about 'createSimpleList'?

openapi.yml
```diff
@@ -342,6 +342,53 @@ paths:
...
+              type: object
+              properties:
+                articles:
+                  type: string
+                list_name:
+                  type: string
+                project:
+                  type: string
+      responses:
+        200:
```
audiodude, line 363: You're returning 400 if the params are invalid right? You need to document that here.

audiodude (changes requested):
Thanks for sticking with this one! Although it is mostly correct as is, I think we can make some changes to really add polish. Thanks!

wp1-frontend/src/components/CreateSimpleList.vue
audiodude: I believe that change was made erroneously. This is the page for creating simple lists right? The name of the file is `CreateSimpleList.vue`. The other page is `MyLists.vue` right?

wp1/web/selection.py
```diff
@@ -0,0 +1,28 @@
+import flask
+from wp1.web import authenticate
+import wp1.logic.selection as selections
+
+selection = flask.Blueprint('selection', __name__)
+
+
+@selection.route('/simple', methods=['POST'])
+def create():
+  response = flask.request.get_json()
```
audiodude, line 10: This is not a response, it is a request, so calling it response is very confusing. Either way `data` or simply `json` would be a better name.

openapi.yml
```diff
@@ -1,8 +1,8 @@
 openapi: 3.0.0
 servers:
-  - url: https://api.wp1.openzim.org/v1
+  - url: http://localhost:5000/v1
```
audiodude, line 3: Please change these back.

wp1-frontend/cypress/integration/createList_page.spec.js
```diff
@@ -9,31 +9,66 @@ describe('the createList page', () => {
     cy.visit('/#/selection/lists/simple/new');
   });
 
-  it('displays wiki projects', () => {
+  it('wiki projects displayed', () => {
```
audiodude, line 12: This one was actually fine "It displays wiki projects" is a good sounding sentence.

wp1-frontend/cypress/integration/createList_page.spec.js
```diff
@@ -9,31 +9,66 @@ describe('the createList page', () => {
...
   });
 
-  it('textbox item validation', () => {
+  it('validates textbox on clicking save', () => {
     cy.get('#saveListButton').click();
     cy.get('#items').contains('Please provide valid items.');
   });
 
-  it('list name validation on blur', () => {
+  it('validates list name on loosing focus', () => {
```
audiodude, line 28: s/loosing/losing/

wp1-frontend/cypress/integration/createList_page.spec.js
```diff
@@ -9,31 +9,66 @@ describe('the createList page', () => {
...
     cy.visit('/#/selection/lists/simple/new');
     cy.get('#listName > .form-control').click();
     cy.get('#listName').contains('Please provide a valid List Name.');
   });
 
-  it('textbox item validation on blur', () => {
+  it('validates textbox on loosing focus', () => {
```
audiodude, line 34: s/loosing/losing/

wp1-frontend/cypress/integration/createList_page.spec.js
```diff
@@ -9,31 +9,66 @@ describe('the createList page', () => {
...
     cy.visit('/#/selection/lists/simple/new');
     cy.get('#items > .form-control').click();
     cy.get('#items').contains('Please provide valid items.');
   });
+
+  it('textbox displayed with invalid article names', () => {
```
audiodude, line 40: It 'displays a textbox with invalid article names'

wp1/web/sites.py
```diff
@@ -23,7 +23,9 @@ def get_sites():
   sites_data = []
   for i in sitematrix:
     for j in sitematrix[i]['site']:
-      sites_data.append(j['url'])
+      project = j['url'].replace('https://', '')
+      sites_data.append(project)
 
   redis.setex('sites_data', timedelta(days=1), value=','.join(sites_data))
+  print(sites_data)
```
audiodude, line 30: nit: remove print.

audiodude (changes requested):
Very close to being done. I actually patched this PR and it worked rather well. Thanks again!

wp1-frontend/cypress/integration/createList_page.spec.js
mahakporwal02: Maybe when we send response `(success=false)` to frontend we should remove the `underscores`?

audiodude: The thing is, at that point we won't know if the user originally typed the underscores or if we added them, so if we universally remove them, we won't be doing the right thing in all cases.

It seems like if we wanted to display the original item, we'd have to change the `validate_list` method to check for forbidden characters before it transforms the spaces. Do you want to file an issue for that?

audiodude (approved):
LGTM

PR #382, opened 2021-07-05
Database migration and model for Selections

### LINKED ISSUE #374: Schema for selection metadata in tool database
I'm thinking of the following schema for the `selections` table in MariaDB:

| s_id | s_name | s_user_id | s_hash | s_project | s_model | s_region | s_bucket | s_object_key | s_last_generated |
| - | - | - | - | - | - | - | - | - | - |
| INTEGER PRIMARY KEY AUTOINCREMENT | VARBINARY(255) | INTEGER [user.u_id] | VARBINARY(255) | VARBINARY(255) [eg: en.wikipedia.org] | INTEGER(8) [enum] | VARBINARY(255) | VARBINARY(255) | VARBINARY(255) | BINARY(20)

- **s_name**: The name that the user gave for the list.
- **s_user_id**: The id of the user that requested the generation of this selection.
- **s_hash**: I realize that I need a filename for the S3 bucket/redis key for each selection list. I'm thinking of using a salted hash of the database id. I don't want to use the database id directly because they are increasing monotonically from 1, so abusive users could just go to `selections/1.tsv`, `selections/2.tsv` etc and list all of the selections which we don't really want to encourage.
- **s_project**: The project that the list belongs to.
- **s_model**: id for a "model", such as a list builder, that was used to generate this selection. I'm imagining this as kind of a URL like a java package so something like "org.openzim.wp1.simplelist:1" or "com.travisbriggs.funkylist:3" where the data after the colon is the version of the model that was used.
- **s_backend**: I'm also thinking about how we get this working in development, and my best guess is that we have multiple 'backends' that can serve the list data, with a defined interface for "get" and "put" of lists. In development, it can be "redis" (because not all developers will have access to S3), while in production it will be "s3".
- **s_region**: The s3 region that the data is located in. NULL for backend == Redis.
- **s_bucket**: The s3 bucket name. NULL for backend == Redis.
- **s_object_key**: The "path" or object key in s3. If backend == Redis, this is the key in Redis.
- **s_last_generated**: The timestamp of when the data was last generated. Note that this is in the same timestamp format that the rest of the database uses, a binary string of length 20.

Also note the use of VARBINARY instead of VARCHAR, because everything in the enwp10 database uses VARBINARY to avoid using database collation.

This comment will be updated as I think of modifications to the proposed schema.

audiodude:
This PR adds a database migration for adding the selections table to the WP10 MariaDB database. Developers will have to migrate their development databases, and the production database migrations will also have to be run.

There is also a model added for Selection.

Fixes #374

### DIFF
--- a/db/migrations/20210705_01_TgcHL-add-selections-table.py
+++ b/db/migrations/20210705_01_TgcHL-add-selections-table.py
@@ -9,7 +9,7 @@ __depends__ = {'20210528_01_xt830-add-users-table'}
 steps = [
     step(
         "CREATE TABLE selections ("
-        "s_id INTEGER NOT NULL PRIMARY KEY, "
+        "s_id INTEGER NOT NULL PRIMARY KEY AUTO_INCREMENT, "
         "s_name VARBINARY(255) NOT NULL, "
         "s_user_id INTEGER NOT NULL, "
         "s_hash VARBINARY(255), "
--- a/db/migrations/20210705_01_TgcHL-add-selections-table.py
+++ b/db/migrations/20210705_01_TgcHL-add-selections-table.py
@@ -1,0 +1,23 @@
+"""
+Add selections table
+"""
+
+from yoyo import step
+
+__depends__ = {'20210528_01_xt830-add-users-table'}
+
+steps = [
+    step(
+        "CREATE TABLE selections ("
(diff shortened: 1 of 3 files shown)


mahakporwal02 (approved):
LGTM

PR #385, opened 2021-07-06
Fix column name in created_at migration

audiodude:
Also add a logger to the Selection model, because it requires it.


mahakporwal02 (commented):
LGTM

PR #387, opened 2021-07-07
added auth to selections

### LINKED ISSUE #386: Authorization required for CreateSimpleList and MyLists
We implemented OAuth specifically so we could protect these endpoints and make login required for them. Now we should actually do it.

As part of this, we should refactor the HTML that is currently displayed on the Manual update page when you're not logged in. We should create a LoginRequired.vue component that we can show on all of these pages, so that if we want to later modify it, we only have to do it in one place.

mahakporwal02:
This fixes #386


wp1-frontend/cypress/fixtures/failure.json
```diff
@@ -0,0 +1,8 @@
+{
```
audiodude, line 1: I would name this file `save_list_failure` because there might be other failure fixtures in the future.

wp1-frontend/cypress/fixtures/success.json
```diff
@@ -0,0 +1 @@
+{ "success": true, "items": {} }
```
audiodude, line 1: Same thing: `save_list_success`.

wp1-frontend/cypress/integration/createList_page.spec.js
```diff
@@ -59,16 +65,22 @@ describe('the CreateSimpleList page', () => {
...
       .click()
       .type('Eiffel_Tower\nStatue of Liberty');
-    cy.intercept('v1/selection/').as('selection');
+    cy.intercept('v1/selection/simple', { fixture: 'success.json' });
     cy.get('#saveListButton').click();
-    cy.wait('@selection');
     cy.url().should('eq', 'http://localhost:3000/#/selection/user');
   });
+
+  it('opens login page', () => {
```
audiodude, line 81: I would put all of the above tests in a `describe('when the user is logged in', ...` block, so that you could put the cy.intercept for identify in a beforeEach and not have to repeat it every time. Then put this one in a `describe('when the user is not logged in', ...)` block.

wp1-frontend/src/components/loginRequired.vue
```diff
@@ -0,0 +1,23 @@
...
+    </h2>
+    <a class="row justify-content-center pt-2" :href="this.loginInitiateUrl"
+      ><button type="button" class="btn btn-primary">Login</button></a
+    >
+  </div>
+</template>
+
+<script>
+export default {
+  name: 'loginRequired',
```
audiodude, line 14: Components should be upper camel case: `LoginRequired`. Be sure to rename the file too.

wp1-frontend/src/components/loginRequired.vue
```diff
@@ -0,0 +1,23 @@
+<template>
+  <div>
+    <h2 class="row justify-content-center pt-5">
```
audiodude, line 3: Shouldn't this be used on the ManualUpdate page too?

audiodude (changes requested):
Good work, just a few minor things to change.

wp1-frontend/src/components/loginRequired.vue
mahakporwal02: Are you talking about the `LoginRequired` component?

audiodude: Yes, just wondering because I don't see any changes to the ManualUpdate component and I assume it should be using the same LoginRequired component.

mahakporwal02: But I have made changes to `Update Page` component. 
And I am using `LoginRequired` component only.

audiodude: Good point. I only counted two usages and forgot the page was called UpdatePage. I guess the missing usage is on MyLists?

mahakporwal02: Should I add it?
I purposely didn't add it

audiodude: Can you expand on your reasoning for not adding it?

mahakporwal02: Sorry I got confused with the other PR.
I'll fix it.

audiodude (approved):
LGTM

Issue #370, opened 2021-06-26
3 differents concepts and storage backends


kelson42:
We have three different concepts around lists:
* List builders, which are methods to build list, stored in the code base. They work as templates aiming for values coming from users.
* User lists, which feeds the list builders with custom user generated values. IMO best stored in the MariaDB
* The selections, which are the output of user lists done at a specific time. They are directly usable by third party systems like Zimfarm. Should be in S3.

Does the current code base fit to that description? Maybe modulo the naming? I ask because we I woukd start the discussion about backup strategy, lifetime on S3 data, etc…

audiodude:
The current codebase fits this description in the sense that we haven't built anything for 1 and 2 yet, and have only just started on the data model for 3.

We also need to store metadata for selections in MariaDB, even though the textual article data will be stored in S3, so that we can build S3 URLS and provide a list of user selections that can be updated or removed.

I'm thinking of the following schema for the `selections` table in MariaDB:

| s_id | s_user_id | s_model | s_backend | s_region | s_bucket | s_object_key | s_last_generated |
| --- | --- | --- | --- | --- | --- | ---| --- |
| INTEGER PRIMARY KEY AUTOINCREMENT | INTEGER [user.u_id] | VARBINARY(255) | INTEGER(8) [enum] | VARBINARY(255) | VARBINARY(255) | VARBINARY(255) | BINARY(20)

- **s_user_id**: The id of the user that requested the generation of this selection.
- **s_model**: id for a "model", such as a list builder, that was used to generate this selection. I'm imagining this as kind of a URL like a java package so something like "org.openzim.wp1.simplelist:1" or "com.travisbriggs.funkylist:3" where the data after the colon is the version of the model that was used.
- **s_backend**: I'm also thinking about how we get this working in development, and my best guess is that we have multiple 'backends' that can serve the list data, with a defined interface for "get" and "put" of lists. In development, it can be "redis" (because not all developers will have access to S3), while in production it will be "s3".
- **s_region**: The s3 region that the data is located in. NULL for backend == Redis.
- **s_bucket**: The s3 bucket name. NULL for backend == Redis.
- **s_object_key**: The "path" or object key in s3. If backend == Redis, this is the key in Redis.
- **s_last_generated**: The timestamp of when the data was last generated. Note that this is in the same timestamp format that the rest of the database uses, a binary string of length 20.

[...]

PR #389, opened 2021-07-10
Persist simple lists

audiodude:
This rather large PR includes the logic that is called from the backend simple list "create" endpoint. The new logic method `persist_simple_list` handles storing a simple list selection in both the database (for its metadata) and the backend s3 storage (for its article list). Includes tests for the endpoint as well as the logic methods and the model methods.


audiodude:
@mahakporwal02 sorry for the rather large PR!

wp1/logic/selection.py
```diff
@@ -29,3 +35,68 @@ def validate_list(items):
...
+    cursor.execute(
+        '''INSERT INTO
+           selections (s_name, s_user_id, s_project, s_model,
+                       s_last_generated, s_created_at)
+           VALUES
+           (%(s_name)s, %(s_user_id)s, %(s_project)s, %(s_model)s,
+            %(s_last_generated)s, %(s_created_at)s)
+    ''', attr.asdict(selection))
+    return cursor.lastrowid
+  wp10db.commit()
```
mahakporwal02, line 51: Shouldn't this line should be above return statement?

wp1/web/app.py
```diff
@@ -125,6 +125,7 @@ def swagger_api_docs_yml():
 
   app.register_blueprint(projects, url_prefix='/v1/projects')
   app.register_blueprint(articles, url_prefix='/v1/articles')
+  app.register_blueprint(selection, url_prefix='/v1/selection')
```
mahakporwal02, line 128: Blueprint has been registered twice.

wp1/web/selection.py
```diff
@@ -14,15 +19,26 @@ def create():
...
         }
-    }
-  return {"success": True, "items": {}}
+    })
+
+  # All valid, time to upload to s3 and store in the application database.
+  s3 = get_storage()
+  wp10db = get_db('wp10db')
+  user_id = flask.session['user']['identity']['sub']
+  logic_selection.persist_simple_list(wp10db, s3, list_name, user_id, project,
```
mahakporwal02, line 41: Shouldn't this should be under try catch?

mahakporwal02 (changes requested):
Some changes I could figure out. Will you also add developers overlay afterwards?

audiodude:
Thanks for the review! I'm actually going to abandon this PR based on the comments on #370 and #374. We need a different schema for list builders/selections and a more generalized way of "materializing" built lists to the storage backend as selections.

PR #391, opened 2021-07-24
New schema for Selections that respects the difference between Selections and List Builders

audiodude:
This PR includes a migration that removes the old selections table, since it conflates the ideas of list builders and selections as discussed in #370. The PR further creates both a table for list builders (named just "builders") which the user can input and save, and selections, which are materialized lists of articles that are generated from builders but stored in the s3-like backend.

There is a one-to-many relationship between builders and selections, in that many selections (presumably of different content types) can be generated from a builder.

This PR also provides basic `attrs` based classes for modeling the data in these tables.

### DIFF
--- a/db/migrations/20210724_01_p2yUM-remove-deprecated-selections-table.py
+++ b/db/migrations/20210724_01_p2yUM-remove-deprecated-selections-table.py
@@ -1,0 +1,9 @@
+"""
+Remove deprecated selections table
+"""
+
+from yoyo import step
+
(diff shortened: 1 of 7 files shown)


kelson42 (approved):
@audiodude Thank you very much for this new PR taking in account my previous comments. I don't filled properly skilled to review/comment the code itself. But, the approach, like discribed in the PR, seems conform to my previous comments and therefore conform to my expectations.

mahakporwal02 (approved):

PR #396, opened 2021-07-29
AbstractBuilder with tests

### LINKED ISSUE #394: Abstract interface for Builder Modules
According to the first item in #370, we need to have code in the repo for Builder Modules, which are units of code (modules) that provide a method for validating user input (params) and calculating the selection from those params. Per #370, there is one Builder Module to many instances of Builders, which are user input params and metadata that are stored in the database.

It seems that we could benefit from an abstract class for this, `AbstractBuilder`, which at the very least could provide `materialize` and `build` methods, where `materialize` calls build with the parameters loaded from the Builder object from the database. In the abstract class, `materialize` would be in charge of creation Selection entries in the database and storing the generated selections in the backend bucket storage. The `build` method would contain the specific logic of the Builder Module to transform a dictionary of params into a selection list. In the abstract class,`build` would simply raise a `MethodNotImplementedError`, while `materialize` would be fully implemented.

In this way, new Builder Modules could simply implement their `build` methods and inherit from `AbstractBuilder` in order to deal with the database and backend storage (they wouldn't have to do it).

The AbstractBuilder could also provide a `validate` method that would return a tuple of (valid_params, invalid_params, error_messages), which would abstract our current logic.selection.validate_list method in the case of the Simple List module. The abstract version of this would also raise `MethodNotImplementedError`.

audiodude:
Includes database migration to add b_model to builders table and necessary test schema files.

This code doesn't do much on it's own, but provides an interface for a subclass to implement the not-implemented build method. Given such an implementation, and when passed a valid builder and a content type, the AbstractBuilder materialize method will build the selection from the builder and store the resultant selection list in s3. It will also save a Selection record in the WP 1.0 database for the selection list.

Fixes #394

### DIFF
--- a/wp1/selection/abstract_builder.py
+++ b/wp1/selection/abstract_builder.py
@@ -1,0 +1,41 @@
+import io
+import json
+
+from wp1.constants import CONTENT_TYPE_TO_EXT
+import wp1.logic.selection as logic_selection
+from wp1.models.wp10.selection import Selection
+
+
+class AbstractBuilder:
+
+  def _upload_to_storage(self, s3, selection, builder):
+    ext = CONTENT_TYPE_TO_EXT.get(selection.s_content_type, '').encode('utf-8')
+    object_key = b'selection/%(model)s/%(user_id)s/%(id)s.%(ext)s' % {
+        b'model': builder.b_model,
+        b'user_id': bytes(builder.b_user_id),
+        b'id': selection.s_id,
+        b'ext': ext,
+    }
+
+    upload_data = io.BytesIO()
+    upload_data.write(selection.data)
+    upload_data.seek(0)
+    s3.upload_fileobj(upload_data, key=object_key.decode('utf-8'))
+
+  def materialize(self, s3, wp10db, builder, content_type):
+    params = json.loads(builder.b_params)
+
+    selection = Selection(s_content_type=content_type,
+                          s_builder_id=builder.b_id)
+    selection.set_id()
+    selection.data = self.build(content_type, **params)
+    selection.set_updated_at_now()
+    self._upload_to_storage(s3, selection, builder)
+
+    logic_selection.insert_selection(wp10db, selection)
+
+  def build(self, content_type, **params):
+    raise NotImplementedError()
+
+  def validate(self, **params):
+    raise NotImplementedError()
--- a/wp1/selection/abstract_builder_test.py
+++ b/wp1/selection/abstract_builder_test.py
@@ -41,7 +41,7 @@ class AbstractBuilderTest(BaseWpOneDbTest):
     self.assertEqual(actual.s_builder_id, 100)
 
   @patch('wp1.models.wp10.selection.uuid.uuid4', return_value='abcd-1234')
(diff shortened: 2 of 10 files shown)


wp1/selection/abstract_builder_test.py
```diff
@@ -0,0 +1,62 @@
...
+  @patch('wp1.models.wp10.selection.uuid.uuid4', return_value='abcd-1234')
+  def test_materialize_selection_id(self, mock_setid):
+    self.test_builder.materialize(self.s3, self.wp10db, self.builder,
+                                  'text/tab-separated-values')
+    actual = _get_first_selection(self.wp10db)
+    self.assertEqual(actual.s_id, b'abcd-1234')
+
+  @patch('wp1.models.wp10.selection.utcnow',
+         return_value=datetime(2020, 12, 25, 10, 55, 44))
+  def test_materialize_selection_updated_at(self, mock_utcnow):
```
mahakporwal02, line 52: `mock_utcnow` is not being used. It has already been mocked in above line.

wp1/selection/abstract_builder_test.py
```diff
@@ -0,0 +1,62 @@
...
+
+  def test_materialize_creates_selection(self):
+    self.test_builder.materialize(self.s3, self.wp10db, self.builder,
+                                  'text/tab-separated-values')
+    actual = _get_first_selection(self.wp10db)
+    self.assertEqual(actual.s_content_type, b'text/tab-separated-values')
+    self.assertEqual(actual.s_builder_id, 100)
+
+  @patch('wp1.models.wp10.selection.uuid.uuid4', return_value='abcd-1234')
+  def test_materialize_selection_id(self, mock_setid):
```
mahakporwal02, line 44: `mock_setid` is not being used.

wp1/selection/abstract_builder.py
```diff
@@ -0,0 +1,41 @@
...
+    upload_data.seek(0)
+    s3.upload_fileobj(upload_data, key=object_key.decode('utf-8'))
+
+  def materialize(self, s3, wp10db, builder, content_type):
+    params = json.loads(builder.b_params)
+
+    selection = Selection(s_content_type=content_type,
+                          s_builder_id=builder.b_id)
+    selection.set_id()
+    selection.data = self.build(content_type, **params)
```
mahakporwal02, line 31: `selection.data` is not defined in selection model.

mahakporwal02 (changes requested):
Conceptually LGTM.

wp1/selection/abstract_builder_test.py
audiodude: I tried leaving out the parameter, but I get an error because the `@patch` decorator injects the patched value as a parameter even if you don't use it.

wp1/selection/abstract_builder_test.py
audiodude: Same thing here.

wp1/selection/abstract_builder.py
audiodude: Done

mahakporwal02 (approved):
LGTM

PR #398, opened 2021-08-02
added simple list builder module

### LINKED ISSUE #395: Implementation for Simple list builder module
Given the AbstractBuilder in #394, we need an implementation of the Simple list module.

It should take as its parameters:

```
{
  "list": [
    "Statue_of_Liberty",
    "Eiffel_Tower"
  ]
}
```

These parameters will be stored in the database when the Simple builder is saved from the frontend.

The `build` method will simply unpack those parameters and return concatenated bytes as a TSV, for content_type = 'text/tab-separated-values'.

The `validate` method will operate much like the current `validate_list` method in the codebase, however it should return data in the same format as the parameters. Also, the third tuple item should be actual error messages rather than just a list of forbidden characters.

So, the `validate` method could return a tuple:

```
(['Statue_of_Liberty', 'Eiffel_Tower'], ['Some_article#hash'], ['The list contained the following invalid characters: #'])
```

mahakporwal02:
This fixes #395.


wp1/selection/simple_builder.py
```diff
@@ -0,0 +1,14 @@
...
+from wp1.selection.abstract_builder import AbstractBuilder
+
+
+class SimpleBuilder(AbstractBuilder):
+
+  def build(self, content_type, **params):
+    return '\n'.join(params['list']).encode('utf-8')
+
+  def validate(self, **params):
+    valid_names, invalid_names, forbiden_chars = validate_list(params['list'])
```
audiodude, line 11: There is no need to have validate_list be it's own function in its own module. It's only used for validating Simple builders like this one. For now, you can move the function to this module as `_validate_list` and call it the way you are doing. But eventually, the validate method on SimpleBuilder should ***be*** that function, if that makes sense.

wp1/selection/simple_builder.py
```diff
@@ -0,0 +1,14 @@
+from wp1.logic.selection import validate_list
+from wp1.selection.abstract_builder import AbstractBuilder
+
+
+class SimpleBuilder(AbstractBuilder):
+
+  def build(self, content_type, **params):
+    return '\n'.join(params['list']).encode('utf-8')
```
audiodude, line 8: We should validate that the params contain a 'list' entry. We might also want to validate that the params don't contain any **other** entries, to protect against accidentally calling the SimpleBuilder with params that looks similar to, but aren't quite, SimpleBuilder params.

For now, you can simply raise `ValueError`, but we should create an issue for tracking a custom Exception hierarchy that we can catch for explicitly.

wp1/selection/simple_builder.py
audiodude: We should validate that 'list' exists here too, see comment above.

audiodude (changes requested):
As part of moving validate_list into the simple_builder.py file, you should move the tests as well. And also add tests for build(). Thanks!

wp1/selection/simple_builder_test.py
```diff
@@ -0,0 +1,71 @@
...
+    simple_test_builder = SimpleBuilder()
+    params = {'list': ['Eiffel_Tower', 'Statue#of_Liberty', 'Libertas']}
+    actual = simple_test_builder.build('text/tab-separated-values', **params)
+    self.assertEqual(b'Eiffel_Tower\nStatue#of_Liberty\nLibertas', actual)
+
+  def test_build_incorrect_params(self):
+    simple_test_builder = SimpleBuilder()
+    params = {'items': ['Eiffel_Tower', 'Statue#of_Liberty', 'Libertas']}
+    actual = simple_test_builder.build('text/tab-separated-values', **params)
+    self.assertRaises(KeyError)
```
audiodude, line 62: You need to use a context manager here:

```
with self.assertRaises(KeyError):
   code_that_raises()
```

Otherwise nothing is being asserted.

wp1/selection/simple_builder.py
```diff
@@ -0,0 +1,49 @@
...
+            "https://en.wikipedia.org/wiki/",
+            "").replace("https://en.wikipedia.org/w/index.php?title=", "")
+        valid_article_names.append(article_name)
+    return (valid_article_names, invalid_article_names, forbiden_chars)
+
+  def build(self, content_type, **params):
+    if len(params) == 1:
+      if 'list' in params.keys():
+        return '\n'.join(params['list']).encode('utf-8')
+      KeyError('Incorrect params present')
```
audiodude, line 41: You need to actually raise the exception:

```
raise ValueError('Missing required param: list')
```

Let's not use KeyError, that has a very specific meaning around dictionaries. It also is very nice to specify the actual key/param that caused the error.

wp1/selection/simple_builder.py
```diff
@@ -0,0 +1,49 @@
...
+    ValueError('Unwanted params present')
```
audiodude, line 42: s/Unwanted/Additional unnecessary/

wp1/selection/simple_builder.py
```diff
@@ -0,0 +1,49 @@
...
+
+  def validate(self, **params):
+    valid_names, invalid_names, forbiden_chars = self._validate_list(**params)
+    if forbiden_chars[0] != 'Empty List':
+      forbiden_chars.insert(
```
audiodude, line 47: Why `insert` and not just `append`?

wp1/selection/simple_builder.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class SimpleBuilder(AbstractBuilder):
+
+  def _validate_list(self, **params):
+    if not params['list']:
+      return ([], [], ['Empty List'])
+    invalid_article_names = []
+    valid_article_names = []
+    forbiden_chars = []
```
audiodude, line 12: Just realized that "forbidden" is spelled wrong throughout this function. Please fix.

wp1/selection/simple_builder.py
```diff
@@ -0,0 +1,49 @@
...
+
+  def build(self, content_type, **params):
+    if len(params) == 1:
+      if 'list' in params.keys():
+        return '\n'.join(params['list']).encode('utf-8')
+      KeyError('Incorrect params present')
+    ValueError('Unwanted params present')
+
+  def validate(self, **params):
+    valid_names, invalid_names, forbiden_chars = self._validate_list(**params)
```
audiodude, line 45: `forbiden_chars` is spelled wrong here too.

wp1/selection/simple_builder.py
```diff
@@ -0,0 +1,49 @@
...
+    if forbiden_chars[0] != 'Empty List':
+      forbiden_chars.insert(
+          0, 'The list contained the following invalid characters: ')
+    return (valid_names, invalid_names, forbiden_chars)
```
audiodude, line 49: The last value of the tuple should be full, English error messages like:

['The list contained the following invalid characters: # <', 'The list contained an item longer than 256 bytes', 'The list did not satisfy some other foo bar requirement']

wp1/selection/simple_builder.py
```diff
@@ -0,0 +1,49 @@
+import urllib.parse
+from wp1.selection.abstract_builder import AbstractBuilder
+
+
+class SimpleBuilder(AbstractBuilder):
```
audiodude, line 5: Please move this file to `wp1/selection/models/simple.py`

wp1/selection/simple_builder_test.py
```diff
@@ -0,0 +1,71 @@
+import unittest
+from wp1.selection.simple_builder import SimpleBuilder
+
+
+class ValidateArticleNameTest(unittest.TestCase):
```
audiodude, line 5: s/ValidateArticleNameTest/SimpleBuilderTest/

wp1/selection/simple_builder.py
```diff
@@ -0,0 +1,49 @@
...
+            invalid_article_names.append(decoded_item)
+            is_valid = False
+      if is_valid:
+        article_name = decoded_item.replace(
+            "https://en.wikipedia.org/wiki/",
+            "").replace("https://en.wikipedia.org/w/index.php?title=", "")
+        valid_article_names.append(article_name)
+    return (valid_article_names, invalid_article_names, forbiden_chars)
+
+  def build(self, content_type, **params):
```
audiodude, line 37: Currently, this ignores the content_type parameter. Since it currently can only output `text/tab-separated-values`, it should raise a `ValueError` if the content_type doesn't match that.

wp1/selection/simple_builder.py
```diff
@@ -0,0 +1,49 @@
...
+    if len(params) == 1:
+      if 'list' in params.keys():
```
audiodude, line 39: nit: I don't think you need .keys() here.

wp1/selection/simple_builder.py
```diff
@@ -0,0 +1,49 @@
+import urllib.parse
+from wp1.selection.abstract_builder import AbstractBuilder
+
+
+class SimpleBuilder(AbstractBuilder):
+
+  def _validate_list(self, **params):
```
audiodude, line 7: Did you remove this function from the selection module? I don't see that change. It's no longer needed there.

audiodude (changes requested):
Good work on moving the function. There are a few subtle cases that I think we should take care of.

mahakporwal02:
@audiodude  Tests are failing because I have not made changes in `wp1.web.selection.py` yet. The route `/selection/simple` calls `validate_list` func which I have removed in the last commit.
Should I make changes there now?

wp1/selection/models/simple_builder_test.py
```diff
@@ -55,17 +55,24 @@ def test_build(self):
     actual = simple_test_builder.build('text/tab-separated-values', **params)
     self.assertEqual(b'Eiffel_Tower\nStatue#of_Liberty\nLibertas', actual)
 
+  def test_build_unrecognized_content_type(self):
+    simple_test_builder = SimpleBuilder()
+    actual = simple_test_builder.build('text/tab-separated-values',
```
audiodude, line 60: nit: you don't need this line for this test.

audiodude (changes requested):
> @audiodude Tests are failing because I have not made changes in `wp1.web.selection.py` yet. The route `/selection/simple` calls `validate_list` func which I have removed in the last commit.
> Should I make changes there now?

Yes, definitely.

wp1/selection/models/simple_builder.py
```diff
@@ -1,4 +1,5 @@
 import urllib.parse
+from wp1.constants import CONTENT_TYPE_TO_EXT
```
audiodude, line 2: I don't think you need this import.

wp1/selection/models/simple_builder.py
```diff
@@ -9,41 +10,44 @@ def _validate_list(self, **params):
...
+      )
+    raise ValueError('Additional unnecessary params present')
 
   def validate(self, **params):
-    valid_names, invalid_names, forbiden_chars = self._validate_list(**params)
-    if forbiden_chars[0] != 'Empty List':
-      forbiden_chars.insert(
-          0, 'The list contained the following invalid characters: ')
-    return (valid_names, invalid_names, forbiden_chars)
+    return self._validate_list(**params)
```
audiodude, line 53: Since this method is now just calling the _validate_list function, can you move the body of _validate_list here and get rid of the extra function?

audiodude (changes requested):
A couple more things

wp1/web/selection.py
```diff
@@ -12,10 +12,11 @@ def create():
   list_name = data['list_name']
   articles = data['articles']
   project = data['project']
+  params = {'list': articles.split('\n')}
   if not articles or not list_name or not project:
     flask.abort(400)
-  valid_names, invalid_names, forbiden_chars = selections.validate_list(
-      articles)
+  simple_builder = SimpleBuilder()
+  valid_names, invalid_names, forbiden_chars = simple_builder.validate(**params)
```
audiodude, line 19: s/forbiden/forbidden/

wp1/web/selection_test.py
```diff
@@ -17,7 +17,9 @@ class ProjectTest(unittest.TestCase):
       "items": {
           'valid': ['Eiffel_Tower'],
           'invalid': ['Statue_of#Liberty'],
-          'forbiden_chars': ['#']
+          'forbiden_chars': [
```
audiodude, line 20: Looks like 'forbiden' made it into the API? Can you change it here and in openapi.yml?

audiodude: Actually, this should be `errors` now, not `forbiden_chars`. Can you file an issue so that they're displayed appropriately on the frontend?

audiodude (changes requested):
Very close to being done, good work.

audiodude:
Your code coverage is failing for the same reason mine was on the last PR: the `wp1/selection/models` directory needs an empty `__init__.py` file in it so that it can be considered a module and nosetests can find the tests.

wp1/web/selection.py
audiodude: Please fix.

audiodude (changes requested):

audiodude (approved):
LGTM

PR #401, opened 2021-08-05
added method to save builder

### LINKED ISSUE #383: Saving Simple lists
We need to actually save the Simple lists, both in the wp10 database and in the backend bucket storage.

It would be nice if there was a single method that the web frontend could call, like `wp1.logic.selection.persist_simple_list`.

I see that method having the following arguments:

```
def persist_simple_list(user_id, name, project, data):
```

Where `data` is the newline separated article names that have already been validated.

It will need to save to the database twice, once to snatch an ID for the list so that it's 'hash' aka external id can be calculated, and again once it has been persisted to bucket storage, to store the region/bucket/object_id. If the bucket storage fails, the `persist_simple_list` method should clean up the entry in the database.

So the basic steps are:

1. Create a Selection model with the information passed to the function
2. Save the Selection in the database with barebones information just to get an id. Store the id on the Selection model.
3. Call a new `object_id` `@property` method on the Selection model. This will call the corresponding `hash` `@property` method on the Selection model which will calculate the hash/external ID using a secret hash salt (stored in credentials.py) and the ID of the selection. The `object_id` should be of the form: `/selection/simple/<user_id>/<hash>.tsv`
4. Upload the data to the bucket storage using KiwixStorage, as imported from a storage.py class that exists alongside db.py in wp1. (Same basic pattern of failing when credentials.py doesn't exist).
5. Update the Selection's entry in the database with the region/bucket/object_id and hash.
6. Return the hash, so that the backend can tell the frontend and the user can be redirected to the "show" page for the list (which hasn't been implemented yet).

mahakporwal02:
This fixes #383.


audiodude (changes requested):
This looks good, we'll need to also update the web frontend method to use this, possibly through another method that takes the name, user_id, project and params and constructs a Builder object, then passes it to this method.

audiodude (changes requested):
Sorry I missed it, but you should also have a test for this function of course.

db/yoyo.ini
```diff
@@ -0,0 +1,5 @@
+[DEFAULT]
```
audiodude, line 1: Delete this file. When yoyo asks you if you want to save the configuration, say no.

wp1/selection/models/simple_builder.py
```diff
@@ -47,3 +50,30 @@ def validate(self, **params):
       errors.append('The list contained the following invalid characters: ' +
                     ', '.join(forbidden_chars))
     return (valid_article_names, invalid_article_names, errors)
+
+  def insert_builder(self, data, wp10db):
```
audiodude, line 54: This should not be part of the Builder class. It should be its own method in `wp1.logic.builder`. Also, the method that creates the Builder model and the method that inserts it into the database should be two separate methods.

wp1/selection/models/simple_builder.py
```diff
@@ -1,4 +1,7 @@
+import attr
 import urllib.parse
+from wp1.models.wp10.builder import Builder
+from flask import session
```
audiodude, line 4: You should never import flask libraries outside of the `web` module. Instead, pass the user_id directly.

wp1/selection/models/simple_builder_test.py
```diff
@@ -1,5 +1,11 @@
+from datetime import datetime
```
audiodude, line 1: Please sort these imports as follows:

library imports (datetime, unittest, etc)

third party library imports (attr, flask, etc)

local imports (wp1.*)

wp1/selection/models/simple_builder_test.py
```diff
@@ -1,5 +1,11 @@
...
 import unittest
+from unittest.mock import patch
+from wp1.web.app import create_app
 from wp1.selection.models.simple_builder import SimpleBuilder
+from flask import session
```
audiodude, line 6: Again, you should not be importing flask libraries outside the `web` directory.

wp1/selection/models/simple_builder_test.py
```diff
@@ -74,3 +80,37 @@ def test_validate_empty_items(self):
...
+          'sub': '1234'
+      }
+  }
+
+  def setUp(self):
+    super().setUp()
+
+  @patch('wp1.models.wp10.selection.utcnow',
+         return_value=datetime(2020, 12, 25, 10, 55, 44))
+  def test_insert(self, mock_utcnow):
```
audiodude, line 100: If you follow my advice and make a separate `insert_builder` function in `wp1.logic.builder` that just inserts into the database, then you can test that separately.

wp1/web/selection.py
```diff
@@ -17,6 +18,8 @@ def create():
     flask.abort(400)
   simple_builder = SimpleBuilder()
   valid_names, invalid_names, errors = simple_builder.validate(**params)
+  wp10db = wp10_connect()
```
audiodude, line 21: In flask methods, we use get_db, see the other web methods for an example:

https://github.com/openzim/wp1/blob/main/wp1/web/db.py#L14

wp1/web/selection.py
```diff
@@ -17,6 +18,8 @@ def create():
...
+  simple_builder.insert_builder(data, wp10db)
```
audiodude, line 22: You shouldn't insert the builder until you know the params are valid.

wp1/web/selection.py
```diff
@@ -27,3 +30,19 @@ def create():
         }
     }
   return {"success": True, "items": {}}
+
+
+@selection.route('/simple/lists')
+@authenticate
+def get_list_data():
```
audiodude, line 37: Can we do this in a separate PR, getting the lists? Including the Vue changes, in a separate PR?

wp1/web/selection_test.py
```diff
@@ -1,4 +1,6 @@
 import unittest
+from wp1.base_db_test import BaseWpOneDbTest
```
audiodude, line 2: Tests in web should inherit from BaseWebTestCase: 

https://github.com/openzim/wp1/blob/main/wp1/web/projects_test.py#L8

audiodude (changes requested):

wp1/selection/models/simple_builder_test.py
```diff
@@ -1,8 +1,8 @@
-import unittest
+from wp1.base_db_test import BaseWpOneDbTest
 from wp1.selection.models.simple_builder import SimpleBuilder
 
 
-class SimpleBuilderTest(unittest.TestCase):
+class SimpleBuilderTest(BaseWpOneDbTest):
```
audiodude, line 5: This does not need to be a BaseWpOneDbTest because it doesn't use the "WpOneDb" at all.

wp1/web/selection.py
```diff
@@ -1,5 +1,8 @@
-from wp1.selection.models.simple_builder import SimpleBuilder
+from wp1.web.db import get_db
```
audiodude, line 1: Please sort these imports.

wp10_test.up.sql
```diff
@@ -93,14 +93,14 @@ CREATE TABLE `users` (
...
 CREATE TABLE `builders` (
-  s_id INTEGER NOT NULL PRIMARY KEY AUTO_INCREMENT,
-  s_name VARBINARY(255) NOT NULL,
-  s_user_id INTEGER NOT NULL,
-  s_project VARBINARY(255) NOT NULL,
-  s_model VARBINARY(255) NOT NULL,
-  s_params BLOB,
-  s_created_at BINARY(20),
-  s_updated_at BINARY(20)
+  b_id INTEGER NOT NULL PRIMARY KEY AUTO_INCREMENT,
```
audiodude, line 96: Good catch!

audiodude (changes requested):
Looking good, just a few comments on naming and style

wp1/logic/builder.py
```diff
@@ -0,0 +1,26 @@
+import attr
```
audiodude, line 1: Import order should be:

Python standard librarires (json)
empty line
Third party libraries (attr)
empty line
wp1 libraries

And they should be in alphabetical order where possible.

wp1/logic/builder_test.py
```diff
@@ -0,0 +1,42 @@
...
+      'b_params': b'{"list": ["a", "b", "c"]}',
+      'b_created_at': None,
+      'b_updated_at': None
+  }
+
+  @patch('wp1.models.wp10.builder.utcnow', return_value=None)
+  def test_save_builder(self, mock_utcnow):
+    save_builder(self.wp10db, 'My Builder', '1234', 'en.wikipedia.fake',
+                 'a\nb\nc')
+    with self.wp10db.cursor() as cursor:
```
audiodude, line 30: nit: why not extract this as a function in this file `_get_builder_by_user_id` since you use it in both test methods?

wp1/logic/builder_test.py
```diff
@@ -0,0 +1,42 @@
...
+      cursor.execute('SELECT * FROM builders WHERE b_user_id=%(b_user_id)s',
+                     {'b_user_id': '1234'})
+      db_lists = cursor.fetchone()
+    self.assertEqual(self.expected, db_lists)
```
audiodude, line 34: You should set utcnow to return a real value and test that created_at and updated_at have the right value.

wp1/web/selection.py
```diff
@@ -1,6 +1,9 @@
-from wp1.selection.models.simple_builder import SimpleBuilder
 import flask
+from flask import session
 from wp1.web import authenticate
```
audiodude, line 3: nit: newline before wp1 imports.

audiodude (changes requested):
Just a few tiny nits, otherwise this looks really good!

audiodude (approved):
Thanks for all your work on this, LGTM.

PR #408, opened 2021-08-10
frontend displays lists in MyLists

### LINKED ISSUE #406: Make sure that the MyLists display accurately reflects simple lists
The page should do the following:

- List all available formats, not just TSV/XLS
- List only the formats that are currently available/materialized, with a "In progress" display in the place of the URL copy for formats that are not available.
- Allow for copying the URL of each format

Additionally, this requires an API method for returning the appropriate data in the right format.

mahakporwal02:
This fixes #406 
I will be adding cypress tests in the following commit.

### DIFF
--- a/wp1-frontend/src/components/MyLists.vue
+++ b/wp1-frontend/src/components/MyLists.vue
@@ -5,11 +5,7 @@
     </div>
     <div class="row">
       <div class=" col-lg-6">
-        <div
-          v-for="item in list"
-          :key="item.project_name"
-          class="card text-center m-5"
-        >
+        <div v-for="item in list" :key="item.id" class="card text-center m-5">
           <div class="card-header">
             {{ item.name }}
           </div>
@@ -29,7 +25,7 @@
                   Copy
                 </button>
               </div>
-              <a :href="item.link" class="btn btn-primary ml-3" download
+              <a :href="item.url" class="btn btn-primary ml-3" download
                 >Download {{ extension }}</a
               >
             </div>
--- a/wp1-frontend/src/components/MyLists.vue
+++ b/wp1-frontend/src/components/MyLists.vue
@@ -5,14 +5,22 @@
     </div>
     <div class="row">
       <div class=" col-lg-6">
-        <div v-for="item in list" :key="item.name" class="card text-center m-5">
+        <div
+          v-for="item in list"
+          :key="item.project_name"
+          class="card text-center m-5"
+        >
           <div class="card-header">
             {{ item.name }}
           </div>
           <div class="card-body ">
             <h5 class="card-title">{{ item.project }}</h5>
-            <div class="input-group col-sm-5 mx-auto mb-3">
-              <input :id="item.name" :value="item.link" class="form-control" />
+            <div
+              v-for="extension in item.extension"
+              :key="extension"
+              class="input-group col-sm-9 mx-auto mb-3"
+            >
+              <input :id="item.name" :value="item.url" class="form-control" />
               <div class="input-group-append">
                 <button
                   class="btn btn-outline-secondary"
@@ -21,18 +29,10 @@
                   Copy
                 </button>
               </div>
-            </div>
-            <div class="col-sm-8 mx-auto">
-              <a :href="item.link" class="btn btn-primary m-2" download
-                >Download .TSV</a
-              >
-              <a :href="item.link" class="btn btn-secondary" download
-                >Download .XLS</a
+              <a :href="item.link" class="btn btn-primary ml-3" download
+                >Download {{ extension }}</a
               >
             </div>
-          </div>
-          <div class="card-footer text-muted">
-            {{ item.timestamp }}
           </div>
         </div>
       </div>
@@ -52,32 +52,7 @@ export default {
   name: 'MyLists',
   data: function() {
     return {
-      list: [
-        {
-          name: 'My-List1',
-          project: 'Project1',
-          link: '<URL1>',
-          timestamp: '2 days ago '
-        },
-        {
-          name: 'My-List2',
-          project: 'Project2',
-          link: '<URL2>',
-          timestamp: '2 days ago'
-        },
-        {
-          name: 'My-List3',
-          project: 'Project3',
-          link: '<URL3>',
-          timestamp: '2 days ago'
-        },
-        {
-          name: 'My-List4',
-          project: 'Project4',
-          link: '<URL4>',
-          timestamp: '2 days ago'
-        }
-      ]
+      list: []
     };
   },
   computed: {
@@ -90,7 +65,21 @@ export default {
       var copyText = document.getElementById(id);
       copyText.select();
       document.execCommand('copy');
+    },
+    getLists: async function() {
+      const response = await fetch(
+        `${process.env.VUE_APP_API_URL}/selection/simple/lists`,
+        {
+          headers: { 'Content-Type': 'application/json' },
+          credentials: 'include'
+        }
+      );
+      var data = await response.json();
+      this.list = data.list_data;
     }
+  },
+  created: function() {
+    this.getLists();
   }
 };
 </script>
--- a/wp1-frontend/src/components/MyLists.vue
+++ b/wp1-frontend/src/components/MyLists.vue
@@ -16,11 +16,11 @@
               :key="extension"
               class="input-group col-sm-9 mx-auto mb-3"
             >
-              <input :id="item.name" :value="item.url" class="form-control" />
+              <input :id="item.id" :value="item.url" class="form-control" />
               <div class="input-group-append">
                 <button
                   class="btn btn-outline-secondary"
-                  v-on:click="copyText(item.name)"
+                  v-on:click="copyText(item.id)"
                 >
                   Copy
                 </button>
--- a/wp1-frontend/src/components/MyLists.vue
+++ b/wp1-frontend/src/components/MyLists.vue
@@ -12,21 +12,25 @@
           <div class="card-body ">
             <h5 class="card-title">{{ item.project }}</h5>
             <div
-              v-for="extension in item.extension"
-              :key="extension"
+              v-for="selection in item.selections"
+              :key="selection.s_id"
               class="input-group col-sm-9 mx-auto mb-3"
             >
-              <input :id="item.id" :value="item.url" class="form-control" />
+              <input
+                :id="selection.s_id"
+                :value="selection.selection_url"
+                class="form-control"
+              />
               <div class="input-group-append">
                 <button
                   class="btn btn-outline-secondary"
-                  v-on:click="copyText(item.id)"
+                  v-on:click="copyText(selection.s_id)"
                 >
                   Copy
                 </button>
               </div>
               <a :href="item.url" class="btn btn-primary ml-3" download
-                >Download {{ extension }}</a
+                >Download {{ selection.content_type }}</a
               >
             </div>
           </div>
@@ -71,7 +75,7 @@ export default {
         }
       );
       var data = await response.json();
-      this.list = data.list_data;
+      this.list = data.builders;
     }
   },
   created: function() {
--- a/wp1/logic/builder.py
+++ b/wp1/logic/builder.py
@@ -30,8 +30,10 @@ def insert_builder(wp10db, builder):
 
 def get_lists(wp10db, user_id):
   with wp10db.cursor() as cursor:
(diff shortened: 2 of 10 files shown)


wp1/logic/builder.py
```diff
@@ -26,3 +26,13 @@ def insert_builder(wp10db, builder):
         VALUES (%(b_name)s, %(b_user_id)s, %(b_project)s, %(b_params)s, %(b_model)s, %(b_created_at)s, %(b_updated_at)s)
       ''', attr.asdict(builder))
   wp10db.commit()
+
+
+def get_lists(wp10db, user_id):
+  with wp10db.cursor() as cursor:
+    cursor.execute('SELECT * FROM builders WHERE b_user_id=%(b_user_id)s',
```
audiodude, line 33: You need to JOIN on the selections on s_builder_id = b_id, so that you know what selections are available for each list/builder.

audiodude (changes requested):
Not an exhaustive list, but the first few things I noticed.

wp1/logic/builder.py
```diff
@@ -26,3 +26,15 @@ def insert_builder(wp10db, builder):
...
+
+def get_lists(wp10db, user_id):
+  with wp10db.cursor() as cursor:
+    cursor.execute(
+        '''SELECT * FROM selections
+                      INNER JOIN builders ON selections.s_builder_id=builders.b_id
+                      WHERE b_user_id=%(b_user_id)s''', {'b_user_id': user_id})
+    db_lists = cursor.fetchall()
+    if not db_lists:
+      return None
```
audiodude, line 39: I think it would simplify the API if you just returned empty list for this scenario.

wp1/logic/builder_test.py
```diff
@@ -49,3 +49,19 @@ def test_insert_builder(self, mock_utcnow):
     insert_builder(self.wp10db, self.builder)
     db_lists = self._get_builder_by_user_id()
     self.assertEqual(self.expected, db_lists)
+
+  @patch('wp1.models.wp10.builder.utcnow',
+         return_value=datetime.datetime(2019, 12, 25, 4, 44, 44))
+  def test_get_lists(self, mock_utcnow):
+    save_builder(self.wp10db, 'My Builder', '1234', 'en.wikipedia.fake',
```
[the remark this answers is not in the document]
audiodude: Please fix.

wp1-frontend/src/components/MyLists.vue
```diff
@@ -5,14 +5,18 @@
...
           <div class="card-body ">
             <h5 class="card-title">{{ item.project }}</h5>
-            <div class="input-group col-sm-5 mx-auto mb-3">
-              <input :id="item.name" :value="item.link" class="form-control" />
+            <div
+              v-for="extension in item.extension"
+              :key="extension"
+              class="input-group col-sm-9 mx-auto mb-3"
+            >
+              <input :id="item.name" :value="item.url" class="form-control" />
```
audiodude, line 19: I don't think you can use the name for the id property, because lists are not guaranteed to have unique names.

wp1/web/selection.py
```diff
@@ -34,3 +35,22 @@ def create():
...
+
+
+@selection.route('/simple/lists')
+@authenticate
+def get_list_data():
+  list_data = []
+  wp10db = get_db('wp10db')
+  user_id = session['user']['identity']['sub']
+  article_lists = get_lists(wp10db, user_id)
+  for article_data in article_lists:
```
audiodude, line 47: If there are multiple selections for a builder, won't this return two "article_lists" for that builder, and you will have duplicates on the UI screen instead of having multiple extensions for one list?

wp1/web/selection_test.py
```diff
@@ -98,3 +110,26 @@ def test_selection_unauthorized_user(self):
...
+        cursor.execute('''INSERT INTO builders
+        (b_name, b_user_id, b_project, b_model)
+        VALUES ('list_name', '11', 'project_name', 'model')
+      ''')
+        cursor.execute(
+            '''INSERT INTO selections VALUES (1, 1, "tsv", '20201225105544')''')
+      with client.session_transaction() as sess:
+        sess['user'] = self.USER
+      rv = client.get('/v1/selection/simple/lists')
+      print(rv.get_json())
```
audiodude, line 134: nit: remove print statement

wp1/web/selection_test.py
```diff
@@ -98,3 +110,26 @@ def test_selection_unauthorized_user(self):
...
+      self.assertEqual(rv.get_json(), self.expected)
```
audiodude, line 135: nit: these are reversed, it should always be `assertEqual(expected, actual)`.

wp1/web/selection_test.py
```diff
@@ -98,3 +110,26 @@ def test_selection_unauthorized_user(self):
                            'project': ''
                        })
     self.assertEqual('401 UNAUTHORIZED', rv.status)
+
+  @patch('wp1.web.selection.get_lists',
```
audiodude, line 114: Why do you insert into the database, if you're also mocking the return value of get_lists? Only one or the other should be necessary, and I would suggest using real database data.

wp1/web/selection_test.py
```diff
@@ -98,3 +110,26 @@ def test_selection_unauthorized_user(self):
...
+         return_value=[{
+             'b_id': 11,
+             'b_name': b'list_name',
+             'b_project': b'project_name',
+             's_content_type': b"tsv",
+         }])
+  def test_get_list_data(self, mockes_utcnow):
```
audiodude, line 121: Per my earlier comment: write a test here, using real get_lists data, that if there are two selections for a builder that the 'list_data' array only contains one list entry. It will probably fail right now.

audiodude (changes requested):
You also need to update openapi.yml for `/simple/lists` right?

wp1/logic/builder.py
```diff
@@ -26,3 +26,15 @@ def insert_builder(wp10db, builder):
         VALUES (%(b_name)s, %(b_user_id)s, %(b_project)s, %(b_params)s, %(b_model)s, %(b_created_at)s, %(b_updated_at)s)
       ''', attr.asdict(builder))
   wp10db.commit()
+
+
+def get_lists(wp10db, user_id):
+  with wp10db.cursor() as cursor:
+    cursor.execute(
+        '''SELECT * FROM selections
+                      INNER JOIN builders ON selections.s_builder_id=builders.b_id
```
audiodude, line 35: I thought about this, and I think you want an RIGHT OUTER JOIN on builders, so that you can get the builder even if there are no selections.

wp1/web/selection_test.py
```diff
@@ -25,14 +25,36 @@ class SelectionTest(BaseWebTestcase):
...
-      'list_data': [{
-          'content_type': "tsv",
-          'extension': ['tsv'],
-          'name': 'list_name',
-          'project': 'project_name',
-          'url': 'https://www.example.com/<id>',
-          'id': 11
-      }]
+      'list_data': {
+          '1': {
```
audiodude, line 29: Although it made sense to have the builders be a dictionary with ids as keys for how you constructed it, for the API it makes more sense for "id" to be just another field on `list_data` and `list_data` itself to be an array.

wp1/web/selection_test.py
```diff
@@ -25,14 +25,36 @@ class SelectionTest(BaseWebTestcase):
...
   expected = {
-      'list_data': [{
-          'content_type': "tsv",
-          'extension': ['tsv'],
-          'name': 'list_name',
-          'project': 'project_name',
-          'url': 'https://www.example.com/<id>',
-          'id': 11
-      }]
+      'list_data': {
```
audiodude, line 28: nit: s/list_data/builders/

wp1/web/selection_test.py
```diff
@@ -111,25 +133,37 @@ def test_selection_unauthorized_user(self):
...
       ''')
         cursor.execute(
             '''INSERT INTO selections VALUES (1, 1, "tsv", '20201225105544')''')
+      self.wp10db.commit()
+      with client.session_transaction() as sess:
+        sess['user'] = self.USER
+      rv = client.get('/v1/selection/simple/lists')
+      self.assertEqual(self.expected, rv.get_json())
+
+  def test_list_with_multiple_selections(self):
```
audiodude, line 152: You should have the same tests here I suggested for the get_lists methods: builder with one selection (you already have this), builder with multiple selections (you already have this), builder with no selections (missing) and no builders (missing).

wp1-frontend/src/components/MyLists.vue
```diff
@@ -16,11 +16,11 @@
               :key="extension"
               class="input-group col-sm-9 mx-auto mb-3"
             >
-              <input :id="item.name" :value="item.url" class="form-control" />
+              <input :id="item.id" :value="item.url" class="form-control" />
```
audiodude, line 19: Although I suggested that each `item` should have an `id`, property, currently they don't. Did you test your frontend? I don't think this is working.

audiodude (changes requested):
Let me know if you have any questions on my comments, thanks!

wp1-frontend/src/components/MyLists.vue
```diff
@@ -5,35 +5,31 @@
...
+        <div v-for="item in list" :key="item.id" class="card text-center m-5">
           <div class="card-header">
             {{ item.name }}
           </div>
           <div class="card-body ">
             <h5 class="card-title">{{ item.project }}</h5>
-            <div class="input-group col-sm-5 mx-auto mb-3">
-              <input :id="item.name" :value="item.link" class="form-control" />
+            <div
+              v-for="extension in item.extension"
```
audiodude, line 15: Actually now that I think of it, this won't work either, because there's no `extension` field in the item, it has a list of selections now.

wp1/logic/builder.py
```diff
@@ -26,3 +26,37 @@ def insert_builder(wp10db, builder):
...
+            'project': data['b_project'].decode('utf-8'),
+            'selections': []
+        }
+      if data['s_id']:
+        result[data['b_id']]['selections'].append({
+            's_id': data['s_id'].decode('utf-8'),
+            'content_type': data['s_content_type'].decode('utf-8'),
+            'selection_url': 'https://www.example.com/<id>'
+        })
+    for list_data, value in result.items():
```
audiodude, line 53: s/list_data/id_/

wp1/logic/builder.py
```diff
@@ -26,3 +26,37 @@ def insert_builder(wp10db, builder):
...
+      article_data.append({
+          'id': list_data,
+          'name': value['name'],
+          'project': value['project'],
+          'selections': value['selections']
+      })
+    if not article_data:
```
audiodude, line 60: This should not be necessary. If the list is empty, you can just return it.

wp1/logic/builder_test.py
```diff
@@ -28,24 +29,131 @@ class BuilderDbUpdateTest(BaseWpOneDbTest):
...
+         return_value=datetime.datetime(2019, 12, 25, 4, 44, 44))
+  def test_get_lists_with_multiple_selections(self, mock_utcnow):
+    with self.wp10db.cursor() as cursor:
+      cursor.execute(
+          '''INSERT INTO selections VALUES (1, 1, 'tsv', '20191225044444')''')
+      cursor.execute(
+          '''INSERT INTO selections VALUES (2, 1, "xls", '20201225105544')''')
+      cursor.execute(
+          '''INSERT INTO builders
+        (b_name, b_user_id, b_project, b_params, b_model, b_created_at, b_updated_at)
```
audiodude, line 117: nit: line length.

wp1/logic/builder_test.py
```diff
@@ -28,24 +29,131 @@ class BuilderDbUpdateTest(BaseWpOneDbTest):
...
+    article_data = get_lists(self.wp10db, '1234')
+    self.assertEqual(self.expected_lists_with_multiple_selections, article_data)
+
+  @patch('wp1.models.wp10.builder.utcnow',
+         return_value=datetime.datetime(2019, 12, 25, 4, 44, 44))
+  def test_get_lists_with_no_selections(self, mock_utcnow):
+    with self.wp10db.cursor() as cursor:
+      cursor.execute(
+          '''INSERT INTO builders
+        (b_name, b_user_id, b_project, b_params, b_model, b_created_at, b_updated_at)
```
audiodude, line 130: nit: line length.

audiodude (changes requested):
Looking great, just a few very minor comments.

audiodude (approved):
LGTM

PR #420, opened 2021-08-22
New process for building production images manually

audiodude:
Since Docker Hub no longer automatically builds images on repo update (on their free plan), we must have a process in place for building and pushing the production images manually.

While we're add it, add a .dockerignore file to speed up the process dramatically (less data transferred for transient directories).

This is actually perhaps a bit more satisfying than waiting for Docker Hub anyways.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+wp1-frontend/node_modules
+log
+venv
--- a/README.md
+++ b/README.md
@@ -253,6 +253,11 @@ to something like:
 
 # Updating production
 
+- Since Docker Hub no longer auto builds images, you must build the images yourself.
+  From the wp1 directory, run the following commands to build and push the images:
+  - `git checkout main`
+  - `git pull origin main`
+  - `./build_production_images.sh`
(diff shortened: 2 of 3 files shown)


kelson42:
@audiodude please have a look how we have solved this for openzin/mwoffliner for example

audiodude:
@kelson42 I just looked at the main repo and the docker directory, but I don't see any pointers to how you build images for docker hub? Or if docker hub is being used at all (maybe that's the point that I'm missing)?

The idea here is that we build and push the production images locally. When you say "solved this" what is "this"?

kelson42:
Use https://github.com/openzim/docker-publish-action

audiodude:
Ah okay, that makes sense thanks. I'll open an issue.

kelson42:
Considering that this are really internal/operational Docker images, we might as well consider not uploading then anymore to Docker hub... and only to Github.

audiodude:
As long as I can pull them easily in the docker-compose.yml file, it doesn't matter to me.

PR #437, opened 2021-08-25
list selections displayed in the form of table

### LINKED ISSUE #433: List selections in a form of a table
Would be better IMO to have the table like in an sheet, something doable easily with https://datatables.net/.

mahakporwal02:
This fixes #433.

### DIFF
--- a/wp1-frontend/src/components/MyLists.vue
+++ b/wp1-frontend/src/components/MyLists.vue
@@ -28,7 +28,10 @@
                 {{ item.selections[0].s_updated_at }}
               </td>
               <td v-if="item.selections.length">
-                <a class="btn btn-primary mb-2 mr-2" download
+                <a
+                  :href="item.selections[0].url"
+                  class="btn btn-primary mb-2 mr-2"
+                  download
                   >Download {{ item.selections[0].extension }}</a
                 >
               </td>
--- a/wp1-frontend/src/components/MyLists.vue
+++ b/wp1-frontend/src/components/MyLists.vue
@@ -3,46 +3,37 @@
     <div>
       <SecondaryNav></SecondaryNav>
     </div>
-    <div class="row">
+    <div class="row ml-4">
       <div>
-        <div v-for="item in list" :key="item.id" class="card text-center m-5">
-          <div class="card-header">
-            {{ item.name }}
-          </div>
-          <div class="card-body ">
-            <h5 class="card-title">{{ item.project }}</h5>
-            <table v-if="item.selections.length" class="card-table table">
-              <thead>
-                <tr>
-                  <th scope="col">Name</th>
-                  <th scope="col">Created At</th>
-                  <th scope="col">Updated At</th>
-                  <th scope="col">Generated At</th>
-                  <th scope="col">Update</th>
-                  <th scope="col">Download</th>
-                </tr>
-              </thead>
-              <tbody v-for="selection in item.selections" :key="selection.s_id">
-                <tr>
-                  <td>selection_name</td>
-                  <td>created at</td>
-                  <td>updated at</td>
-                  <td>generated at</td>
-                  <td>
-                    <button class="btn btn-primary ml-3">
-                      Update
-                    </button>
-                  </td>
-                  <td>
-                    <a :href="item.url" class="btn btn-primary ml-3" download
-                      >Download {{ selection.content_type }}</a
-                    >
-                  </td>
-                </tr>
-              </tbody>
-            </table>
-          </div>
-        </div>
+        <table class="table table-hover">
+          <thead>
+            <tr>
+              <th scope="col">Name</th>
+              <th scope="col">Created At</th>
+              <th scope="col">Updated At</th>
+              <th scope="col">Generated At</th>
+              <th scope="col">Download</th>
+            </tr>
+          </thead>
+          <tbody v-for="item in list" :key="item.id">
+            <tr v-if="item.selections.length < 1">
+              <td>{{ item.name }}</td>
+              <td>{{ item.created_at }}</td>
+              <td>{{ item.updated_at }}</td>
+            </tr>
+            <tr v-else v-for="selection in item.selections" :key="selection.id">
+              <td>{{ item.name }}</td>
+              <td>{{ item.created_at }}</td>
+              <td>{{ item.updated_at }}</td>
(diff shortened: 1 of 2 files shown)


wp1-frontend/src/components/MyLists.vue
```diff
@@ -85,6 +76,7 @@ export default {
       );
       var data = await response.json();
       this.list = data.builders;
+      console.log(this.list);
```
audiodude, line 79: nit: remove console.log

wp1-frontend/src/components/MyLists.vue
```diff
@@ -3,46 +3,37 @@
...
+            </tr>
+          </thead>
+          <tbody v-for="item in list" :key="item.id">
+            <tr v-if="item.selections.length < 1">
+              <td>{{ item.name }}</td>
+              <td>{{ item.created_at }}</td>
+              <td>{{ item.updated_at }}</td>
+            </tr>
+            <tr v-else v-for="selection in item.selections" :key="selection.id">
+              <td>{{ item.name }}</td>
```
audiodude, line 25: It would be better if the things that were the same for each builder were set with rowspan=2 instead of being repeated.

wp1-frontend/src/components/MyLists.vue
```diff
@@ -3,46 +3,37 @@
...
+              <td>{{ item.created_at }}</td>
+              <td>{{ item.updated_at }}</td>
+              <td>{{ selection.s_updated_at }}</td>
+              <td>
+                <a class="btn btn-primary mb-2 mr-2" download
+                  >Download {{ selection.content_type }}</a
```
audiodude, line 31: You're showing the content-type again when it should be the extension.

wp1-frontend/src/components/MyLists.vue
```diff
@@ -3,46 +3,37 @@
...
+              <td>{{ item.created_at }}</td>
+              <td>{{ item.updated_at }}</td>
+            </tr>
+            <tr v-else v-for="selection in item.selections" :key="selection.id">
+              <td>{{ item.name }}</td>
+              <td>{{ item.created_at }}</td>
+              <td>{{ item.updated_at }}</td>
+              <td>{{ selection.s_updated_at }}</td>
+              <td>
+                <a class="btn btn-primary mb-2 mr-2" download
```
audiodude, line 30: You need the href of the button to be the download URL.

wp1/logic/builder.py
```diff
@@ -89,12 +93,19 @@ def get_builders_with_selections(wp10db, user_id):
...
+            # 's_updated_at': datetime.datetime.strptime(s_updated_at, TS_FORMAT_WP10)
         })
     for id_, value in builders.items():
       result.append({
           'id': id_,
           'name': value['name'],
           'project': value['project'],
+          'created_at': value['created_at'],
+          'updated_at': value['updated_at'],
+          # 'created_at': datetime.datetime.strptime(value['created_at'], TS_FORMAT_WP10),
```
audiodude, line 107: The fact that you're getting a ValueError with strptime might mean that you never applied the database migrations that turned the timestamps from `BINARY(20)` to `BINARY(14)`, which would explain the extra null values at the end of the string.

But either way, passing a `datetime.datetime` to a dictionary that is going to be serialized to JSON isn't going to work.

audiodude (changes requested):
A few things that need to be cleaned up here.

wp1-frontend/src/components/MyLists.vue
```diff
@@ -3,38 +3,57 @@
...
-              <input
-                :id="selection.id"
-                :value="selection.url"
-                class="form-control"
-              />
-              <div class="input-group-append">
-                <button
-                  class="btn btn-outline-secondary"
-                  v-on:click="copyText(selection.id)"
+              <td v-if="item.selections.length">
```
audiodude, line 43: I think you need the `<td>`s that represent all of the other fields, so that the rowspan will work. You should create cypress fixtures with:

1. A builder with no selections
2. A builder with one selection
3. A builder with multiple selections

And then write tests and verify they all look correct.

wp1/logic/builder.py
```diff
@@ -74,11 +75,14 @@ def get_builders_with_selections(wp10db, user_id):
         builders[b['b_id']] = {
             'name': b['b_name'].decode('utf-8'),
             'project': b['b_project'].decode('utf-8'),
+            'created_at': b['b_created_at'].decode('utf-8'),
```
audiodude, line 78: You had the right idea before, in general. We shouldn't display "20210102105523" as a big garbled string to the user. We should display something that makes more sense.

Maybe you can make a utility function called `wp10_timestamp_human_readable(timestamp_str)` that takes in these strings, parses them according to TS_FORMAT_WP10 and returns a human readable value.

audiodude (changes requested):
Just a few more comments

wp1-frontend/src/components/MyLists.vue
```diff
@@ -3,38 +3,57 @@
...
+            <tr>
+              <th scope="col">Name</th>
+              <th scope="col">Created At</th>
+              <th scope="col">Updated At</th>
+              <th scope="col">Generated At</th>
+              <th scope="col">Download</th>
+            </tr>
+          </thead>
+          <tbody v-for="item in list" :key="item.id">
+            <tr>
```
audiodude, line 19: Maybe add a class to this td, `class="selection-row"` and then style it with `line-height: 2em` so that the buttons line up better with the text?

wp1-frontend/src/components/MyLists.vue
```diff
@@ -3,38 +3,57 @@
...
+              <td v-if="item.selections.length">
+                <a
+                  :href="item.selections[0].url"
+                  class="btn btn-primary mb-2 mr-2"
+                  download
+                  >Download {{ item.selections[0].extension }}</a
+                >
+              </td>
+            </tr>
+            <tr
```
audiodude, line 39: Make this `selection-row` too.

kelson42:
@mahakporwal02 Would be really nice if you could next help to move this PR forward.

kelson42:
@audiodude Still plans to merge this PR?

audiodude:
@kelson42 yes I can look at patching/fixing/merging.

audiodude:
Superseeded by #449.

PR #438, opened 2021-09-06
Allow for editing of list builders

audiodude:
Currently you can click on the name of the builder in the card view in order to open the edit page.

This PR refactors some of the endpoints under `/v1/selection` to be `/v1/builders`.

The CreateSimpleList.vue page is renamed to SimpleList.vue and reworked so that it handles both initial input and editing builders.

Currently builders are NOT re-materialized when they are edited.


audiodude:
Yes, let's merge.

PR #449, opened 2022-01-23
Newer version of displaying selections in a table

### LINKED ISSUE #433: List selections in a form of a table

audiodude:
Fixes #433.

Instead of cards, on the selections (MyLists.vue) page, display a table of selection downloads using [datatables](https://datatables.net).

Since a "list builder" (list of article names for Simple Builder) can have multiple "selections" (materialized download link with a particular format), this table is likely to eventually contain duplicate values for name/created at/updated at. Datatables does not support rowspan, because it breaks its sorting.

Looks like this
Screen Shot of selections table
:


audiodude:
Updated screenshot, with times adjusted to local time and example of a builder without any selections materialized:
Screen Shot of selections table

kelson42:
@audiodude Thx, at a first sight, LGTM.

PR #451, opened 2022-01-23
Only install production npm modules when building frontend app

audiodude:


audiodude:
Needed to move vue-cli-service and it's peer dependencies to the production stanza in package.json

PR #510, opened 2022-09-06
Change builder IDs from monotonically increasing integer primary keys to UUIDs

### LINKED ISSUE #501: Selection ids are incremental
This is not recommended to have incremental primary ids. We should better have UUI... or any (big enough) random id.

audiodude:
Fixes #501 

The migration generates a new ID for all builders, a UUID. It updates all materialized selections to reference the new ID. This means that old URLs of the form `https://api.wp1.openzim.org/v1/builders/15/selection/latest.tsv` will now be something like `https://api.wp1.openzim.org/v1/builders/e1af5673-3929-4c3c-84c5-63e37369b6db/selection/latest.tsv`, and the old URL will no longer function.

A reverse migration is provided, just in case.

The code and tests have been updated to deal with builder IDs as byte strings.

### DIFF
--- a/db/migrations/20220905_01_4yJzk-change-builder-id-column.py
+++ b/db/migrations/20220905_01_4yJzk-change-builder-id-column.py
@@ -1,0 +1,43 @@
+"""
+Change builder id column
+"""
+import uuid
+
+from yoyo import step
+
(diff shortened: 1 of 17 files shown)


audiodude:
I will wait to merge this until tomorrow, when I can deploy it

audiodude:
This has been deployed

PR #513, opened 2022-09-18
Generalize selection/builder CRUD and add placeholder SPARQL page/model

### LINKED ISSUE #476: Frontend of WP1 tool for inputting selection as SPARQL query

audiodude:
Fixes #476 

This PR removes many of the places in the frontend/backend/API where "simple" lists were hardcoded. For example, the endpoints for creating and retrieving builders expected a list of `articles`, whereas that is now generalized to the `params` that are used to create a builder.

I have also generalized retrieving the `Builder` class for a model based on the model's python path (eg `wp1.selection.models.simple` or `wp1.selection.models.sparql`) and added a placeholder `Builder` in the `wp1.selection.models.sparql` model, that always passes validation and builds the list by simply converting the params dict to string.

I have added a new Vue component, `SparqlList`, which is used for creating and editing the aforementioned builders. It provides some help text around the creation of the builder and, looking ahead, provides an additional field for entering the variable in the SPARQL `SELECT` clause that should be interpreted as an article and whose article names should be retrieved.

With all this done, it is now possible to create, edit and materialize (with the placeholder materialization) SPARQL lists, provided that one is aware of the endpoint for doing so (which is `https://wp1.openzim.org/#/selections/sparql`). This endpoint is currently not referenced anywhere in the UI, until SPARQL lists are fully implemented.


audiodude:
One possible downside of this approach is that I copied `SimplList.vue` to `SparqlList.vue` without much consideration for re-use. Most of the page didn't change. There is an opportunity in the future to refactor out common functionality to child components, but for now, it's quicker and less complicated to do it this way.

audiodude:
Hi @rgaudin, it was suggested by @kelson42 that you could do code reviews for the _Wikipedia On Demand_ project. This is the first PR for that project. Can you take a look and add any questions or comments? Thanks!

wp1/selection/models/sparql.py
```diff
@@ -0,0 +1,10 @@
+from wp1.selection.abstract_builder import AbstractBuilder
+
+
+class Builder(AbstractBuilder):
+
+  def build(self, content_type, **params):
```
rgaudin, line 6: Given this serves as an Interface, it'd be a good place to add a docsting describing what the expected output should be.

wp1/selection/models/sparql.py
```diff
@@ -0,0 +1,10 @@
...
+    return str(params).encode('utf-8')
+
+  def validate(self, **params):
```
rgaudin, line 9: Same as above

rgaudin (approved):
Ah github lost my comment text 😤
Anyway this looks good!

I wanted to bring up the fact that the API changed. I wouldn't go through the trouble of making a new version for it ; not sure there are tons of external users anyway.

Please check inline comments ; I think there's a typo in a cypress test.

Will check the README later as many cypress tests fails locally (but I didn't invest much time in the setup)

wp1-frontend/cypress/fixtures/simple_builder.json
```diff
@@ -0,0 +1,8 @@
+{
+  "params": {
+    "list": ["Eiffel_Tower", "Statue_of_Liberty"]
+  },
+  "name": "Builder 1",
+  "model": "wp1.selection.models.sparql",
```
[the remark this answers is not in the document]
audiodude: Done

wp1/selection/models/sparql.py
audiodude: Done

wp1/selection/models/sparql.py
audiodude: Done

audiodude:
This has been deployed.

Issue #540, opened 2022-11-15
Upgrade from deprecated vue-cli to Vite for frontend


audiodude:
The vue-cli project is in maintenance mode, as stated on its homepage: https://cli.vuejs.org/

This means that even if it does receive security updates, it will take longer to get them. We already have a dependabot alert for a package version which is depended on by `babel-loader@8.0.6` (required by `vue-cli@5.0.8`), while the latest babel-loader is v9. It is possible that vue-cli will **never** upgrade this dependency.

For these reasons, we plan on upgrading the frontend build system to Vite, the latest build system provided by the Vue project. It has support for both Vue v2 (which we use) and v3.

PR #541, opened 2022-11-15
Remove usage of deprecated vue-cli and replace with vite

audiodude:
Guide that was followed is here: https://vueschool.io/articles/vuejs-tutorials/how-to-migrate-from-vue-cli-to-vite/

- This results in babel and associated dependencies being removed from the package.json for the frontend.
- The new default port for the frontend developer server is 5173
- The command to run the frontend server has changed to `yarn dev` from `yarn serve`.
- All component dependencies have to be imported with `.vue` extensions.
- The variable for accessing API URL has changed from `process.env.VUE_APP_API_URL` to `import.meta.env.VITE_API_URL`

Smoke test of dev frontend was successful. Assuming Cypress and frontend-build tests pass, this will be considered WAI.

Fixes https://github.com/openzim/wp1/issues/540


audiodude:
Fixes #540

PR #547, opened 2022-12-10
If materialization fails, show error messages in Selection List view as well as on the edit page

### LINKED ISSUE #526: Provide feedback to the user when materialization fails
Currently there is no feedback mechanism when selection materialization fails. If it is the first attempt, the download spinner will show in the selection list indefinitely. If it is a later update, then due to #522 the selection will simply be locked to the old version and the user might not even notice.

We should add a field to the selections table, `status`, which holds a JSON object that has the following fields:

```
{
   status: "OK"  // Or 'CAN_RETRY', or 'FAILED'
   error_messages: [
      "This is error message one",
      "This is error message two",
   ]
}
```

If the status is "OK" then the field is ignored. If the status is CAN_RETRY or FAILED, the list UI should have a way of displaying the error messages to the user. In the specific case of CAN_RETRY (perhaps there was a connection issue with WDQS), there should be the ability for the user to re-queue the selection for materialization.

audiodude:
Fixes #526.

Previous to this PR, when selection materialization failed, there was no feedback to the user. Now, we show an error message in the overall list view, with more information and possibly a retry button on the edit view for the selection that failed.

Screenshots:

error_list

error_retryable

error_fatal

Note that each list of error messages contains a message that we wrote, such as "Could not connect to Wikidata" and the raw message that was returned by the Requests API or other API that failed, such as "HTTPSConnectionPool(host="query.wikidata.org' ...". The latter is not particularly user friendly but will help with debugging errors when they occur.

The "Retry" button on the edit page simply submits the current Builder as an update action, which will always cause materialization to be retried even if nothing has changed. Note that even though we classify most errors as non-retryable (since, unless the Builder params change they will simply fail again), the user could always click "Update" without making changes and invoke this same behavior.

Also includes Cypress tests for the new table entries in the list view and for the edit page.

### DIFF
--- a/wp1/logic/builder.py
+++ b/wp1/logic/builder.py
@@ -164,8 +164,7 @@ def latest_selection_url(wp10db, builder_id, ext):
            FROM selections AS s JOIN builders as b
              ON s.s_builder_id = b.b_id
              AND s.s_version = b.b_current_version
-             AND s.s_content_type = %s
-           WHERE b.b_id = %s
+           WHERE s.s_content_type = %s AND b.b_id = %s
         ''', (content_type, builder_id))
     data = cursor.fetchone()
   if data is None:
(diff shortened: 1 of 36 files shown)


wp1/logic/builder.py
```diff
@@ -174,17 +173,54 @@ def latest_selection_url(wp10db, builder_id, ext):
...
+           WHERE b.b_id = %s AND s.s_status IS NOT NULL
+        ''', (builder_id,))
+    data = cursor.fetchall()
+  if data is None:
+    logger.warning(
+        'Could not find latest selections with errors for builder id=%s',
+        builder_id)
+    return None
+
+  print(data)
```
rgaudin, line 198: Still want this?

wp1/logic/builder.py
```diff
@@ -174,17 +173,54 @@ def latest_selection_url(wp10db, builder_id, ext):
...
+  for db_selection in data:
+    status = {
+        'status':
+            db_selection['s_status'].decode('utf-8'),
+        'ext':
+            CONTENT_TYPE_TO_EXT.get(
+                db_selection['s_content_type'].decode('utf-8'), '???')
+    }
+    status.update(
+        **json.loads(db_selection['s_error_messages'].decode('utf-8')))
```
rgaudin, line 210: Is there a possibility to reach this for a `db_selection` with a `NULL` `s_error_messages` ? That would return an empty string which `json.loads()` would not accept (`JSONDecodeError`).

rgaudin (approved):
LGTM ; haven't actually tested the code obviously and only skimmed through the Cypress stuff but it looks all good.

Spotted what I believe is a debug print and so behavior I'm not sure about; check the inline comments.

wp1/logic/builder.py
audiodude: Good catch, thanks!

wp1/logic/builder.py
audiodude: I don't think that's possible, because the query is for selections where `s_status IS NOT NULL` and the error_messages are set at the same time as the status. Still, I will add some defense in depth just in case. Thanks!

PR #559, opened 2022-12-26
Raise error if SPARQL query results are non-error but empty

### LINKED ISSUE #555: SPARQL based collection of articles is not computed properly
I have a SPARQL request which is:
https://query.wikidata.org/index.html#SELECT%20DISTINCT%20%3Ftopic%20%3FtopicLabel%20%3Farticle%20WHERE%20%7B%0A%20%20%3Fcourse%20wdt%3AP921%20%3Ftopic%3B%0A%20%20%20%20wdt%3AP361%20wd%3AQ113556931.%0A%20%20%3Fc%20wdt%3AP279%20%3Fcurriculum.%0A%20%20%3Fcurriculum%20wdt%3AP31%20wd%3AQ600134.%0A%20%20%3Farticle%20schema%3Aabout%20%3Ftopic%3B%0A%20%20%20%20schema%3AisPartOf%20%3Chttps%3A%2F%2Fen.wikipedia.org%2F%3E.%0A%20%20SERVICE%20wikibase%3Alabel%20%7B%20bd%3AserviceParam%20wikibase%3Alanguage%20%22%5BAUTO_LANGUAGE%5D%2Cen%22.%20%7D%0A%7D

image

It is saved properly and it looks like the collection of article is generated properly because no error:
image

But when I click on the "Download tsv", I get:
image

I would expect that I get the TSV with the same list of article like on Wikidata OR a proper error that something went wrong.

audiodude:
Fixes #555.

As explained in #555, that specific query was causing confusing results because the query variable was wrong. In general though, we need to be resilient against the possibility of there being no results in response from Wikidata. For this purpose, an empty results set is now considered a "fatal" error, since it couldn't be fixed by simply retrying and the query itself (or in this case the query variable) needs to be edited.

### DIFF
--- a/wp1/logic/builder.py
+++ b/wp1/logic/builder.py
@@ -174,6 +174,8 @@ def latest_selection_url(wp10db, builder_id, ext):
     return None
 
   if data['object_key'] is None:
+    logger.warning('Object key for selection was None, builder id=%s',
+                   builder_id)
     return None
 
   return logic_selection.url_for(data['object_key'].decode('utf-8'))
--- a/wp1/logic/builder.py
+++ b/wp1/logic/builder.py
@@ -191,11 +191,6 @@ def latest_selections_with_errors(wp10db, builder_id):
            WHERE b.b_id = %s AND s.s_status IS NOT NULL AND s.s_status != 'OK'
         ''', (builder_id,))
     data = cursor.fetchall()
-  if data is None:
-    logger.warning(
-        'Could not find latest selections with errors for builder id=%s',
-        builder_id)
-    return None
 
   res = []
   for db_selection in data:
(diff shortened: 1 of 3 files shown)


kelson42:
@audiodude Thank you for the fix. Maybe we will get an error explaining what is wrong... because I still don't understand it.

audiodude:
I put this error text in the error that will now be returned:

> The Wikidata response did not contain any URLs for subjects
> in your query identified with the variable "?article".
> Check that "?%(query_variable)s" refers to a Wikidata "topic"
> aka something with a QID. It may also be the case that none of
> the subjects your query selected have articles in the project
> you selected

It's a long winded explanation, I fully acknowledge that. I'm open to better ways of explaining it.

Issue #483, opened 2022-08-13
The Zimfarm form should be able to be submitted to the backend API of the Zimfarm installation to create a ZIM file


audiodude:

rgaudin:
This would be done easily using the [zimfarm API](https://api.farm.openzim.org/v1#/restricted/addRequestedTask).

I suggest following an approach similar to what the zimit-frontend (youzim.it) does: create a unique schedule, request a task for it, delete the schedule.

https://github.com/openzim/zimit-frontend/blob/main/api/src/routes/requests.py#L44

For this, you would need:
- an endpoint (#480)
- an API Token
- an agreed-upon resource profile (cpu, mem, disk) for those tasks (TBD)
- S3 bucket and access for uploads
- maybe mailgun credentials if you intend to use email

audiodude:
The API Token makes sense, I assume we would set one up and I would include it in every request.

How are the S3 bucket/mailgun credentials specified? Do they need to be set up in the youzimit project? Is that what you referred to in #480, that there's only one `ZIM_UPLOAD_URI`?

rgaudin:
Given the token expires after some time, it's best you store the username/password and re-request a token when needed.
There's an example of how to do it at https://github.com/openzim/zimit-frontend/blob/main/api/src/zimfarm.py
You may actually consider reusing that module altogether.

As for S3, you should not need it unless you plan on uploading additional files as ZIM files will be directly uploaded to it (yes, on the zimit bucket).

For mailgun, we'd need to create a sender and configure it on the DNS records then eventually create the token. Let me know if you'll need to send emails and if so, what would be the sending domain (`wp1.openzim.org`?)

Issue #482, opened 2022-08-13
The Zimfarm form must contain any additional metadata that is required for the ZIM file


audiodude:

rgaudin:
See https://github.com/openzim/zimfarm/blob/master/dispatcher/backend/src/common/schemas/offliners/mwoffliner.py

kelson42:
UI should be as simple as possible, we should really do only what is mandatory and offer a formular only if a metadata can not be automatically computed/given.

audiodude:
How do we know which of those fields is required?

rgaudin:
look for required=True

audiodude:
From discussion on slack:

> Meta data `Description` and `LongDescription` should be given, see https://wiki.openzim.org/wiki/Metadata to check which metadata should be given

kelson42:
@audiodude What is mandatory is explain on wiki.openzim.org, assuming everything is clear here. Otherwise please asking.

PR #583, opened 2023-03-04
Create zimfarm.py module for getting Zimfarm access tokens and requesting ZIM file creation tasks

audiodude:
This PR is part of #483. It provides the backend API for:

1. Connecting to a Zimfarm instance as configured in credentials.py, using user name and password and storing `access_token` and `refresh_token` in Redis
2. Requesting a ZIM file creation task from the Zimfarm based on a specific Builder, using it's latest selection that is in TSV format

Currently, this is "fire and forget". The library does not monitor the requested task or provide for any way, in WP1, to find out its status or download the resulting ZIM file.

It is also standalone and not called from anywhere in WP1.

Future work will first integrate this functionality with the web API at api.wp1.openzim.org, so that there is a way to request ZIM files based on WP1 Builders. As part of that, we will create a job tracking system so that we know the status of the ZIM file creation on the Zimfarm, with webhooks and some polling. We will also create a way for the user to download the "latest" ZIM file from the WP1 API. As part of this work, we will likely revisit the error handling for `schedule_zim_farm`, to possibly internally retry transient HTTP errors, as well as wrap fatal errors (400 BAD REQUEST) in a WP1 specific error class so it can be caught appropriately at the API layer.

The final step will be to build UI pages that expose this functionality.

### DIFF
--- a/wp1/logic/util_test.py
+++ b/wp1/logic/util_test.py
@@ -14,5 +14,4 @@ class TestUtil(unittest.TestCase):
   def test_safe_name(self):
     for test, expected in self.safe_name_tests:
       actual = util.safe_name(test)
-      print(test, expected, actual)
       self.assertEqual(expected, actual)
--- a/wp1/logic/util_test.py
+++ b/wp1/logic/util_test.py
@@ -1,0 +1,18 @@
+import unittest
+
+from wp1.logic import util
+
+
+class TestUtil(unittest.TestCase):
+  safe_name_tests = [
+      ('Foo Bar', 'FooBar'),
+      ('Français', 'Français'),
+      ('âme île & **hôtel-sûr_être', 'âmeîlehôtel-sûr_être'),
+      ('爱', '爱'),
+  ]
+
+  def test_safe_name(self):
+    for test, expected in self.safe_name_tests:
+      actual = util.safe_name(test)
+      print(test, expected, actual)
+      self.assertEqual(expected, actual)
--- a/wp1/zimfarm.py
+++ b/wp1/zimfarm.py
@@ -123,7 +123,7 @@ def _get_params(wp10db, builder_id):
           'memory': TASK_MEMORY,
           'disk': TASK_DISK,
       },
-      'platform': None,
+      'platform': 'wikimedia',
       'monitor': False,
       'flags': {
           'mwUrl': 'https://%s/' % project,
--- a/wp1/zimfarm.py
+++ b/wp1/zimfarm.py
@@ -1,0 +1,55 @@
+from wp1.credentials import CREDENTIALS, ENV
+from wp1.logic import util
+import wp1.logic.builder as logic_builder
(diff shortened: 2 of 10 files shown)


wp1/logic/util_test.py
```diff
@@ -0,0 +1,18 @@
...
+      ('Foo Bar', 'FooBar'),
+      ('Français', 'Français'),
+      ('âme île & **hôtel-sûr_être', 'âmeîlehôtel-sûr_être'),
+      ('爱', '爱'),
+  ]
+
+  def test_safe_name(self):
+    for test, expected in self.safe_name_tests:
+      actual = util.safe_name(test)
+      print(test, expected, actual)
```
rgaudin, line 17: Debug print?

wp1/zimfarm.py
```diff
@@ -0,0 +1,194 @@
...
+      'image': {
+          'name': MWOFFLINER_IMAGE.split(':')[0],
+          'tag': MWOFFLINER_IMAGE.split(':')[1],
+      },
+      'resources': {
+          'cpu': TASK_CPU,
+          'memory': TASK_MEMORY,
+          'disk': TASK_DISK,
+      },
+      'platform': None,
```
rgaudin, line 126: platform should be `wikimedia` as those tasks would target Wikimedia servers and there are (in-zimfarm) per-worker throttling involved for wikimedia-platform related tasks.

wp1/zimfarm.py
```diff
@@ -0,0 +1,194 @@
...
+      'monitor': False,
+      'flags': {
+          'mwUrl': 'https://%s/' % project,
+          'adminEmail': 'contact+wp1@kiwix.org',
+          'articleList': logic_selection.url_for_selection(selection),
+          'customZimTitle': util.safe_name(builder.b_name.decode('utf-8')),
```
rgaudin, line 132: I believe @kelson42 requested to set the description and longdescription…

rgaudin (approved):
Great ; I've added a couple inline comments. `platform` needs to be set to `wikimedia`.

wp1/logic/util_test.py
audiodude: Good catch, fixed.

wp1/zimfarm.py
audiodude: Done

wp1/zimfarm.py
audiodude: Still waiting to hear what he wants to set them to. Tracked in #482

kelson42: @audiodude not clear what is expected from me there.

kelson42:
@audiodude ZIM Medata should be improved:
* `Description` is mandatory
* Ability to set `LongDescription` would be great
* In both cases please check max size against spec.

I have rebased and now everything looks good in the CI.

PR #591, opened 2023-03-21
Upgrade flask to 2.2.3 to get werkzeug updates. Fork some libraries.

audiodude:
Use forked versions of rq-dashboard and flask-session for compatibility.

Fixes a dependabot warning.


audiodude:
Looks like there's some problems with rq-dashboard and the test environment.

kelson42:
@rgaudin I rely on you for this, but not in favour of:
* Keeping unmaintained dependences without clear replacement agenda
* Relying on forks

I wonder why @audiodude faces these problems and we don't, considering he has adopted a stack we recommended AFAIK?

@audiodude Possible to have a clear assessment about the project deps (with updaye/renewal plan)?

rgaudin:
> @rgaudin I rely on you for this, but not in favour of:
> 
>     * Keeping unmaintained dependences without clear replacement agenda 
>     * Relying on forks


> I wonder why @audiodude faces these problems and we don't, considering he has adopted a stack we recommended AFAIK?

Obviously we only recommend Flask as a framework but a python project, especially a complex web one like wp1 relies on a lot of other dependencies.

Here @audiodude upgrades to current latest Flask which is the proper thing to do.
One dependency that he uses – we don't (Flask-Session) uses a now deprecated thing. Flask is less than a month old and that lib is 2 years old.

He created [a PR upstream](https://github.com/fengsp/flask-session/pull/153) but there are a lot of pending PRs there so he created a fork. That's the way to go.

Now what could be done is to create a ticket here with a reminder in a few months to reassess: been merged ? not ? Find a replacement and refactor ?

As for rq-dashboard, we don't use it. AFAIK it's not required for wp1 and is more of a monitor tool. There's a [PR upstream](https://github.com/Parallels/rq-dashboard/pull/413) but it's a similar situation as for flask-session

audiodude:
@kelson42 @rgaudin Should we merge this for now to patch the security vulnerability? Then decide what to do with the forked libraries later, as outlined in the related issues?

kelson42 (approved):
@audiodude LGTM
@rgaudin I still don't fully grasp the particular situation of WP1 regarding sessions, but I expect us to have a clear recommend techno way to solve that kind of problem. Current situation is problematic.

rgaudin:
I've already approved the PR. We have a ticket about those forks so it's good to go from my POV.

audiodude:
Okay, I've responded on the appropriate tickets, but I think we should use `Flask-Session2` and drop `rq-dashboard`. I'll prepare PRs that do each of those, then another that upgrades Flask. Abandoning this one for now.

audiodude:
Looks like Flask-Session2 requires Flask 2.2.3 so I'll have to upgrade them at the same time. And once I upgrade Flask I'll have to remove rq-dashboard, so it's all coming in the same PR.

audiodude:
See #601

PR #601, opened 2023-04-06
Upgrade to Flask 2.2.3, switch to Flask-Session2, and drop rq-dashboard

### LINKED ISSUE #592: Decide what should be done with rq-dashboard
We have a dependency on the `rq-dashboard` project, which gives us real time insights into how our many redis queues and workers are doing.

Pros
- It provided very helpful insight when the project was more unstable into what was going on with cron failures
- It's lightweight and operates outside of the normal web stack, so can be useful for monitoring

Cons
- It hasn't been used much in the past year or so
- Only I (@audiodude) know how to use it
- The library that implements it is out of date with the latest Flask version (2.2.3) and completely obsolete with the upcoming Flask version (2.3 -- warnings will be elevated to exceptions)
- As such, the only way to keep it working is to maintain our own fork

Pros
- That fork is already done and will likely take very little effort to maintain
- We will get dependabot alerts in the fork if there are any security vulnerabilities which require us to revisit our support

### LINKED ISSUE #593: Decide what should be done with Flask-Session
Flask, by default, stores session data wholly in a client side cookie. From what I understand, this cookie is encrypted so is resistant against client side tampering.

For WP1, we use the library `Flask-Session` (https://flask-session.readthedocs.io/en/latest/), which allows us to store session data in Redis. The user still has a secure cookie which provides their session key, which I believe is just equivalent to a Redis key for the data.

Cons:
- The Flask-Session library requires forking/patching to be compatible with Flask 2.2.3 and above
- We probably can get by without the functionality it provides, based on our limited use of session data

Pros
- It's already implemented, tested, and working. Refactoring the code to work without it is risky. AFAIK there is no integration or end2end test which exercises sessions
- The fork is already done and should be easy to maintain (we will get dependabot alerts if there are any security vulnerabilities that need to be fixed).

audiodude:
Fixes #592. Fixes #593.


rgaudin (approved):
Great ; thanks 👍

PR #609, opened 2023-04-22
Deal with ZIM file errors on the Zimfarm

### LINKED ISSUE #604: Check for ZIM failures when 'ended' webhook is called
The `ended` webhook appears to be called even when the request had an error or failure. Let's check the POSTed JSON for failure messages and update the status of the selection ZIM request accordingly.

Otherwise, WP1 will be in a perpetual "waiting" state.

audiodude:
Fixes #604.

Check for `status: failed` in the JSON returned by the Zimfarm, both when the webhook is called and when polling for file ready status. If a failure state is detected in either case, mark the Selection Zim status as `FAILED`.

When this status is encountered in the frontend, display a failure message with an option to retry. Also display failed status in the My Lists table with an option to go to the Zim file page and retry.

### DIFF
--- a/wp1-frontend/cypress/e2e/myLists.cy.js
+++ b/wp1-frontend/cypress/e2e/myLists.cy.js
@@ -11,7 +11,7 @@ describe('the user selection list page', () => {
     it('successfully loads', () => {});
 
     it('displays the datatables view', () => {
-      cy.get('.dataTables_info').contains('Showing 1 to 9 of 9 entries');
+      cy.get('.dataTables_info').contains('Showing 1 to 10 of 10 entries');
     });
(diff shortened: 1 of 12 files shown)


rgaudin (approved):
LGTM 👍 

Just wanted to state that this assumes a zimfarm task is `failed` if a ZIM is failed.
It's probably a safe assumption in wp1 context but zimfarm's slightly more flexible.

- there can be multiple ZIM files per task
- a `failed` task can have successful file(s)
- a `failed` task can have no erroneous files
- currently we could even have a successful task without a ZIM but we have a ticket to fix that as this would indicate something went wrong

PR #611, opened 2023-04-23
Fail ZIM file requests that have been polling for more than 2 hours

### LINKED ISSUE #608: Stop Zimfarm polling after some timeout
As written, ZIM file polling for the completed file set will continue indefinitely. This could be a problem if there's some unforeseen error in the process and the files will never be ready. We should quit after some set time, maybe 2 hours?

audiodude:
Fixes #608

### DIFF
--- a/wp1/logic/builder.py
+++ b/wp1/logic/builder.py
@@ -1,9 +1,10 @@
 import json
 import logging
+import time
 
 import attr
 
-from wp1.constants import CONTENT_TYPE_TO_EXT, EXT_TO_CONTENT_TYPE, TS_FORMAT_WP10
+from wp1.constants import CONTENT_TYPE_TO_EXT, EXT_TO_CONTENT_TYPE, MAX_ZIM_FILE_POLL_TIME, TS_FORMAT_WP10
 from wp1.credentials import CREDENTIALS, ENV
 from wp1.exceptions import ObjectNotFoundError, UserNotAuthorizedError
 import wp1.logic.selection as logic_selection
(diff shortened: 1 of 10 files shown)


wp1/selection/models/sparql.py
```diff
@@ -116,6 +116,7 @@ def build(self, content_type, **params):
       raise Wp1FatalSelectionError('Wikidata response was larger than 10 MB')
 
     articles = self._extract_articles(project, params_query, data)
+    print(articles)
```
rgaudin, line 119: debug?

wp1/logic/builder.py
```diff
@@ -331,6 +333,15 @@ def on_zim_file_status_poll(task_id):
                                         'FILE_READY',
                                         set_updated_now=True)
   elif result == 'REQUESTED':
+    requested = logic_selection.zim_file_requested_at_for(wp10db, task_id)
+    if requested is not None:
+      now = utcnow().timestamp()
+      print(now, requested, now - requested)
```
rgaudin, line 339: debug?

rgaudin (commented):
From what I understand, when wp1 requests a new ZIM to the Zimfarm, it stores the time and starts polling for updates.
Once it reaches 2h, it stops polling and marks the requests (on wp1 side) as failed.

If that's what's happening I don't think it's correct.

While we (I) should make sure the infrastructure is ready to accept tasks as quickly as possible, it's **not uncommon** that a requested tasks sits in the pipe for a while ; including for several hours. We (I) should monitor this as we don't want this frequently, but it can't be considered an error at this stage.

Secondly, running mwoffliner over any SPARQL query result can take over two hours to process I suppose. Unless there is some kind of limit on the number of articles in wp1, I can imagine some queries returning tens of thousands of articles…
Should not be considered an error either.

Another approach could be to reduce the polling frequency incrementally up to a larger threshold.

@audiodude did I get that right?

audiodude:
Thanks for the review. I think it's a bit more nuanced though.

The polling doesn't start until the webhook has been called in regards to the task being marked as `ended`. So conceptually, all we're polling for is the files to be ready (uploaded to s3). In my mind, this shouldn't take more than N hours (even if 2 is too conservative).

What I've seen happen is that the task "succeeds", and is "ended", with mwoffliner having scraped all the articles and possibly generated the ZIM, but then it fails inexplicably (in one case it was because the name of the ZIM file hadn't changed so the system couldn't upload the file). The polling is necessary in my mind because we can't provide the file link before the file is uploaded. But we can't poll indefinitely either.

So to be crystal clear, the tens of thousands of articles being scraped and processed all happens *before* the task is marked as `ended`, correct? If that's the case then I think this PR makes sense and we might just need to tweak the value of N hours.

rgaudin (approved):
Ah OK ; it's all good then. I believe 2h is a fine duration in this case. Stackoverflow took about 2h to upload and that's one of the largest. Please remove the two prints before merging

Issue #581, opened 2023-02-27
What to do with outdated ZIM files?


audiodude:
If a user updates their selection, how should we handle the (now) outdated ZIM file? A couple of options:

1. Remove it from the interface and force the user to regenerate it if they want a ZIM file.
2. Leave the old ZIM file, with an "updated date" displayed and a notice that it's out of date, but allow the user to download the old file and manually request a new one.
3. Same as 2, but automatically re-request and replace the old file in the background.
4. Provide an interface to download any ZIM file that was ever created, with some combination of 2) and 3) for generating a new file.

I think 2) is probably the simplest and most robust for a v1. But 4) isn't completely out of scope, since we're already versioning and saving every generated selection.

kelson42:
I propose:
* Automatically schedule the build of a new ZIM file (cancel one if necessary)
* Keep providing the "old" zim file as long as nothing more recent is available
* When new filw ready, replace the old link by the new link

@

rgaudin:
> @rgaudin how long zim files are kept in the S3?

2 weeks in the current `zimit` bucket. We could (should we?) move to a dedicated bucket with a different config.

> Do we have a financial benefit to delete old zim files from S3?

Of course! This storage ain't free ! 😁 We pay **monthly** for the **amount of data** stored. Even if we delete a file after 1mn, we'd still pay for the first 30days so there's no incentive to delete before it's 30d old.

kelson42:
> > @rgaudin how long zim files are kept in the S3?
> 
> 2 weeks in the current `zimit` bucket. We could (should we?) move to a dedicated bucket with a different config.


I think we are good, but that means after two weeks, wp1 should not provide anymore a dead link. That probably implies that the user should quickly have a way to trigger the generation of a new version of the ZIM.

audiodude:
@rgaudin Do the Zimfarm tasks expire after 2 weeks as well?

I'm asking because I think it would be better to re-request a ZIM when we try to download it but the task is expired, rather than having a cron job that re-generates all the ZIMs every two weeks. Imagine 1000 users generate ZIM files, they're happy with them, they forget they ever used the WP1 tool. But then we're still re-generating 1000 ZIMs every two weeks forever and using resources.

rgaudin:
> @rgaudin Do the Zimfarm tasks expire after 2 weeks as well?

No. Nothing happens on the Zimfarm. It's just that the S3 bucket is configured with 2w auto-delete.

[...]

PR #614, opened 2023-05-13
Re-factor ZIM files to their own table

### LINKED ISSUE #612: Refactor ZIM files into their own table
Based on the requirements in #581 and #610, we should refactor the ZIM file data into its own table, rather than being columns on the Selection table. This is because when the Builder is updated and a new Selection is generated, we want to maintain the link to the old ZIM file task/URL while we re-generate a new one. We also need a place to store `description` and `long_description` for the ZIM file, which we need to re-provide on error cases and when generating a new ZIM. It seems awkward to store that data in the Selection table.

audiodude:
Fixes #612.

This is mostly a refactor in preparation for #581. We need to have a separate ZIM file table, since the old file needs to be retained while the new one is auto-generated. We also need to keep track of the `description` and `long_description` of the ZIM file so that we can automatically re-request it without re-prompting the user.

This PR adds no new features and the migrations are designed so that everything already in place keeps working.


rgaudin (approved):
LGTM 👍

audiodude:
Force push for rebase to main

PR #617, opened 2023-05-24
Fix swapped values and add test

audiodude:
In the previous PR, the values for `description` and `long_description` were swapped. This fixes it and adds a test.


audiodude:
I ran the following SQL in the production database:

```
SELECT * FROM zim_files WHERE z_description IS NOT NULL;

UPDATE zim_files SET z_description=(@temp:=z_description), z_description = z_long_description, z_long_description = @temp; 
```

Luckily, no one had used a `long_description` longer than 255 characters, so no data was truncated and this worked.

PR #625, opened 2023-06-19
Cancel outdated ZIM tasks when a new ZIM is auto-updated

### LINKED ISSUE #622: Fix race condition when multiple ZIMs requested
Currently, when a Selection with a ZIM file is updated, we automatically request a new ZIM file and update the associated download link once the new file is ready. However, if a user does many updates to their Selection in succession, while the auto requested ZIM is pending, there is a race condition where existing requests aren't cancelled and the last ZIM to finish is considered the most up to date file.

We should keep track of pending Zimfarm requests for a zim_file and cancel them when a new request is made.

audiodude:
Fixes #622.

Here, we check for any ZIM file objects in the database for the given builder that are in the REQUESTED state, when auto-queueing a new ZIM file. That way, we can cancel these pending tasks before queueing the new ZIM file for creation. This prevents the race condition described in the bug.

### DIFF
--- a/wp1/zimfarm.py
+++ b/wp1/zimfarm.py
@@ -304,3 +304,33 @@ def zim_file_url_for_task_id(task_id):
         'Configuration error, could not find ZIMFARM["s3_url"] in credentials')
 
   return f'{base_url}{warehouse_path}/{name}'
+
+
+def cancel_zim_by_task_id(redis, task_id):
+  token = get_zimfarm_token(redis)
+  if token is None:
+    raise ZimfarmError('Error retrieving auth token for request')
+  base_url = get_zimfarm_url()
+  headers = _get_zimfarm_headers(token)
+
+  logger.info('Deleting task_id=%s', task_id)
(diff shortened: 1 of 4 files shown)


wp1/zimfarm.py
```diff
@@ -304,3 +304,33 @@ def zim_file_url_for_task_id(task_id):
...
+  base_url = get_zimfarm_url()
+  headers = _get_zimfarm_headers(token)
+
+  logger.info('Deleting task_id=%s', task_id)
+  r = requests.delete('%s/requested-tasks/%s' % (base_url, task_id),
+                      headers=headers)
+
+  try:
+    r.raise_for_status()
+    # Didn't raise, operation was successful.
```
rgaudin, line 322: Using `try`/`except`/`else` makes it explicit and removes the need for the comment (IMO)

rgaudin (approved):
LGTM 👍

wp1/zimfarm.py
audiodude: Good point, done.

PR #627, opened 2023-06-23
Zim title

### LINKED ISSUE #626: Use full Selection name as ZIM title
We are currently sending the `--customZimTitle` field to the Zimfarm. However, we are putting our desired file name slug in that field, instead of a human readable title. We should use the actual name of the Selection, spaces and all, as the ZIM title.

audiodude:
Fixes #626.

Minor change to update `customZimTitle` to be the full name of the Builder, and use `filenamePrefix` to create a unique slug for the ZIM file.

During the testing of this change, a couple of bugs with the previous PRs were found:

1. The `task_id` being sent to the Zimfarm to cancel was bytes, not str, so it created an encoding error in the URL
1. ZIM files weren't being marked as CANCELLED once they were cancelled, so if another new ZIM was requested, the old ZIMs would be (harmlessly) requested to be cancelled over and over.


rgaudin (approved):
LGTM 👍

PR #633, opened 2023-06-26
Fix bug with deleted ZIMs

audiodude:


audiodude:
Abandoning, breaks e2e tests. See #634.

PR #642, opened 2023-07-13
Update ZIM page to allow re-requesting ZIM files that have expired

### LINKED ISSUE #637: Impossible to recreate updated ZIM file
Here is my selections:
image

I click on "Create ZIM" for the "Apple" selection:
image

At this stage I'm not really sure what is going on. I have clicked on a button called "Create ZIM", then here it's writen "once you request a ZIM file" and there is a button "Download ZIM" ... Whatever, I click on "Download ZIM"

image

I would expect to get either the ZIM file or a clear indication that the creation of a new ZIM has been launched. The message should be also fixed so we use always the same verbs and this is clear what is left to do and/or ongoing.

### LINKED ISSUE #639: Unable to create a ZIM
Here my selections:
image

I want to get the ZIM for "Bardo Test Selection", so I click on "Create ZIM"
image

The button "Download ZIM" is greyed and nothing happen when I click on it.

I expect to download or have a way to create the ZIM.

audiodude:
Fixes #637

This change passes the `is_deleted` flag to the ZIM creation page. If it sees that value, it notifies the user that their ZIM has expired and allows them to re-request.

### DIFF
--- a/wp1-frontend/cypress/e2e/zimFile.cy.js
+++ b/wp1-frontend/cypress/e2e/zimFile.cy.js
@@ -67,7 +67,8 @@ describe('the zim file creation page', () => {
               statusCode: 204,
             }).as('request');
 
-            cy.get('#desc').click().type('Description from user');
+            cy.get('#desc').click();
+            cy.get('#desc').type('Description from user');
             cy.get('#request').click();
           });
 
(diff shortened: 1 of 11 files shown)


rgaudin (approved):
LGTM 👍

PR #643, opened 2023-07-16
Allow for Selection based on Petscan URL

audiodude:
[Petscan](https://petscan.wmflabs.org/) allows for powerful methods for selecting articles based on union or intersection of categories, among other features. Particularly notable is the fact that for a category such as `Living_people`, which has over a million results, Petscan is able to generate the article list. It is dubious if our SPARQL Builder would be able to create such a list.

This PR implements a Petscan Builder, as well as a frontend that allows for Petscan selection based on URL.

### DIFF
--- a/wp1-frontend/src/components/PetscanBuilder.vue
+++ b/wp1-frontend/src/components/PetscanBuilder.vue
@@ -1,0 +1,80 @@
+<template>
+  <BaseBuilder
+    :key="$route.path"
+    :listName="'Petscan Selection'"
+    :model="'wp1.selection.models.petscan'"
+    :params="params"
+    :builderId="$route.params.builder_id"
+    :invalidItems="invalidItems"
+    @onBuilderLoaded="onBuilderLoaded"
+    @onBeforeSubmit="onBeforeSubmit"
+    @onValidationError="onValidationError"
+  >
+    <template #create-desc>
+      <p>
+        Use this tool to create an article selection list for the Wikipedia
+        project of your choice, based off a Petscan URL. Your selection will be
+        saved in public cloud storage and can be accessed through URLs that will
+        be provided once it has been saved.
+      </p>
+      <p class="mb-0">
+        For more information on creating a Petscan selection, see the
+        <a href="https://wp1.readthedocs.io/en/latest/user/selections/"
+          >end user documentation</a
+        >
+        or the
+        <a href="https://meta.wikimedia.org/wiki/PetScan/en"
+          >Petscan User Manual</a
+        >.
+      </p>
+    </template>
+    <template #extra-params>
+      <div id="items" class="form-group m-4">
+        <label for="items">URL</label>
+        <input
+          id="petscanUrl"
+          ref="petscanUrl"
+          class="form-control my-2"
+          v-model="params.url"
+        />
+        <div class="invalid-feedback">Please provide a valid URL</div>
+      </div>
+    </template>
+  </BaseBuilder>
+</template>
+
+<script>
+import BaseBuilder from './BaseBuilder.vue';
+
+export default {
+  components: { BaseBuilder },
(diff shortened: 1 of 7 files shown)


wp1-frontend/src/components/PetscanBuilder.vue
```diff
@@ -0,0 +1,80 @@
...
+    </template>
+  </BaseBuilder>
+</template>
+
+<script>
+import BaseBuilder from './BaseBuilder.vue';
+
+export default {
+  components: { BaseBuilder },
+  name: 'SimpleBuilder',
```
rgaudin, line 51: Why not `PetscanBuilder`?

wp1/selection/models/petscan.py
```diff
@@ -0,0 +1,54 @@
...
+
+class Builder(AbstractBuilder):
+
+  def build(self, content_type, **params):
+    if content_type != 'text/tab-separated-values':
+      raise Wp1FatalSelectionError('Unrecognized content type')
+    if 'url' not in params:
+      raise Wp1FatalSelectionError('Missing required param: url')
+    if not isinstance(params['url'], str):
+      raise Wp1FatalSelectionError('Param `url` was not str')
```
rgaudin, line 22:
```suggestion
      raise Wp1FatalSelectionError('Param `url` is not str')

```
Looks like you don't use past tense for others

rgaudin (approved):
Probably not a concern for now but there might be some scaling issue.

I notice you haven't set any timeout on the petscan URL fetching while it seems that it runs the query on every call and this can take some time for those such as `Living_people`. Relying on petscan's 502 might be fragile if they increase their timeout over your own.

Also, you're loading all the response JSON into memory which for `Living_people` is 172MB.
Not worth handling now IMO but you might want to open a ticket…

wp1-frontend/src/components/PetscanBuilder.vue
audiodude: It's a copy/paste mistake. Though apparently the version of Vue I'm using ignores this field I think? It doesn't cause any errors or bugs in the frontend. Fixed.

audiodude:
> Probably not a concern for now but there might be some scaling issue.
> 
> I notice you haven't set any timeout on the petscan URL fetching while it seems that it runs the query on every call and this can take some time for those such as `Living_people`. Relying on petscan's 502 might be fragile if they increase their timeout over your own.

I think what you're saying is that I want to set a longer timeout than the `requests` default, to allow for Petscan processing time, is that right? Good idea.

> Also, you're loading all the response JSON into memory which for `Living_people` is 172MB. Not worth handling now IMO but you might want to open a ticket…

I'm not sure how else I would parse the JSON. I know that for certain datasets, they are presented as JSONL/"JSON lines" where you can parse smaller objects one at a time.

Actually, now that I'm looking at Petscan, it seems they have a "Plain text" option which returns just article names, which is what I want to store in S3 anyways. So I can probably stream that response directly into S3 somehow.


Thanks for the review!

audiodude:
Actually according to SO, there is no default `requests` timeout (https://stackoverflow.com/a/17782541/41060).

So I think this is behaving properly? Unless you think I should abandon some Petscan queries?

audiodude:
Filed #648

rgaudin:
> I think what you're saying is that I want to set a longer timeout than the `requests` default, to allow for Petscan processing time, is that right? Good idea.

No, requests by default has no timeout. What I mean is that your service will be served by a reverse proxy which has a timeout (60s by default in nginx). This timeout is likely to be shorter than Petscan's one. Without controlling a timeout on your request, you may start a long lasting request that in the mean time gets shut down by your reverse proxy resulting in a confusing 502 for your user.
I'm suggesting setting a timeout on your requests call. But as I said, maybe overkill for now

Issue #640, opened 2023-07-09
Allow to configure ZIM title


kelson42:
Currently it is impossible, it always take (I guess) the selection name
![image](https://github.com/openzim/wp1/assets/1029718/c27139b7-8e2f-4cd1-b2d7-0fafed8d426d)

We should have the ZIM title configurable with per default the selection name set.

audiodude:
Will add this as part of #677

PR #645, opened 2023-07-16
Allow users to create selections from Wikipedia books

### LINKED ISSUE #644: Save Wikipedia books (created with Book creator) in ZIM
New selection type: wikipedia books created with https://en.wikipedia.org/w/index.php?title=Special:Book&bookcmd=book_creator.

First with could allow to just give a Wikipedia Book URL.

Once this is working we could provide a direct way, on Wikipedia, to export in ZIM format.

audiodude:
Fixes #644.

Users can create selections based on a Wikipedia "book" url, such as https://en.wikipedia.org/wiki/User:Audiodude/Books/test. The Book Builder fetches the wikitext of the page that is saved by this tool and parses out the article names, then saves them in S3. A frontend is provided for creation of Book selections, which verifies that the URL is present, valid, and matches the project selected in the dropdown.

Book selections can be used to create ZIM files just like any other selection.

### DIFF
--- a/wp1/selection/models/book.py
+++ b/wp1/selection/models/book.py
@@ -1,0 +1,66 @@
+import logging
+
+import mwparserfromhell
+import requests
+import validators
+
+from wp1.constants import WP1_USER_AGENT
+from wp1.exceptions import Wp1FatalSelectionError
+from wp1.selection.abstract_builder import AbstractBuilder
+
+logger = logging.getLogger(__name__)
+
+
+class Builder(AbstractBuilder):
+
+  def build(self, content_type, **params):
+    if content_type != 'text/tab-separated-values':
+      raise Wp1FatalSelectionError('Unrecognized content type')
+    if 'url' not in params:
+      raise Wp1FatalSelectionError('Missing required param: url')
+    if 'project' not in params:
+      raise Wp1FatalSelectionError('Missing required param: project')
+
+    if not isinstance(params['url'], str):
+      raise Wp1FatalSelectionError('Param `url` was not str')
+    if not isinstance(params['project'], str):
+      raise Wp1FatalSelectionError('Param `project` was not str')
+
+    book_name = params['url'].split('wiki/')[1]
+    final_url = (
+        'https://%s/w/api.php?'
+        'action=query&prop=revisions&rvprop=content&format=json&rvslots=main'
+        '&titles=%s' % (params['project'], book_name))
+
+    resp = requests.get(final_url, headers={'User-Agent': WP1_USER_AGENT})
+    try:
+      resp.raise_for_status()
+    except requests.exceptions.HTTPError as e:
+      logger.exception('Error status received from Wikipedia API')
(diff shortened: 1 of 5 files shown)


kelson42:
@audiodude Does it grab title and subtitle?

audiodude:
> @audiodude Does it grab title and subtitle?

It does not. I didn't see that requirement in the bug. What would you expect to use this data for? Selections currently only store their article lists (in S3) and don't have fields for additional metadata. I assume you want to use title and subtitle for metadata in the ZIM?

kelson42:
@audiodude yes yes yes

audiodude:
Okay I think we should create a separate bug to track that work, since it is substantially more complicated than what's in this PR and would bloat this one (which is, of course, already finished).

audiodude:
I'm thinking of something following from #640 as well, where we can surface more of the ZIM metadata on the ZIM page, under an optional expanding section, with some of it pre-populated based on the result of the selection materialization.

wp1/selection/models/book.py
```diff
@@ -0,0 +1,78 @@
...
+      raise Wp1FatalSelectionError('Missing required param: url')
+    if 'project' not in params:
+      raise Wp1FatalSelectionError('Missing required param: project')
+
+    if not isinstance(params['url'], str):
+      raise Wp1FatalSelectionError('Param `url` was not str')
+    if not isinstance(params['project'], str):
+      raise Wp1FatalSelectionError('Param `project` was not str')
+
+    book_name = params['url'].split('wiki/')[1]
```
rgaudin, line 30: This might raise `IndexError` on incorrect URL. Maybe check it's present in `validate`?

wp1/selection/models/book.py
```diff
@@ -0,0 +1,78 @@
...
+    pages = data['query']['pages']
+    page = list(pages.values())[0]
+    wikitext = page['revisions'][0]['slots']['main']['*']
+
+    parsed = mwparserfromhell.parse(wikitext)
+    unique = set()
+    titles = []
+    for link in parsed.filter_wikilinks():
+      title = link.strip('[]').replace(' ', '_')
+      if title not in unique:
```
rgaudin, line 54: I don't understand the need for the two sequences. Why not just using a `set`, directly adding to it? str.join supports it

rgaudin (approved):

wp1/selection/models/book.py
audiodude: Good point, thanks! Done.

wp1/selection/models/book.py
audiodude: I originally had it just joining the set.

Then I realized that to be courteous, and in case it matters, we don't want to alter the ordering of the articles in the book, since that is the specific order that the user put them in. As you know, set() has an indeterminate enumeration order.

rgaudin: You're right ; I didn't realize order mattered.

Issue #664, opened 2023-09-28
Upgrade to latest Python 3 version


audiodude:
3.11.5 at the time of this writing.

Should just be a matter of upgrading Python, running tests, and updating the Docker base images.

audiodude:
Swinging for the fences, going for 3.12.0

audiodude:
Looks like `nose` cannot be updated to 3.12.0 because the `imp` package has been removed: https://docs.python.org/3.12/whatsnew/3.12.html#imp

Since `nose`/`nosetests` is in maintenance mode anyways, it is unlikely this will ever be addressed.

More reason to prioritize #588

audiodude:
The 3.12 update was not applied evenly and needs to be done again.

PR #667, opened 2023-09-28
Remove mkdocs from main requirements.txt

audiodude:
This was included in the main `requirements.txt`, probably as mkdocs was being developed, but is not needed there.


audiodude:
#664

PR #673, opened 2023-10-04
Try ignoring paths in codecov

### LINKED ISSUE #668: Update codecov to not count *_test.py files
I think we're "cheating" a bit on our code coverage metrics, because codecov is counting test files as part of the code. Obviously a test file will always have 100% coverage, but that's not really what the definition of code coverage implies.

audiodude:
Fixes #668


audiodude:
Yay we only lost 4.29% test coverage!

PR #714, opened 2024-03-18
rq 1.13.0->1.12.0

### LINKED ISSUE #713: pytest 'resource' module not found
image

logan-keede:
rq 1.13.0 gives "resource" module not found on running pytest, due to rq instability https://github.com/rq/rq/issues/948
Should resolve #713 

replace rq==1.13.0 with rq==1.12.0 passes import checks

Note:-this error was found in windows, just in case if that has any effect


audiodude:
As mentioned in #713, this is a Windows specific error. I don't feel comfortable downgrading `rq` for this reason, especially since it implies that we would be stuck on that version forever in order to support Windows.

PR #729, opened 2024-05-02
Add WikiProject Builder model, only in Python not exposed to frontend yet

audiodude:
In WP1, we have the concept of a Builder, which is a sort of module that knows how to "build" selections from various data sources and parameters. For example, the Simple builder takes a `list=['Article 1', 'Article 2', ...]` parameter and does some validation before echoing back the list as a selection. The SPARQL Builder sends a SPARQL query to the Wikidata API and formats the response into a selection.

Note that in the codebase, "builder" is also the name given to a user's specific instance of a Builder, with the parameters instantiated and saved in the database.

In this PR, we introduce the WikiProject Builder, to create a selection based on the articles that are part of one or more WikiProjects. WP1 already has a mapping of WikiProjects -> articles, based on the work of the WP1.0 Bot. The WikiProject Builder takes two parameters:

`add` - A list of WikiProjects whose articles should be added, calculated as the articles that are a unique set of articles from the union of these projects

`subtract` - A list of WikiPorjects whose articles should be excluded, again calculated as the articles that are a unique set of articles from the union of the projects.

The end result is the set difference between the articles generated by `add` and those generated by `subtract`. So the full formula is:

```
For A, B in ADD; C, D in SUBTRACT:
(A ∪ B) - (C ∪ D)
```

Additionally, the validation step confirms that every WikiProject specified in either parameter actually exists in the database.

Finally, we add a new parameter that AbstractBuilder sends to it's subclasses, `wp10db=`, which is the database connection to the application database. This is possible because the `build` and `validate` methods have a contract where they accept a `**params` list of parameters, so additional keyword arguments can be added freely and are ignored by Builder implementations that are not expecting them.

As part of this system, the `b_params` field of the `builders` table in the database is a JSON encoded string that represents the `params` dictionary. So in the case of the SimpleBuilder mentioned above:

```
b_params == '{"list": ["Article 1", Article 2"]}'
```

This system makes it easy to store arbitrary parameters that a Builder might need, and pass them to the Builder implementation at runtime without any knowledge of a "schema".

### DIFF
--- a/wp1/selection/abstract_builder.py
+++ b/wp1/selection/abstract_builder.py
@@ -50,7 +50,6 @@ class AbstractBuilder:
 
     selection.set_updated_at_now()
     # Data might be None if build operation didn't succeed.
-    print(selection.data)
     if selection.data:
       self._upload_to_storage(s3, selection, builder)
 
--- a/wp1/selection/abstract_builder.py
+++ b/wp1/selection/abstract_builder.py
@@ -37,16 +37,20 @@ class AbstractBuilder:
     try:
       selection.data = self.build(content_type,
                                   project=builder.b_project.decode('utf-8'),
+                                  wp10db=wp10db,
                                   **params)
     except Wp1RetryableSelectionError as e:
+      logger.exception('Error materializing builder id=%s', builder.b_id)
       selection.s_status = 'CAN_RETRY'
       logic_selection.set_error_messages(selection, e)
     except (Wp1FatalSelectionError, Wp1SelectionError) as e:
+      logger.exception('Error materializing builder id=%s', builder.b_id)
       selection.s_status = 'FAILED'
       logic_selection.set_error_messages(selection, e)
 
     selection.set_updated_at_now()
     # Data might be None if build operation didn't succeed.
+    print(selection.data)
     if selection.data:
       self._upload_to_storage(s3, selection, builder)
 
--- a/wp1/selection/models/wikiproject.py
+++ b/wp1/selection/models/wikiproject.py
@@ -11,7 +11,7 @@ def _get_articles_for_project(wp10db, project):
 
 def _project_exists(wp10db, project):
   with wp10db.cursor() as cursor:
-    cursor.execute('SELECT 1 FROM ratings WHERE r_project = %s',
+    cursor.execute('SELECT 1 FROM projects WHERE p_project = %s',
                    (project.encode('utf-8'),))
     return cursor.fetchone() is not None
 
--- a/wp1/selection/models/wikiproject.py
+++ b/wp1/selection/models/wikiproject.py
@@ -23,28 +23,31 @@ class Builder(AbstractBuilder):
       raise Wp1FatalSelectionError('Unrecognized content type')
     if 'wp10db' not in params:
       raise Wp1FatalSelectionError('Missing param `wp10db`')
-    if 'add' not in params:
-      raise Wp1FatalSelectionError('Missing param `add`')
+    if 'include' not in params:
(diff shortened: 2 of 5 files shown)


wp1/selection/abstract_builder.py
```diff
@@ -37,16 +37,20 @@ def materialize(self, s3, wp10db, builder, content_type, version):
...
       selection.s_status = 'CAN_RETRY'
       logic_selection.set_error_messages(selection, e)
     except (Wp1FatalSelectionError, Wp1SelectionError) as e:
+      logger.exception('Error materializing builder id=%s', builder.b_id)
       selection.s_status = 'FAILED'
       logic_selection.set_error_messages(selection, e)
 
     selection.set_updated_at_now()
     # Data might be None if build operation didn't succeed.
+    print(selection.data)
```
benoit74, line 53: Debug left-over?

wp1/selection/models/wikiproject.py
```diff
@@ -0,0 +1,57 @@
...
+
+  def build(self, content_type, **params):
+    if content_type != 'text/tab-separated-values':
+      raise Wp1FatalSelectionError('Unrecognized content type')
+    if 'wp10db' not in params:
+      raise Wp1FatalSelectionError('Missing param `wp10db`')
+    if 'add' not in params:
+      raise Wp1FatalSelectionError('Missing param `add`')
+
+    add = set()
```
benoit74, line 29: I would rename that `added_articles` to make it clear that it is a set of articles (if I'm not mistaken)

wp1/selection/models/wikiproject.py
```diff
@@ -0,0 +1,57 @@
...
+    for project in params['add']:
+      add.update(_get_articles_for_project(params['wp10db'], project))
+
+    subtract = set()
```
benoit74, line 33: `substracted_articles`

wp1/selection/models/wikiproject.py
```diff
@@ -0,0 +1,57 @@
...
+    for project in params.get('subtract', []):
+      subtract.update(_get_articles_for_project(params['wp10db'], project))
+
+    return '\n'.join(add.difference(subtract)).encode('utf-8')
+
+  def validate(self, **params):
+    if 'wp10db' not in params:
+      raise Wp1FatalSelectionError('Cannot validate without param `wp10db`')
```
benoit74, line 41: This could probably be covered quite easily (even if as we said yesterday, coverage is not a goal in itself)

benoit74 (changes requested):
I find the `add` and `substract` semantics a bit weird. I'm obviously biased on this, but I like browsertrix crawler semantic which is using `include` and `exclude` terms, which I find clearer. I don't mind if you prefer to use `add` and `substract`, I just wanted to mention this.

Other than that, just small propositions, looks pretty good to me.

wp1/selection/abstract_builder.py
audiodude: yup, removed.

wp1/selection/models/wikiproject.py
audiodude: Done.

wp1/selection/models/wikiproject.py
audiodude: Done.

wp1/selection/models/wikiproject.py
audiodude: Done.

audiodude (commented):
Thanks for the review. I agree that include/exclude is probably easier to understand.

benoit74 (approved):
LGTM, I let you merge this at your convenience

PR #730, opened 2024-05-03
Frontend for WikiProject Builder

audiodude:
The WP1 frontend is written in Vue js (version 2). We have created a component, which essentially acts as a standalone "page", for each type of Builder. The component for each Builder handles both create/update. Conceptually, the individual Builder components "inherit from" the BaseBuilder.vue, though the mechanism is actually more like parameterization.

The WP1 API, which corresponds to Python functions in wp1.web, operates on instances of Builders generically. Similar to how the backend Python code works, builders simply have a list of properties that are common to all of them, with the `params` key in the request/response JSON varying.

The BaseBuilder handles concerns that are common to all Builder components, like providing instruction text, allowing the user to select a Wikimedia project that the Builder applies to, naming and validating the name of the Builder, doing the POST request and responding to generic errors, as well as simpler concerns such as disabling the save button and showing a spinner while the network request is processing.

WIth this design, the subclassed Builder component provides an HTML block for the specification of it's unique parameters. In this PR, that is the `#add-items` and `#subtract-items` text fields. The subclass also provides hooks for pre and post processing of the input. For example, the text field is a newline separated list of WikiProjects, which is `split('\n')` to produce an array for the JSON request. For updates, when the model is loaded from the API, there is a corresponding `join(\n)` that readies the data for display and editing in the text fields.

A notable addition from this PR is the `projectFilter` property on BaseBuilder. This allows child components to provide a filter function (function that corresponds to the [Array.filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) API) in order to select which Wikimedia projects the Builder is valid for. This is necessary because the WikiProject Builder only works on en.wikipedia.org. The BaseBuilder defaults to an "empty" filter that returns true for every element if the filter is undefined.

The cypress tests naturally have a large degree of duplication, since they all test pages that are based off of BaseBuilder. However, conceptually, it is appropriate for each page to test that, for example, the builder cannot be saved without a name, even though we know this functionality is implemented in BaseBuilder for the following reasons:

1. We can't test BaseBuilder directly, and
2. In the future, the implementation may change or be refactored, but we would still want this property to hold true.

### DIFF
--- a/wp1-frontend/src/components/WikiProjectBuilder.vue
+++ b/wp1-frontend/src/components/WikiProjectBuilder.vue
@@ -23,12 +23,12 @@
     </template>
     <template #extra-params>
       <div id="lists" class="form-group m-4">
-        <label for="add-items">WikiProjects to include</label>
+        <label for="include-items">WikiProjects to include</label>
         <textarea
-          id="add-items"
-          ref="addItems"
+          id="include-items"
+          ref="includeItems"
           class="form-control my-2"
-          v-model="addText"
+          v-model="includeText"
           rows="5"
           required
         ></textarea>
@@ -37,12 +37,12 @@
           Please provide WikiProjects to include
         </div>
 
-        <label for="subtract-items">WikiProjects to exclude</label>
+        <label for="exclude-items">WikiProjects to exclude</label>
         <textarea
-          id="subtract-items"
-          ref="subtractItems"
+          id="exclude-items"
+          ref="excludeItems"
           class="form-control my-2"
-          v-model="subtractText"
+          v-model="excludeText"
           rows="5"
         ></textarea>
       </div>
@@ -58,19 +58,19 @@ export default {
   name: 'WikiProjectBuilder',
   data: function () {
     return {
-      addText: '',
-      subtractText: '',
+      includeText: '',
+      excludeText: '',
       invalidItems: '',
       params: {},
     };
   },
   methods: {
     onBuilderLoaded: function (builder) {
-      this.addText = builder.params.add.join('\n');
-      this.subtractText = builder.params.subtract.join('\n');
+      this.includeText = builder.params.include.join('\n');
+      this.excludeText = builder.params.exclude.join('\n');
     },
     onBeforeSubmit: function () {
-      this.$refs.addItems.setCustomValidity('');
+      this.$refs.includeItems.setCustomValidity('');
     },
     onValidationError: function (data) {
       this.invalidItems = data.items.invalid.join('\n');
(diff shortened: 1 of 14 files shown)


wp1-frontend/src/components/WikiProjectBuilder.vue
```diff
@@ -0,0 +1,103 @@
...
+    @onValidationError="onValidationError"
+  >
+    <template #create-desc>
+      <p>
+        Use this tool to create an article selection list for English Wikipedia
+        only (other languages not supported), based off the articles included in
+        the specified
+        <a href="https://en.wikipedia.org/wiki/WikiProject">WikiProjects</a>.
+        Your selection will be saved in public cloud storage and can be accessed
+        through URLs that will be provided once it has been saved.
```
benoit74, line 21: Is there really multiple URLs provided? I think there is only one and would suggestion to replace `through URLs` => `through a URL` (singular)

wp1-frontend/src/components/WikiProjectBuilder.vue
```diff
@@ -0,0 +1,103 @@
...
+    @onBeforeSubmit="onBeforeSubmit"
+    @onValidationError="onValidationError"
+  >
+    <template #create-desc>
+      <p>
+        Use this tool to create an article selection list for English Wikipedia
+        only (other languages not supported), based off the articles included in
+        the specified
+        <a href="https://en.wikipedia.org/wiki/WikiProject">WikiProjects</a>.
+        Your selection will be saved in public cloud storage and can be accessed
```
benoit74, line 20: I tend to prefer `online` to `public cloud storage`. It is obviously less precise, but I don't think the precision does not bring much value but fear / lack of understanding for non-tech users. Just a suggestion.

benoit74: And you could replace `can be accessed` by `will be accessible`

benoit74 (commented):
LGTM, only few suggestions.

wp1-frontend/src/components/WikiProjectBuilder.vue
audiodude: Done

wp1-frontend/src/components/WikiProjectBuilder.vue
audiodude: Done

audiodude (commented):
Thanks for the review. I made the recommended language changes, through all Builder components (because it was copied in each of them).

benoit74 (approved):
LGTM, I let you merge this at your convenience

PR #754, opened 2024-08-13
Upgrade axios for dependabot

audiodude:


audiodude:
No patched version yet

Issue #171, opened 2020-05-14
Integrate WP1 selection tools


kelson42:
Available at https://github.com/openzim/wp1_selection_tools
Using the Python framework.

kelson42:
@audiodude One of a first step would be maybe to integrate its source with the same repository and put it in the `docker-compose.yml` ?

audiodude:
We'd like to start this task in earnest.

- [x] Design doc written (link redacted)
- [x] We have requested a Trove database in Wikimedia Cloud VPS, to allow for the growth of our article tables
- [x] Some foundational work has already been explored with getting the pageviews database ingested into the WP1 db

Data ingestion of views, links, and lang links can happen on a monthly cadence to keep parity with download.openzim.org. During this process, we should produce tsv "dump" files in the same format as the current wp1_selection_tools, and preferably upload them to that website. As a side effect, we can also write the data to the page_scores table described in the design doc, since we will already have the calculated values.

kelson42:
@audiodude I see nothing I disagree in your comment. I try to phrase it differently to ensure we are aligned:
* External sources (databases mostly) need to scraped/scanned to extract the information we need for publishing TSV at https://download.openzim.org/wp1 and save them in our DB
* One this done, we will generate our custom selections and scoring:
** `all.tsv.zip`
** `custom/*tsv`
** `tops/*tsv`
* Then finally we will put in our schedulers, TSV export routines to keep publishing straight:
** `all.tsv.zip`
** `langlinks.tsv.zip
** `pagelinks.tsv.zip`
** `pages.tsv.zip`
** `pageviews.tsv.zip`
** `projects.zip`
** `redirects.tsv.zip`

kelson42:
@audiodude Completing this task is critical because:
* We can not generate new selections because (at least) https://github.com/openzim/wp1_selection_tools/issues/48
* We have lost old data after the massage storage outage of December.

We need this selection to create all selection TOP, thematics, etc... configured in the Zimfarm

PR #756, opened 2024-08-18
Set up db connections to use SOCKS proxy in Development environment

audiodude:
Previously, the development environment only needed to connect to the English wikipedia replica database, so it was enough to set up a ssh tunnel and connect through that. However, in future versions of the code, we will need to connect to a variety of wiki replica database, in order to select `links` and `lang_links`. So, for example, for French wikipedia, we will need to connect to `frwiki.analytics.db.svc.wikimedia.cloud`. In fact, we will need to connect to all db replicas for all wikipedia languages.

This code will not be used in production, it is only for development. In production, the `*.analytics.db.svc.wikimedia.cloud` hosts are all reachable from the production cloud VPS server. For local testing and debugging, we need the SOCKS proxy.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -30,6 +30,7 @@ attrs = "==18.2.0"
 pytest-cov = "*"
 supervisor = "~=4.2"
 yoyo-migrations = "==7.3.2"
+pysocks = "*"
 
 # Versions required for dependabot
 sqlparse = "~=0.5.0"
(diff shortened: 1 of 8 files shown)


audiodude:
Part of #171

audiodude:
@rgaudin can you do a sanity check please?

rgaudin (approved):
LGTM

PR #771, opened 2024-10-31
Use `docker compose` instead  of `docker-compose`

benoit74:
Fix documentation, `docker-compose` command does not exists anymore now that compose is part of docker, one has to use `docker compose`


audiodude:
This is not completely correct. The restoration machine was installed with the latest version of docker, which uses `docker compose`. But if you are deploying on the current box, it's `docker-compose`. I didn't install docker on the existing `mwcurator` box, so I don't know how to update it.

benoit74:
Then let's park this PR until we've updated it ^^

PR #805, opened 2025-03-13
Added logic to perform entire pageview update in one transaction

harsh-0409:
Applied transaction management to the update_pageviews function in order to ensure data integrity during the update process.


wp1/scores.py
```diff
@@ -192,17 +192,27 @@ def update_pageviews(filter_lang=None, commit_after=50000):
...
+  # Start a transaction
+  wp10db.begin()
+
+  try:
+    for lang, article, page_id, views in pageview_components():
+      if filter_lang is None or lang == filter_lang:
+        update_db_pageviews(wp10db, lang, article, page_id, views)
+
+      n += 1
+      if n >= commit_after:
```
audiodude, line 205: This doesn't work, because if we commit after every 50,000 rows, then we are by definition not commiting the entire update in one transaction.

audiodude (changes requested):

wp1/scores.py
harsh-0409: Hey, I have updated the function and will send a new PR as I accidentally closed this PR. Really sorry for the delay, I was travelling for a few days.

Issue #721, opened 2024-04-20
Extend URL stripping in Simple selection builder to non EN projects


audiodude:
Currently there is code in the Simple Selection validator that strips the following prefixes from article names:

`https://en.wikipedia.org/wiki/`
`https://en.wikipedia.org/w/index.php?title=`

This is a courtesy to prevent invalid article names from making it to the ZIM farm, however it should be obvious that this only works for en.wikipedia.org. Instead, we should use the project name associated with the builder to strip.

PR #807, opened 2025-03-17
feat: extend URL stripping to support non-EN Wikipedia projects

arjitcodes:
Fixes: #721 

This PR updates the Simple Selection validator to dynamically strip Wikipedia article URLs based on the project name instead of hardcoding `en.wikipedia.org`. This allows the validator to support non-English Wikipedia projects (e.g., `fr.wikipedia.org`, `de.wikipedia.org`, etc.).
Related Issues

- [X] Code follows project coding standards
- [X] Changes have been tested

### DIFF
--- a/wp1/selection/models/simple.py
+++ b/wp1/selection/models/simple.py
@@ -69,28 +69,15 @@ class Builder(AbstractBuilder):
         invalid_article_names.append(decoded_item)
         is_valid = False
       if is_valid:
-        article_name = self.extract_article_name(decoded_item)
-        if not article_name:
-          invalid_article_names.append(decoded_item)
-        else:
-          valid_article_names.append(article_name)
+        project_domain = 'en.wikipedia.org'
+        if 'project' in params:
+          project_domain = params['project']
+        wiki_prefix = f"https://{project_domain}/wiki/"
+        index_prefix = f"https://{project_domain}/w/index.php?title="
+        article_name = decoded_item.replace(wiki_prefix,
+                                            "").replace(index_prefix, "")
+        valid_article_names.append(article_name)
     if forbidden_chars:
       errors.append('The list contained the following invalid characters: ' +
                     ', '.join(forbidden_chars))
     return (valid_article_names, invalid_article_names, errors)
-
-  def extract_article_name(self, url):
-    if not url:
-      return None
-
-    parsed_url = urllib.parse.urlparse(url)
-    domain = parsed_url.netloc
-    if not domain:
-      return url
-    if not domain.endswith("wikipedia.org"):
-      return None
-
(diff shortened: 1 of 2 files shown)


audiodude (changes requested):
This is a great idea, and I definitely appreciate the PR! Obviously, stripping en.wikipedia.org from the URL only works for selections using the `en.wikipedia.org` project, so while it's not necessarily a bug, this "helpful" code is useless for other projects.

Thank you for the test cases too. They helped me quickly realize that the logic isn't quite right.

All WP1 selections have a specific Wikimedia project that they operate for. You can see the list here: 

image

So really, we don't need to parse the url to get the netloc, because the article list should only contain articles in the domain of the project that was selected.

That project is passed inside the `**params` dict as `'project'` as you can see here:
https://github.com/openzim/wp1/blob/main/wp1/web/builders.py#L38

So overall, great idea and great work, and I would only ask that you tweak the logic so that it checks for the domain of the chosen project.

arjitcodes:
Thank you for the feedback! I’ve updated the logic. 

But I was wondering—why are we not checking whether the project name matches the domain in the URL? For example, if the selected project is `en.wikipedia.org`, but an article from `de.wikipedia.org` (e.g.`https://de.wikipedia.org/wiki/Europa`) is included, it currently gets added to `valid_article_names`. Shouldn't it instead be marked as invalid?

Just curious about the reasoning behind this—let me know your thoughts!

audiodude:
> Thank you for the feedback! I’ve updated the logic.
> 
> But I was wondering—why are we not checking whether the project name matches the domain in the URL? For example, if the selected project is `en.wikipedia.org`, but an article from `de.wikipedia.org` (e.g.`https://de.wikipedia.org/wiki/Europa`) is included, it currently gets added to `valid_article_names`. Shouldn't it instead be marked as invalid?
> 
> Just curious about the reasoning behind this—let me know your thoughts!

The URL stripping is just a sort of courtesy. It's kind of an afterthought. We don't really expect people to paste in full URLs in general, and I'm not sure where they would get a `fr.wikipedia.org` URL from if they were building a `de.wikipedia.org` list anyways. You may also notice that there's no validation that anything typed into the box is an actual article on any given Wikimedia project. If it's not, it will fail later in the ZIM stage.

audiodude (approved):
LGTM

arjitcodes:
Just a small ping to check if anything is blocking the merge or if I should wait for further steps. Let me know!

arjitcodes:
This PR has been hanging for a while—it looks ready to merge now.

audiodude:
> This PR has been hanging for a while—it looks ready to merge now.

Quite right, thanks for the ping.

Issue #790, opened 2025-01-27
Task requested on farm.zimit.kiwix.org with too much memory


benoit74:
We have a task on farm.zimit.kiwix.org which requests more memory than mwoffliner workers have available:

![Image](https://github.com/user-attachments/assets/7e685e20-7768-47f6-9b43-a01189a4d255)

![Image](https://github.com/user-attachments/assets/bb17771b-b740-4a00-acd1-dc191f28b531)

I removed the task on the farm.

Is it an issue to fix on WP1 side or was this just a test?

audiodude:
Here's the logic that WP1 uses for requesting resources:

https://github.com/openzim/wp1/blob/main/wp1/logic/selection.py#L187

I don't remember how that was determined, but you can see that it's variable.

kelson42:
To me it seems that this would have more than 5M of lines! We should not propose to do that. This infra is not conceived to make very large ZIM files.

audiodude:
Should we get rid of the dynamic resource sizing attempt and just request max resources every time?

What are reasonable limits for the number of articles we allow for a ZIM? I think the UI would need some updating if we were to establish such limits.

benoit74:
We have currently two tasks "abusing" (the word is a bit strong) our infrastructure:
- https://farm.zimit.kiwix.org/pipeline/6123148d-1d10-4753-9b11-25101fcd606e/debug : 443497 articles
- https://farm.zimit.kiwix.org/pipeline/65334923-5200-40d2-a58d-07c34ab6a0a9/debug : 242146 articles, 534990 files

I think that we should put a cap on number of articles in the selection, aiming at the 2 hours limit we already have set for zimit.kiwix.org (a basic rule of thumb is probably pretty easy to compute). This limit should ideally be set per user, so that we can grant ourselves more freedom, and that so that we can sell more time/articles to the ones who are ready to pay for it.

Given the difficulties we have to update wikipedia ZIMs currently, it is probably even wise to do it quite quickly, sooner or later someone will have the idea of using wp1 to create his own updated ZIM.

kelson42:
To ke this issue is the "sister" of #794 and should be fixed in priority. The message should invite users to open a request to openzim/zim-requests.

audiodude:
> To ke this issue is the "sister" of [#794](https://github.com/openzim/wp1/issues/794) and should be fixed in priority. The message should invite users to open a request to openzim/zim-requests.

Yes these issues are closely related.

[...]

PR #814, opened 2025-03-23
Use static resource profile for requesting ZIM creation

### LINKED ISSUE #794: Unable to create zim files
Hello,

After trying many time for the following, they keep failing:

Image

- Other biology
  Log: https://farm.zimit.kiwix.org/pipeline/2ef6896e-acd0-4a8f-8edf-21d3fd8d4aeb/debug
  Project: https://wp1.openzim.org/#/selections/wikiproject/6ed5ef65-5829-4fc9-95ed-865aab7a2d37
- Lists
  Log: https://api.farm.zimit.kiwix.org/v1/tasks/96d8e7b1-c69e-4b99-9e6a-c55b3eecc7ab
  Project: https://wp1.openzim.org/#/selections/wikiproject/90b881b8-985f-4f11-ac6d-cbf613d394fb
- Military
  Log: https://api.farm.zimit.kiwix.org/v1/tasks/65334923-5200-40d2-a58d-07c34ab6a0a9
  Project: https://wp1.openzim.org/#/selections/wikiproject/696e76b7-1284-4e1f-94fd-57906f9260d9

audiodude:
Fixes #794.

Originally, we were trying to estimate how many resources we would need on Zimfarm based on the size of the selection. It seems the estimates were based on the idea that the max resource allocation of, say, 15 GB of memory would be enough to scrape almost all of English Wikipedia, so 3 GB would be enough to scrape 1M articles, or about 1/5 of that. This is obviously not right, and #790 makes it clear that there should be a certain maximum limit on the number of articles that can be requested through WP1.

Instead of doing any kind of heuristic calculation, we now just statically request the maximum amount of resources that we happen to know (from outside channels) the Zimfarm is able to support.

### DIFF
--- a/wp1/logic/selection.py
+++ b/wp1/logic/selection.py
@@ -5,16 +5,16 @@ import urllib.parse
 
 import attr
 
+from wp1 import zimfarm
 from wp1.constants import CONTENT_TYPE_TO_EXT, TS_FORMAT_WP10, ZIM_FILE_TTL
+from wp1.logic import util
 from wp1.models.wp10.selection import Selection
 from wp1.models.wp10.zim_file import ZimFile
 from wp1.storage import connect_storage
-from wp1.logic import util
 from wp1.timestamp import utcnow
-from wp1 import zimfarm
 
 try:
-  from wp1.credentials import ENV, CREDENTIALS
+  from wp1.credentials import CREDENTIALS, ENV
   S3_PUBLIC_URL = CREDENTIALS.get(ENV, {}).get('CLIENT_URL', {}).get(
       's3', 'http://credentials.not.found.fake')
 except ImportError:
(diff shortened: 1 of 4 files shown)


benoit74:
@kelson42 I think we should discuss what is a fair usage of wp1. Being a free service, I wonder if we should apply any sort of limit (e.g. in terms of number of articles). This PR is opening the door to much bigger ZIMs, and I'm not sure this is really what we want (at least not without a safety net).

audiodude:
> @kelson42 I think we should discuss what is a fair usage of wp1. Being a free service, I wonder if we should apply any sort of limit (e.g. in terms of number of articles). This PR is opening the door to much bigger ZIMs, and I'm not sure this is really what we want (at least not without a safety net).

I agree, see https://github.com/openzim/wp1/issues/790#issuecomment-2746295558

benoit74:
We've discussed the matter live yesterday with kelson and the outcome is:
- I must provide the updated heuristic about which memory / disk / cpu to assign ; it is still unclear if we need something depending on selection size or just a flat profile, but at least we do not want to use "maximum available on worker"
  - I will come back to you asap on this
- I've opened https://github.com/openzim/wp1/issues/816 about the topic of limiting WP1 to reasonable requests

kelson42:
@audiodude Last time it worked, WPEN was scrapable in 16GB AND the memory consumption was related (if not proportional) to the number of artivles. The current situation is the result of changes in MWoffliner.

audiodude:
Superseeded by #858

PR #815, opened 2025-03-23
Perform entire pageview update in one transaction

### LINKED ISSUE #761: Perform entire pageview update in one transaction
We'd like to keep all of the cumulative pageviews in one column per article. However, this poses a problem if a update process gets interrupted, because we wouldn't know which rows have already been incremented with this month's data.

The solution is to do the entire update as one transaction, so that if the database updates, it means that all rows were successfully read and processed.

arjitcodes:
Fixes #761 

This PR ensures atomic pageview updates while handling database size limitations. Instead of committing every 50,000 rows directly to `scores_db`, updates are first written to a temporary table (`temp_db`) in batches.

1. If successful, `temp_db` is swapped into `scores_db`.
2. If failed, `temp_db` is cleared, preventing partial updates.
3. This avoids exceeding the 1GB undo log limit while ensuring data integrity.
---
- [X] **Data swap test**: `test_swap_temp_pageviews_to_scores`
- [X] **Update pageviews test**: `test_update_pageviews`
- [X] **Error handling test**: `test_update_pageviews_rollback_on_error`
- [X] **Temp DB cleared test**: `test_truncate_temp_pageviews`
- [X]  **Insert into temp DB test**: `test_insert_temp_pageviews`

### DIFF
--- a/wp1/scores_test.py
+++ b/wp1/scores_test.py
@@ -256,9 +256,84 @@ class ScoresTest(BaseWpOneDbTest):
 
     self.assertEqual(expected, actual)
 
-  def test_update_db_pageviews(self):
-    scores.update_db_pageviews(self.wp10db, 'en', 'Statue_of_Liberty', 1234,
-                               100)
+  def test_reset_missing_articles_pageviews(self):
+    with self.wp10db.cursor() as cursor:
+      cursor.execute(
+          'INSERT INTO page_scores (ps_lang, ps_article, ps_page_id, ps_views)'
+          '''VALUES
+        ("en", "Statue_of_Liberty", 1234, 100),
+        ("fr", "Tour_Eiffel", 5678, 200),
+        ("es", "Sagrada_Familia", 9101, 150)''')
+
+    with self.wp10db.cursor() as cursor:
+      cursor.execute(
+          'INSERT INTO temp_pageviews (tp_lang, tp_article, tp_page_id, tp_views)'
+          '''VALUES
+        ("fr", "Tour_Eiffel", 5678, 200),
+        ("es", "Sagrada_Familia", 9101, 150)''')
+
+    scores.reset_missing_articles_pageviews(self.wp10db)
+
+    with self.wp10db.cursor() as cursor:
+      cursor.execute('SELECT COUNT(*) as cnt FROM page_scores')
+      n = cursor.fetchone()['cnt']
+      self.assertEqual(3, n)
+
+    with self.wp10db.cursor() as cursor:
+      cursor.execute('SELECT * FROM page_scores WHERE ps_page_id = 1234')
+      result = cursor.fetchone()
+      self.assertIsNotNone(result)
+      self.assertEqual(result['ps_lang'], b'en')
+      self.assertEqual(result['ps_article'], b'Statue_of_Liberty')
+      self.assertEqual(result['ps_page_id'], 1234)
+      self.assertEqual(result['ps_views'], 0)
+
+  def test_insert_temp_pageviews(self):
+    scores.insert_temp_pageviews(self.wp10db, 'en', 'Statue_of_Liberty', 1234,
+                                 100)
+
+    with self.wp10db.cursor() as cursor:
+      cursor.execute('SELECT * FROM temp_pageviews WHERE tp_page_id = 1234')
+      result = cursor.fetchone()
+      self.assertIsNotNone(result)
+      self.assertEqual(result['tp_lang'], b'en')
+      self.assertEqual(result['tp_article'], b'Statue_of_Liberty')
+      self.assertEqual(result['tp_page_id'], 1234)
+      self.assertEqual(result['tp_views'], 100)
+
+  def test_insert_temp_pageviews_existing(self):
+    with self.wp10db.cursor() as cursor:
+      cursor.execute(
+          'INSERT INTO temp_pageviews (tp_lang, tp_article, tp_page_id, tp_views)'
+          'VALUES ("en", "Statue_of_Liberty", 1234, 100)')
+
+    scores.insert_temp_pageviews(self.wp10db, 'en', 'Statue_of_Liberty', 1234,
+                                 200)
+
+    with self.wp10db.cursor() as cursor:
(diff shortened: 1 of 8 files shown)


wp1/scores_test.py
```diff
@@ -256,40 +256,85 @@ def test_pageview_components_errors(self, mock_file_open):
...
+        ("en", "Statue_of_Liberty", 1234, 100),
+        ("fr", "Tour_Eiffel", 5678, 200),
+        ("es", "Sagrada_Familia", 9101, 150)''')
+
+    scores.swap_temp_pageviews_to_scores(self.wp10db)
+
+    with self.wp10db.cursor() as cursor:
+      cursor.execute('SELECT COUNT(*) as cnt FROM page_scores')
+      n = cursor.fetchone()['cnt']
+      self.assertEqual(3, n)
```
audiodude, line 304: Can you also assert at least one row of the actual values?

wp1/scores_test.py
```diff
@@ -299,11 +344,55 @@ def test_update_pageviews(self, mock_components, mock_download):
...
+      self.assertEqual(0, n)
+
+  @patch('wp1.scores.wp10_connect')
+  @patch('wp1.scores.insert_temp_pageviews', side_effect=Exception("DB error"))
+  @patch('wp1.scores.truncate_temp_pageviews',
+         side_effect=scores.truncate_temp_pageviews)
+  @patch('wp1.scores.download_pageviews')
+  @patch('wp1.scores.pageview_components')
+  def test_update_pageviews_rollback_on_error(self, mock_components,
+                                              mock_download, mocke_truncate,
```
audiodude, line 368: nit: typo, `mocke_truncate`

wp1/scores_test.py
```diff
@@ -256,40 +256,85 @@ def test_pageview_components_errors(self, mock_file_open):
...
+  @patch('wp1.scores.truncate_temp_pageviews',
+         side_effect=scores.truncate_temp_pageviews)
   @patch('wp1.scores.download_pageviews')
   @patch('wp1.scores.pageview_components')
-  def test_update_pageviews(self, mock_components, mock_download):
+  def test_update_pageviews(
+      self,
+      mock_components,
+      mock_download,
+      mock_truncate,
```
audiodude, line 334: I would prefer a test that uses the test database (`self.wp10db`) and actually performs the truncate, swap and insert.

audiodude (approved):
This looks really good! Good job especially on updating the `wp10_test*.sql` files and adding the db migration.

benoit74:
Well done! I might be wrong, but I don't see where you "reset" pageviews. Should an article go missing, it looks like it will not be removed or updated by the "big swap" and will keep its old pageview count forever.

audiodude:
> Well done! I might be wrong, but I don't see where you "reset" pageviews. Should an article go missing, it looks like it will not be removed or updated by the "big swap" and will keep its old pageview count forever.

Good catch! Yes, if the article doesn't appear in the current month's data, it should have a pageview count of 0.

benoit74:
> it should have a pageview count of 0.

or be simply removed from table to not clog the table with deleted / renamed articles ?

audiodude:
> > it should have a pageview count of 0.
> 
> or be simply removed from table to not clog the table with deleted / renamed articles ?

Sorry no, this is wrong. The scores table, currently, only contains pageviews. But in the future it will contain all of the score components necessary to calculate things like "top 1000 Wikipedia articles", so it's not appropriate to remove a row from the table just because it received no pageviews.

wp1/scores_test.py
```diff
@@ -299,11 +384,55 @@ def test_update_pageviews(self, mock_components, mock_download):
     scores.update_pageviews(commit_after=2)
 
     mock_download.assert_called_once()
+    mock_swap.assert_called_once()
```
audiodude, line 387: Please update this test to not mock the relevant database methods. You should not have:

```
mock_truncate, mock_swap, mock_insert, mock_reset
```

This test should be a full test of updating the pageviews, with those methods actually performing their actual jobs.

audiodude:
Just one comment and we should be ready to merge!

wp1/scores_test.py
arjitcodes: The test ensures the database operations run as expected while also tracking their calls. The patched methods still execute their actual logic using `side_effect`:
```
@patch('wp1.scores.reset_missing_articles_pageviews', side_effect=scores.reset_missing_articles_pageviews)
@patch('wp1.scores.insert_temp_pageviews', side_effect=scores.insert_temp_pageviews)
@patch('wp1.scores.swap_temp_pageviews_to_scores', side_effect=scores.swap_temp_pageviews_to_scores)
@patch('wp1.scores.truncate_temp_pageviews', side_effect=scores.truncate_temp_pageviews)
```
This approach keeps the test fully functional while verifying call counts. Let me know if you have any concerns!

arjitcodes:
> The test ensures the database operations run as expected while also tracking their calls. The patched methods still execute their actual logic using `side_effect`:
> 
> ```
> @patch('wp1.scores.reset_missing_articles_pageviews', side_effect=scores.reset_missing_articles_pageviews)
> @patch('wp1.scores.insert_temp_pageviews', side_effect=scores.insert_temp_pageviews)
> @patch('wp1.scores.swap_temp_pageviews_to_scores', side_effect=scores.swap_temp_pageviews_to_scores)
> @patch('wp1.scores.truncate_temp_pageviews', side_effect=scores.truncate_temp_pageviews)
> ```
> 
> This approach keeps the test fully functional while verifying call counts. Let me know if you have any concerns!

Do I still need to remove the mocks in this case?

audiodude:
> Do I still need to remove the mocks in this case?

Sorry, I missed the fact that you were "calling through" to the actual implementations. That's a good thing.

In general, my opinion is that tests should not measure the specific order of dependent calls, or the number of times they are called, or the specific arguments they are called with, except in the case that the actual implementation isn't being used or there is nothing else to assert against.

In this case, I think the mocks should be removed, yes.

arjitcodes:
This PR looks ready. Please review and merge when you have a moment.

Issue #816, opened 2025-03-25
Limit the size of the selections we can ZIM through WP1


benoit74:
While WP1 usage has been very limited / confidential so far, it begins to get some traction and we begin to see very big selections being requested.

WP1 being so far a free tool, we cannot accept to use unlimited resources for unlimited size.

We need to:
- define a limit of acceptable selection size which can be ZIMed
  - a guess estimate is that we process around 400 to 500 articles per minute on wikipedia.org projects, and since this is our most popular project, I propose to limit to 50k articles (more or less 2 hours), which has the advantage to be also the limit for our top 50k selections
- define a message we want to display when this limit is hit
  - @Popolechien we need you input on this
  - we've agreed that for now we will redirect users to create a zim-request when they are blocked, we do not know yet if this should/could become a paid offer or just wait for the zim-request to be processed for free (which might take too long for some users) or just ask more resources to the foundation for a dedicated worker
- block users requesting a ZIM bigger than this at API + UI level with this message

Popolechien:
Here's some proposed text:

> Oh no! It seems that you have hit the limit for the number of articles :-(
>
> Could it be that such a big selection could be useful to others? How about opening a [zim-request](https://github.com/openzim/zim-requests/issues) so that we look into it, and add straight to the Kiwix library?

benoit74:
Shouldn't we add the limit in the text so that users know what to aim for ?

Popolechien:
With a limit at 50k articles I would assume that changes won't be made at the 1-2 units margin, but ok. Edited the text a bit.

kelson42:
@benoit74 Beside #790 and #794, this issue is indeed part of the solution triptic to fix our memory mgmt regression at WP1. I hope we can implement them next week.

benoit74:
I don't think I'm the good person to implement this, to me this is something to be done at WP1 level and I have absolutely no knowledge of the code. Since this needs both UI and backend changes ...

kelson42:
@audiodude @benoit74 @rgaudin We have currently a pretty broken WP1 ZIM building feature since weeks meanwhile. My input is: we have to find a solution to fix it very soon. This is a priority to fix/implement them.

audiodude:
@kelson42 @benoit74 I am actively working on this issue.

audiodude:
I've merged #818 and deployed it, so now all selections have an `s_article_count` column that lists the number of articles in the selection, which was backfilled by a database migration and is updated any time the selection is built.

[...]

PR #818, opened 2025-03-25
Calculate and store article counts when materializing selections.

audiodude:
Includes migration code to retrieve all selections from S3 and populate article counts.

First backend part of addressing #816.


audiodude:
Merging without review after 2 weeks.

PR #823, opened 2025-03-29
refactor: store logs as Redis hashes

### LINKED ISSUE #49: Assessment logs should be written in Redis instead of the database
That way we could:
* Speed up the writting/reading process of logs
* Secure easily that old logs are not saved for years without any reason
* Better SW architecture where temporary data are stored in temporary storage

### LINKED ISSUE #778: Logging table should be periodically truncated
The `logging` table is set up to grow indefinitely. Recently we had an outage because it got too full:

```
+--------------------+---------------------------------------+------------+
| Database           | Table                                 | Size in MB |
+--------------------+---------------------------------------+------------+
| enwp10_prod        | logging                               |   42164.95 |
| enwp10_prod        | ratings                               |    8176.63 |
| enwp10_prod        | moves                                 |     750.84 |
| enwp10_prod        | global_articles                       |     686.00 |
| enwp10_prod        | selection_data                        |     181.89 |
| enwp10_prod        | cache                                 |      91.86 |
| enwp10_prod        | namespacename                         |      10.55 |
| enwp10_prod        | categories                            |       8.53 |
| enwp10_prod        | reviews                               |       5.03 |
| enwp10_prod        | workingselection                      |       2.52 |
| enwp10_prod        | automatedselection                    |       2.52 |
| enwp10_prod        | builders                              |       2.52 |
| enwp10_prod        | manualselectionlog                    |       0.77 |
| enwp10_prod        | projects                              |       0.34 |
| enwp10_prod        | manualselection                       |       0.30 |
| enwp10_prod        | selections                            |       0.13 |
| enwp10_prod        | zim_files                             |       0.08 |
| enwp10_prod        | users                                 |       0.06 |
| enwp10_prod        | global_rankings                       |       0.03 |
| enwp10_prod        | releases                              |       0.03 |
+--------------------+---------------------------------------+------------+
```

This was accelerated by the fact that "Articles by Quality" and "Category class" templates were merged, which basically generated a superfluous log entry for every article. However, even if that hadn't happened, the logs table would have eventually  run out of space, since it only grows over time.

Since we only ever show logs for the past 7 days, we don't need to hold data older than 7 days. Although #49 would make this easier and bring its own benefits, at the very least we could have a cron job that deletes data older than 7 days, or do it as part of generating the logs.

elfkuzco:
# Rationale
This PR implements the functionality to save logs to Redis. By setting expiry on the logs while saving, we can be sure that logs are not saved for years and there is no need for reliance on a cron job to delete old logs.

## Changes
- set up connection to Redis in `BaseWPOneDbTest`
- save logs to Redis as hashes instead of SQL DB
- use a key naming pattern which allows us to filter logs that match the pattern. This emulates SQL-like `AND` queries for logs
- update tests to use appropriate `self.redis` 

## Fixes
This fixes #778 
This fixes #49

### DIFF
--- a/wp1/base_db_test.py
+++ b/wp1/base_db_test.py
@@ -8,6 +8,10 @@ import pymysql
 from wp1.environment import Environment
 from wp1.models.wp10.selection import Selection
 
+from wp1.redis_db import connect as redis_connect
+
+from wp1.models.wp10.rating import Rating
+
 logger = logging.getLogger(__name__)
 
 try:
@@ -93,9 +97,27 @@ class BaseWpOneDbTest(WpOneAssertions):
         cursor.execute(stmt)
     self.wp10db.commit()
 
+  def connect_redis_db(self):
+    if ENV != Environment.TEST:
+      raise ValueError(
+          'Database tests destroy data! They should only be run in the TEST env'
+      )
+    return redis_connect()
+
+  def _setup_redis_db(self):
+    self.redis = self.connect_redis_db()
+    self.redis.ping()
+    self.redis.flushdb()
+
+  def _cleanup_redis_db(self):
+    self.redis.flushdb()
+
   def setUp(self):
     self.addCleanup(self._cleanup_wp_one_db)
     self._setup_wp_one_db()
+
+    self.addCleanup(self._cleanup_redis_db)
+    self._setup_redis_db()
 
 
 class BaseWikiDbTest(WpOneAssertions):
@@ -143,6 +165,9 @@ class BaseCombinedDbTest(BaseWikiDbTest, BaseWpOneDbTest):
 
     self.addCleanup(self._cleanup_wp_one_db)
     self._setup_wp_one_db()
+
+    self.addCleanup(self._cleanup_redis_db)
+    self._setup_redis_db()
 
 
 def get_first_selection(wp10db):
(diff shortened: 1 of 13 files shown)


elfkuzco:
If the PR is accpeted, there should be a newer PR to drop the logging table. This would be implemented if all the logs are moved from SQL to Redis in order to avoid data loss. Thus, I have avoided the need to include a migration in this PR.

audiodude (changes requested):
This looks great!

But I think we do need the migration in the same PR, because once this gets deployed, we will lose the logs from the past 7 days unless we migrate. We don't need to put the entire logging table into redis, just the logs from the past 7 days. Probably the easiest way to do it is as a YoYo migration, even though it doesn't actually touch the database. That is, we should not drop the logging table as part of the migration, _that_ should be done as a separate manual step.

audiodude:
If you like, I can write the migration and append it to this PR. Just let me know.

elfkuzco:
> If you like, I can write the migration and append it to this PR. Just let me know.

Okay. You can append it to the PR.

audiodude:
Oh actually I don't think I can add to the PR because it's on a branch in your repo. I've opened #826

audiodude (approved):

PR #829, opened 2025-04-01
remove redundant ImportErrors

### LINKED ISSUE #486: Clean up ImportError checks for wp1.credentials in production/test code
After #473, there should always be a `credentials.py` file. In previous iterations, the `credentials.py` file was deleted in order to run the nosetests.

Because of this, there is lots of code like this:

```
try:
  from wp1.credentials import ENV, CREDENTIALS
  S3_PUBLIC_URL = CREDENTIALS.get(ENV, {}).get('CLIENT_URL', {}).get(
      's3', 'http://credentials.not.found.fake')
except ImportError:
  S3_PUBLIC_URL = 'http://credentials.not.found.fake'
```

The point of this code is to provide default values and fallbacks when the credentials file can not be imported. We should clean these up, not only because it is ugly, but because the ImportError code is no longer used and hurts our test coverage.

elfkuzco:
## Rationale
Remove redundant `ImportError` since there is always a `credentials.py`. This fixes #486

### DIFF
--- a/enqueue-all.py
+++ b/enqueue-all.py
@@ -8,13 +8,6 @@ from wp1.redis_db import connect as redis_connect
 
 logger = logging.getLogger(__name__)
 
-try:
-  from wp1.credentials import ENV, CREDENTIALS
-except ImportError:
-  logger.exception('The file credentials.py must be populated manually in '
-                   'order to connect to Redis')
-  raise
-
 
 def main():
   logging.basicConfig(level=logging.INFO)
--- a/enqueue-global.py
+++ b/enqueue-global.py
@@ -4,19 +4,12 @@ from redis import Redis
 from rq import Queue
 
 from wp1.environment import Environment
+from wp1.credentials import ENV, CREDENTIALS
 from wp1 import constants
 from wp1 import tables
 import wp1.logic.project as logic_project
 
 logger = logging.getLogger(__name__)
-
-try:
-  from wp1.credentials import ENV, CREDENTIALS
-except ImportError:
-  logger.exception('The file credentials.py must be populated manually in '
-                   'order to connect to Redis')
-  CREDENTIALS = None
-  ENV = None
 
 
 def main():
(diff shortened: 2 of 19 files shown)


audiodude:
Looks like the CONF_LANG import still needs to be wrapped in a try/except for tests. Please fix.

audiodude:
Wait, actually, aren't all the imports going to fail in CI? Because it doesn't have a `credentials.py`.

elfkuzco:
> Wait, actually, aren't all the imports going to fail in CI? Because it doesn't have a `credentials.py`.
> 
> EDIT: Nvm, we have `credentials.py.e2e` that we copy over in the CI environment. So really it's just the language configuration that needs to be guarded, like I said above.

Given the `CONF_LANG` is set in all the sample credentials, wouldn't it be better to include it in the `credentials.e2e` rather than going back to the try/catch flow. What do you think?

audiodude:
> > Wait, actually, aren't all the imports going to fail in CI? Because it doesn't have a `credentials.py`.
> > EDIT: Nvm, we have `credentials.py.e2e` that we copy over in the CI environment. So really it's just the language configuration that needs to be guarded, like I said above.
> 
> Given the `CONF_LANG` is set in all the sample credentials, wouldn't it be better to include it in the `credentials.e2e` rather than going back to the try/catch flow. What do you think?

Yeah definitely. `CONF_LANG` is a very half baked feature, so I assumed there was some reason why we couldn't have it in CI. But if that works, then great!

audiodude (approved):

elfkuzco:
So, the python test failed because a test was patching `ENV` to None and my commit "blindly" assumed it would always be set. So, I have restored that functionality to check if it is None. That should pass the test.

audiodude (changes requested):
See comment

elfkuzco:
> See comment

I have removed the test since ENV is never going to be None.

audiodude (approved):

audiodude:
Thanks again for the PR and patiently working through the issues!

PR #833, opened 2025-04-02
Fix: Handle expired ZIM files by checking Wasabi storage (Closes #819)

### LINKED ISSUE #819: Unable to download zim
Image
Clicking on any download link

Salmathmullali:
This PR improves the handling of expired ZIM files by pre-fetching the Wasabi storage URL before showing the expiration message. Instead of relying only on the timestamp, we now make a HEAD request to check if the file is still available. If the request returns a 404 Not Found, we redirect the user to a frontend page explaining that the ZIM file has expired and providing a button to re-request it.

Changes Made:

* Added a HEAD request to check ZIM file availability.
*  Redirect users to an expiration page if the file is missing.
* Improves user experience by making the expiration message more visible.

Motivation:

Previously, users could miss the expiration message because it was based only on timestamps. This change ensures that users are clearly informed when a ZIM file is no longer available.

How It Works:

Before displaying the expired message, the system now sends a HEAD request to Wasabi storage.

If the file exists, everything continues as usual.

If the response is 404 (Not Found), the user is redirected to a page that explains the expiration and provides a "Re-request ZIM" button.

Testing:

* Manually tested with both available and expired ZIM files.
* Checked for proper redirection when files are missing.

Notes:

I'm a beginner in open-source and trying to contribute to good projects. I'm excited to learn and grow, and I hope to find great mentors to support my journey. Feedback is welcome!

### DIFF
--- a/wp1/zimfarm.py
+++ b/wp1/zimfarm.py
@@ -281,31 +281,39 @@ def is_zim_file_ready(task_id):
 
 def zim_file_url_for_task_id(task_id):
   data = _get_task_by_id(task_id)
-
   if data is None:
-    return None
+        return None
 
   files = data.get('files', {})
   name = None
   for _, value in files.items():
-    name = value.get('name')
-    break
+        name = value.get('name')
+        break
 
   if name is None:
-    raise ZimFarmError('Could not find filename for ZIM file, task_id = %s' %
-                       task_id)
+        raise ZimFarmError(f'Could not find filename for ZIM file, task_id = {task_id}')
 
   warehouse_path = data.get('config', {}).get('warehouse_path')
   if warehouse_path is None:
-    raise ZimFarmError(
-        'Could not get warehouse path for ZIM file, task_id = %s' % task_id)
+        raise ZimFarmError(f'Could not get warehouse path for ZIM file, task_id = {task_id}')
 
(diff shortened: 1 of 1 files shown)


audiodude (changes requested):
Thanks for the PR!

This is the right basic idea, but I think it's in the wrong place. I would probably prefer to put this logic here:

https://github.com/openzim/wp1/blob/main/wp1/web/builders.py#L227

Additionally, a full end to end PR for this issue would include a frontend page that explains the issue to the user and provides a button to re-request the ZIM. Oddly, you mention in your PR description that you included that, but I don't see any commits. Did you forget to push them?

Finally, we will need tests for any frontend and backend code.

Salmathmullali:
Thanks for your fast response, I appreciate your guidance and will move the logic to wp1/web/builders.py as suggested. I also realize my mistake in mentioning the frontend page , I haven’t implemented it yet but can work on it if needed.

I'm still new to open source and trying my best to learn. Your feedback has been really helpful, and I truly appreciate it. Even though I try to figure things out on my own, having a mentor like you would make a big difference in my growth. I’ll carefully follow your suggestions and improve with each contribution. Thanks again for your patience and support.

audiodude:
Superseded by #839

Issue #832, opened 2025-04-02
Include Minio in docker-compose-dev.yml


audiodude:
In #796 we allowed the S3 URL to be configurable in credentials.py. As a follow up, it would be nice to include and configure a Minio instance in the dev docker-compose graph, with the credentials pre-populated in `credentials.py.dev.example`.

kelson42:
From what I have heard, using minio with MWoffliner is everything but straight forward.

audiodude:
Why is that so? I think the use cases between WP1 and MWoffliner might be different.

If it's too difficult, we could just set up a "mock S3" that takes the bytes and writes them to /dev/null.

kelson42:
> Why is that so? I think the use cases between WP1 and MWoffliner might be different.

I indeed assumed it was for MWoffliner. If I remember properly (user feedback is already a few years old), it was not that easy to save articles with long title on the fs backend (via Minio).

nicolapace:
I published a PR with something working: #834.

kelson42:
Does that mean that our instance of WP1 would rely on a local objct storage? If "yes", we need to make a proper risk assessment before merging/deploying anything.

nicolapace:
My proposed change is only for `docker-compose-dev.yml`, so only for development propose. No changes on production. If you think this might be interesting to explore for production as well we can also do that

kelson42:
Implemented by #834

PR #834, opened 2025-04-02
#832: Include Minio in docker-compose-dev.yml

nicolapace:
I added two new containers in docker-compose-dev.yml, one to run MinIO, and one to setup buckets in it.
I used the default docker images minio/minio to run MinIO and minio/mc to setup the buckets (with a script to do that in a clean way).
I create only a org-kiwix-dev-wp1 bucket in the script. 
On [localhost:9001](http://localhost:9001) there's also the MinIO web console available to easily check uploaded files in buckets.

I also added default credential for MinIO in credentials.* and updated comments on those files.

I tested it locally and it works: I am able to download the .tsv files.

Solves #832, and will make the setup for development of WP1 a bit easier. :)

### DIFF
--- a/wp1/credentials.py.dev.example
+++ b/wp1/credentials.py.dev.example
@@ -67,22 +67,24 @@ CREDENTIALS = {
         'CLIENT_URL': {
             'domains': ['http://localhost:5173'],
             'homepage': 'http://localhost:5173/#/',
+            's3': 'http://localhost:9000/org-kiwix-dev-wp1',
             'api': 'http://localhost:5000',
         },
 
-        # Configuration for the storage backend for storing selection lists. You MUST
-        # edit this line for your selections to be stored in a S3-like storage backend.
-        # You have two options for this setting:
+        # Configuration for the storage backend for storing selection lists.
+        # For development, you can use the local MinIO instance that's included in 
+        # the docker-compose setup (settings below are pre-configured for this).
+        # If you need to use an external service instead, you have two options:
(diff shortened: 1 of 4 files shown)


wp1/credentials.py.dev.example
```diff
@@ -67,22 +67,24 @@ CREDENTIALS = {
         'CLIENT_URL': {
             'domains': ['http://localhost:5173'],
             'homepage': 'http://localhost:5173/#/',
+            's3': 'http://localhost:9000/org-kiwix-dev-wp1',
```
audiodude, line 70: I don't think this is necessary. Where do you see this key being read?

wp1/credentials.py.dev.example
```diff
@@ -67,22 +67,24 @@ CREDENTIALS = {
...
             'api': 'http://localhost:5000',
         },
 
-        # Configuration for the storage backend for storing selection lists. You MUST
-        # edit this line for your selections to be stored in a S3-like storage backend.
-        # You have two options for this setting:
+        # Configuration for the storage backend for storing selection lists.
```
audiodude, line 74: Great comment!

audiodude (approved):
Thanks so much, this is fantastic! Just one little nit, but otherwise this is ready to go.

wp1/credentials.py.dev.example
nicolapace: It is not necessary in fact, probably an oversight.

nicolapace: Removing in a commit

Issue #736, opened 2024-05-08
Fix utcnow deprecation warning


audiodude:
```
/usr/local/lib/python3.12/site-packages/rq/utils.py:229: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
```

Check if `rq` has a fix for this.

nicolapace:
Proposed a fix in : #837

kelson42:
Closed by #837

PR #837, opened 2025-04-03
fix: update timestamp function to return UTC time and upgrade rq dependencies solving deprecation warnings

nicolapace:
Solves #736

I updated `utcnow()` function in `timestamp.py` to remove deprecation warnings using the suggested alternative.
I researched a way to remove the deprecation warning for the same thing coming from `rq`. It seems to be fixed [here](http://github.com/rq/rq/issues/2044). But the fix is not available in version 1.16.2, so I just updated `rq` to latest version (2.2.0).
I also updated `rq-scheduler` to the compatible version (0.14.0 which hase support for `rq` > 2.0) (as per their [CHANGELOG.md](https://github.com/rq/rq-scheduler/blob/master/CHANGELOG.md)).
I also re-build the Pipfile.lock file.

Tested and found no compatibility issues and no more deprecation warnings.


audiodude:
Thank you so much! I've been too lazy to address this to be honest. I'm a little worried about jumping a major version of rq. Did you test generating selections? Like creating a selection and saving it, and then making sure you get a TSV link?

nicolapace:
Yes, I also tested creating selections and I correctly got the `.tsv` files.

PR #841, opened 2025-04-04
feat: enhance WikiProject builder with autocomplete for project selection

nicolapace:
Fixes #775

I tried implementing something like what is proposed in #775. I reused the Autocomplete component from the Project home page and did some modifications to make it work correctly (especially to get errors and so on).
I used bootstrap for the style.
I modified the unit tests (although I was not able to test them all correctly cause the Autocomplete is not working properly when I run tests locally (but it works in my local deployment)). I will for sure continue working on the unit tests part :)

Hope to have your feedback. :)

Here's a screenshot of the modified page:

Screenshot

PR #843, opened 2025-04-05
fix: update VITE_API_URL to use 127.0.0.1 instead of localhost in .env.development

nicolapace:
As described in #841, unit tests are not working properly in the local development environment. 

In particular, I was not able to use the autocomplete component properly during test with cypress. 
Debugging I found out that inside cypress test environment I was not able to reach `localhost:5000/v1`.

Changing from the `VITE_API_URL` to use `127.0.0.1` instead of `localhost` in file `.env.development` solves the issue.

### DIFF
--- a/wp1-frontend/.env.development
+++ b/wp1-frontend/.env.development
@@ -1,1 +1,1 @@
-VITE_API_URL=http://localhost:5000/v1
+VITE_API_URL=http://127.0.0.1:5000/v1


audiodude (approved):

audiodude:
Again, I will state that this has never been an issue for me and is probably related to the network stack on your development computer. However, given that it's essentially a no-op change, it's fine.

PR #852, opened 2025-04-13
fix(logs): add Automatically generated template to log pages

hao7babe:
Add {{Automatically generated}} template to log pages to prevent manual renaming and editing.


audiodude (approved):

audiodude (changes requested):
Please update failing tests.

Issue #651, opened 2023-07-20
Alter timeout for requests to Petscan so we can support long queries


audiodude:
Discussed in #643.

Although `requests` doesn't have a timeout, and apparently Petscan doesn't either (or it's very long), an intermediate reverse proxy or other middleware might timeout our request if we make a very large one.

We should set a timeout on the request when we make it.

audiodude:
@rgaudin Did I get this right?

rgaudin:
Yep

audiodude:
I don't believe this is necessary. From my experience, the Petscan server is configured with an extremely long, or possibly nonexistent timeout, specifically because the developer of the service knows that complex queries can take longer than 60 seconds.

See for example: https://petscan.wmcloud.org/?labels_yes=&wikidata_label_language=&cb_labels_any_l=1&output_limit=&ores_prob_from=&min_redlink_count=1&referrer_name=&max_sitelink_count=&larger=&langs_labels_no=&ns%5B0%5D=1&project=wikipedia&common_wiki_other=&sitelinks_no=&templates_yes=&show_soft_redirects=both&interface_language=en&subpage_filter=either&depth=100&before=&search_max_results=500&outlinks_yes=&source_combination=&sortby=none&negcats=&ores_prob_to=&common_wiki=auto&links_to_all=&format=html&after=&edits%5Bflagged%5D=both&langs_labels_any=&sitelinks_any=&langs_labels_yes=&cb_labels_yes_l=1&categories=Biography+of+Living+Persons&language=en&min_sitelink_count=&cb_labels_no_l=1&search_query=&search_wiki=&search_filter=&doit=

Which runs for:

```
Query took 81.013 seconds.
```

Try creating a Petscan selection with this url: https://petscan.wmcloud.org/?psid=33735380

It runs for a minute or two and then produces a valid selection/tsv with all the rows.

PR #853, opened 2025-04-13
fix(selection): add timeout to Petscan Builder requests

hao7babe:
- Set request timeout to 45 seconds to keep timeout under nginx's 60s default limit to prevent 502 errors
- Add descriptive error message for timeout cases
- Fix: #651


audiodude:
https://github.com/openzim/wp1/issues/651#issuecomment-2799846185

Apologies that the issue was still open!

PR #854, opened 2025-04-14
chore: update dev db with new migrations

nicolapace:
The dev db was not up to date with the latest migrations, resulting in problems with the latest changes.

To fix the dev. env, I applied the following migrations that were missing:
 - [20250323_01_xyzAB-add-temp-pageviews-table]
 - [20250324_02_arj6b-add-ps-article-index-to-page-scores]
 - [20250330_01_zaRQa-migrate-last-7-days-of-logs]
 - [20250325_01_sFddz-add-article-count-to-selection-table]

### DIFF
--- a/docker/dev-db/enwp10_dev.dump.sql
+++ b/docker/dev-db/enwp10_dev.dump.sql
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:b7b2e1458f1c7ec397f0885e3823a5f86b9b3b190f807352cae9252e0b277c24
-size 21920793
+oid sha256:4716cbb9650d8e547df7f990dd1bb775f3985eda08a6f49f3c1cd58ec467b2ba
+size 31105032


audiodude (approved):
Just want to make sure, you followed the instructions here right? https://github.com/openzim/wp1/blob/main/docker/dev-db/README.md

Especially the part about destroying the dev db and getting a 'clean' copy before running the migrations?

I don't want to end up with user data in the dev db by mistake.

Thanks!

nicolapace:
Yes, i followed exactly that guide, delated the db container, rebuilding and then applying changes.

I also double checked the dump and everything seem to be ok. 

(I see my laptop name and OS-username mentioned in the `_yoyo_log` table but I suppose this is expected)

audiodude:
> (I see my laptop name and OS-username mentioned in the `_yoyo_log` table but I suppose this is expected)

Awesome, thanks for the confirmation!

Yes I don't know if there's a way to keep that information out, but I believe it's inconsequential. If you'd like, you can manually edit that table definition to redact it. Just make sure to test loading the dev-db image "from scratch" (`--build`) before you check in a new version. If it really becomes an issue, we could write an `awk` script or something that redacts it automatically as part of the update process.

PR #855, opened 2025-04-15
feat: add title field and metadata handling for ZIM file creation

### LINKED ISSUE #677: Limit length of ZIM file `zimTitle` field
Currently, the `customZimTitle` field that we send to the zimfarm is copied directly from the Builder name, which has no restrictions on its length, besides being truncated to 255 characters in the database. ZIM titles, on the other hand, can only be 30 characters max.

A couple of options:

1. Inherit the ZIM restriction and limit Builder titles to 30 characters
1. Truncate the Builder title when creating a ZIM

I tend to prefer the second option, because I don't see why the limitations of a ZIM should limit the entire WP1 UI.

nicolapace:
Fixes #677 

I added a `Title` field in the Zim creation page. This field is automatically filled with the `Selection Title` reduced to 30 characters. In this way users can change it if they want. This field is required and is of max 30 chars. (I also added a char counter for both the title and the description).

I also added backend validation for the size of zim metadata (title (30) and description (80) as per https://wiki.openzim.org/wiki/Metadata).

I added new unit test for the new "title field" and the validation, for both the backend and the frontend. I also fixed some unit tests and functions to support the new `zimtitle` field.

Here is the Zim creation page with the new field:

Screenshot

### DIFF
--- a/wp1-frontend/src/components/ZimFile.vue
+++ b/wp1-frontend/src/components/ZimFile.vue
@@ -109,7 +109,7 @@
                 v-on:input="validateInput('longDescription', maxLongDescriptionLength)"
                 rows="6"
                 class="form-control"
-                :class="{'is-invalid': isLongDescriptionInvalid}"
+                :class="{'is-invalid': !isLongDescriptionValid}"
                 placeholder="ZIM file created from a WP1 Selection"
               ></textarea>
               <small class="form-text" :class="{'text-muted': graphemeCount(longDescription) < maxLongDescriptionLength, 'text-warning': graphemeCount(longDescription) === maxLongDescriptionLength}">
@@ -345,15 +345,15 @@ export default {
         this.status === 'ENDED'
       );
     },
-    isLongDescriptionInvalid: function() {
-      if (!this.longDescription || !this.description) return false;
+    isLongDescriptionValid: function() {
+      if (!this.longDescription || !this.description) return true;
       const longCount = this.graphemeCount(this.longDescription);
       const descCount = this.graphemeCount(this.description);
-      return longCount < descCount || this.longDescription === this.description;
+      return longCount >= descCount && this.longDescription !== this.description;
     },
     hasLengthErrors: function() {
       // Prevent empty required fields and invalid long description
-      return !this.zimTitle || !this.description || this.isLongDescriptionInvalid;
+      return !this.zimTitle || !this.description || !this.isLongDescriptionValid;
     }
   },
   watch: {
--- a/wp1-frontend/src/components/ZimFile.vue
+++ b/wp1-frontend/src/components/ZimFile.vue
@@ -20,7 +20,7 @@
           <h2>Create ZIM file</h2>
           <p v-if="status === 'NOT_REQUESTED'">
             Use this form to create a ZIM file from your selection, so that you
-            can browse the articles it contains offline. The
+            can browse the articles it contains offline. The &quot;Title&quot;,
             &quot;Description&quot; and &quot;Long Description&quot; fields are
             required, but generic defaults will be used if they're not provided.
           </p>
@@ -66,6 +66,21 @@
             class="needs-validation"
             novalidate
           >
+            <div id="zimtitle-group" ref="zimtitle_form_group" class="form-group">
+              <label for="zimtitle">Title</label>
+              <input
+                id="zimtitle"
+                ref="zimtitle"
+                v-on:blur="validationOnBlur"
+                v-model="zimTitle"
+                class="form-control"
+                maxlength="30"
+                placeholder="ZIM title"
+                required
+              />
+              <small class="form-text text-muted">{{ zimTitle.length }}/30 characters</small>
+              <div class="invalid-feedback">Please provide a title</div>
+            </div>
             <div id="desc-group" ref="form_group" class="form-group">
               <label for="desc">Description</label>
               <input
@@ -78,6 +93,7 @@
                 placeholder="ZIM file created from a WP1 Selection"
                 required
               />
+              <small class="form-text text-muted">{{ description.length }}/80 characters</small>
               <div class="invalid-feedback">Please provide a description</div>
             </div>
             <div class="form-group">
@@ -160,6 +176,7 @@ export default {
   components: { SecondaryNav, LoginRequired, PulseLoader },
   data: function () {
     return {
+      zimTitle: '',
       description: '',
       errors: [],
       errorMessages: [],
@@ -198,6 +215,10 @@ export default {
       } else {
         this.notFound = false;
         this.builder = await response.json();
+        // Set default ZIM title from builder name, truncated to 30 chars if necessary
+        if (this.builder && this.builder.name) {
+          this.zimTitle = this.builder.name.substring(0, 30);
+        }
         this.$emit('onBuilderLoaded', this.builder);
       }
     },
@@ -214,8 +235,6 @@ export default {
 
       const data = await response.json();
       this.status = data.status;
-      this.description = data.description;
-      this.longDescription = data.long_description;
       this.isDeleted = data.is_deleted;
       if (this.status === 'FILE_READY') {
         this.stopProgressPolling();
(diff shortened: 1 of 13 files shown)


audiodude:
> @nicolapace Thank you for your PR! But the unity for counting is not char, but lexem. There is a javascript/typescript implementation of such a counter in openzim/mwoffliner.

https://github.com/openzim/wp1/issues/677#issuecomment-2807939549

What is a lexem and how does it compare to a grapheme? I feel like I'm going to have to read a Chomsky book to implement this PR. :upside_down_face:

audiodude (changes requested):
This looks really good, great work!

I think the handling of the error conditions on the backend needs to be improved however.

kelson42:
@audiodude I meant grapheme, sorry.

wp1/zimfarm.py
```diff
@@ -121,7 +125,23 @@ def get_webhook_url():
                                                  urllib.parse.quote(token))
 
 
-def _get_params(s3, wp10db, builder, description='', long_description=''):
+def validate_zim_metadata(title=None, description=None):
+  """Validate ZIM metadata fields against length limits."""
+  errors = []
+  print(f"Validating ZIM metadata: title='{title}', description='{description}'")
```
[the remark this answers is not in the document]
nicolapace: Sure :)

nicolapace:
@audiodude @kelson42 @benoit74 Thanks for the feedback and the suggestions!

I pushed a new commit that addresses the discussed issues, in particular:
 - I am now using `split-by-grapheme` npm package and `regex` python module to check for num. of graphemes in both frontend and backend as suggested by @kelson42 and @benoit74
 - as suggested by @audiodude I changed the `validate_zim_metadata()` function to Raise exceptions directly and I created new exceptions
 - I also added validation for the `long_description` field as described in https://wiki.openzim.org/wiki/Metadata#Graphemes
 - I disable the submit button if inputs are not valid
 - I changed unit tests to support the new changes and added some new unit tests (for the long description validation and the handling of graphemes)

(I pushed a new commit instead of amending the last one for better clarity. We can squash them later if you prefer.)

wp1/zimfarm.py
```diff
@@ -121,7 +125,23 @@ def get_webhook_url():
...
+  if title and len(title) > ZIM_TITLE_MAX_LENGTH:
+    errors.append(f"Title exceeds maximum length: {ZIM_TITLE_MAX_LENGTH} characters.")
+  
+  if description and len(description) > ZIM_DESCRIPTION_MAX_LENGTH:
+    errors.append(f"Description exceeds maximum length: {ZIM_DESCRIPTION_MAX_LENGTH} characters.")
+    
+  if errors:
+    return False, " ".join(errors)
```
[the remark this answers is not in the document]
nicolapace: Thanks! And yes, totally agree, makes sense. We handle user feedback on the frontend and just return errors on the backend calls. I’ve switched to raising specific exceptions like `InvalidZimTitleError` and `InvalidZimDescriptionError` (as subclasses of `ZimFarmError`) instead of using status tuples. (I've also added `InvalidZimLongDescriptionError` for `long_description` related errors). Thanks again for the feedback!

wp1-frontend/src/components/ZimFile.vue
```diff
@@ -72,13 +72,15 @@
...
+                v-on:input="validateInput('zimTitle', maxTitleLength)"
                 v-model="zimTitle"
                 class="form-control"
-                maxlength="30"
                 placeholder="ZIM title"
                 required
               />
-              <small class="form-text text-muted">{{ zimTitle.length }}/30 characters</small>
+              <small class="form-text" :class="{'text-muted': graphemeCount(zimTitle) < maxTitleLength, 'text-warning': graphemeCount(zimTitle) === maxTitleLength}">
+                {{ graphemeCount(zimTitle) }}/{{ maxTitleLength }} graphemes{{ graphemeCount(zimTitle) === maxTitleLength ? ' (max reached)' : '' }}
```
audiodude, line 82: This is a lot of information/processing to be inside of a {{ ... }} block. Please refactor to a helper function. Specifically the last part with the ternary operator. I would  a full on `displayGraphemeLimitText()` helper.

wp1-frontend/src/components/ZimFile.vue
```diff
@@ -87,13 +89,15 @@
...
+                v-on:input="validateInput('description', maxDescriptionLength)"
                 v-model="description"
                 class="form-control"
-                maxlength="80"
                 placeholder="ZIM file created from a WP1 Selection"
                 required
               />
-              <small class="form-text text-muted">{{ description.length }}/80 characters</small>
+              <small class="form-text" :class="{'text-muted': graphemeCount(description) < maxDescriptionLength, 'text-warning': graphemeCount(description) === maxDescriptionLength}">
+                {{ graphemeCount(description) }}/{{ maxDescriptionLength }} graphemes{{ graphemeCount(description) === maxDescriptionLength ? ' (max reached)' : '' }}
```
audiodude, line 99: Same feedback here.

wp1-frontend/src/components/ZimFile.vue
```diff
@@ -314,6 +341,16 @@ export default {
         this.status === 'ENDED'
       );
     },
+    isLongDescriptionInvalid: function() {
```
audiodude, line 344: nit: prefer booleans that test for positive attributes. So `isLongDescriptionValid()` and then use `!isLongDescriptionValid()` in your test

wp1/zimfarm_test.py
```diff
@@ -408,11 +408,27 @@ def test_schedule_zim_file_task_create_raises(self, get_params_mock,
...
       actual = zimfarm.schedule_zim_file(s3, redis, self.wp10db, self.builder)
 
+  @patch('wp1.zimfarm.requests')
+  @patch('wp1.zimfarm.get_zimfarm_token')
+  @patch('wp1.zimfarm._get_params')
+  def test_schedule_zim_valid_graphemes(self, get_params_mock, get_token_mock,
+                             mock_requests):
+    redis = MagicMock()
+    s3 = MagicMock()
+    get_params_mock.return_value = {'name': 'bar'}
```
audiodude, line 418: Optional: I don't think you need to set up all the response values for the mocks here. We just need to send in any mock as a requirement for the function, so that it gets to the point where it can raise the exception.

I recognize that other `test_invalid_foo` methods in this test do that, so it would be optional but nice to clean it up, especially as we are adding a few more methods of this type. Alternately you could simply fix it in the new methods and ignore the existing calls.

wp1/zimfarm_test.py
```diff
@@ -423,7 +439,24 @@ def test_schedule_zim_file_long_name(self, get_params_mock, get_token_mock,
...
+        zimfarm.schedule_zim_file(s3, redis, self.wp10db, self.builder, title=wrong_title)
+
+  @patch('wp1.zimfarm.requests')
+  @patch('wp1.zimfarm.get_zimfarm_token')
+  @patch('wp1.zimfarm._get_params')
+  def test_schedule_zim_file_too_long_title(self, get_params_mock, get_token_mock,
+                             mock_requests):
+    redis = MagicMock()
+    s3 = MagicMock()
+    get_params_mock.return_value = {'name': 'bar'}
```
audiodude, line 452: Same comment here.

wp1/zimfarm_test.py
```diff
@@ -439,12 +472,60 @@ def test_schedule_zim_file_too_long_description(self, get_params_mock, get_token
...
+
+
+  @patch('wp1.zimfarm.requests')
+  @patch('wp1.zimfarm.get_zimfarm_token')
+  @patch('wp1.zimfarm._get_params')
+  def test_schedule_zim_file_too_long_long_description(self, get_params_mock, get_token_mock,
+                             mock_requests):
+    redis = MagicMock()
+    s3 = MagicMock()
+    get_params_mock.return_value = {'name': 'bar'}
```
audiodude, line 487: Same

audiodude (changes requested):
Looking good, just a few small comments.

audiodude (approved):
Great work!

PR #860, opened 2025-04-20
Limit the number of articles that can be made into a ZIM file

### LINKED ISSUE #816: Limit the size of the selections we can ZIM through WP1

audiodude:
Fixes #816 

This PR includes both backend limits as well as frontend validation and messaging. The article limit is configured in a single place and passed to the frontend in the same request that provides the number of articles to the ZIM request page. Even if this request is tampered with, or if a request is made through the WP1 API, the backend zimfarm module checks the limits before requesting a ZIM.

### DIFF
--- a/wp1-frontend/cypress/e2e/zimFile.cy.js
+++ b/wp1-frontend/cypress/e2e/zimFile.cy.js
@@ -4,7 +4,7 @@ describe('the zim file creation page', () => {
   describe('when the user is logged in', () => {
     beforeEach(() => {
       cy.intercept('v1/oauth/identify', { fixture: 'identity.json' }).as(
-        'identity'
+        'identity',
       );
     });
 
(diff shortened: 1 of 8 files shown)


audiodude:
image

PR #865, opened 2025-04-26
Delete logging table

### LINKED ISSUE #828: TODO: Drop logging table and remove references in code
After we deploy #823 and #827, we should wait a week to make sure logging tables are being processed properly.

After that, we can drop the logging table from the database and clean up references in the code. This should be done as a yoyo migration.

audiodude:
Fixes #828 

We have simply added a migration to drop the logging table, and updated the test schemas.

This will save 6 GB in the database and the corresponding amount for every database backup:

```
MariaDB [enwp10_prod]> SELECT      table_name AS `Table`,      round(((data_length + index_length) / 1024 / 1024), 2) `Size in MB`  FROM information_schema.TABLES  WHERE table_schema = "enwp10_prod"
   AND table_name = "logging";
+---------+------------+
| Table   | Size in MB |
+---------+------------+
| logging |    5929.91 |
+---------+------------+
1 row in set (0.020 sec)
```

The `Log` model is retained, because it is still used in the Redis version of the code.

I have confirmed that logs have been successfully running off Redis for several weeks now.

Thank you again @elfkuzco!

### DIFF
--- a/db/migrations/20250426_01_i656U-drop-logging-table.py
+++ b/db/migrations/20250426_01_i656U-drop-logging-table.py
@@ -1,0 +1,12 @@
+"""
+Drop logging table
+"""
+
+from yoyo import step
+
+__depends__ = {
+    '20250325_01_sFddz-add-article-count-to-selection-table',
+    '20250330_01_zaRQa-migrate-last-7-days-of-logs'
+}
+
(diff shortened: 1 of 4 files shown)


elfkuzco:
You are welcome @audiodude . I am glad the solution is bring about savings in size as expected.

PR #894, opened 2025-06-08
Allow mwoffliner image to be configurable, and use ActionParse renderer

### LINKED ISSUE #891: Use released version of mwoffliner
Currently, tasks are requested with the `latest` version of mwoffliner: https://github.com/openzim/wp1/blob/b69e12ce9b194e87db4257044a15164bff5def07/wp1/zimfarm.py#L23

This is not a good design since it means we have little control on which version of the scraper is used for a given task.

And it could have bad consequences should there be a significant change in terms of scraper parameters / behavior.

We should rather use a released image, e.g. `ghcr.io/openzim/mwoffliner:1.15.0` today.

Once done, we will update mwoffliner release todo list to include an item about wp1 update.

### LINKED ISSUE #892: Move to ActionParse renderer
Since 1.15, mwoffliner scraper supports the `ActionParse` renderer. This renderer is expected to become the default renderer in the coming months, but it is not yet the case since we are uncertain about its stability.

For WP1, we however want to use this new renderer since scraper usage is under our control and we expect results to be significantly better.

We hence need to pass an additional flag to tasks we create:
- name: `forceRender`
- value: `ActionParse`

https://github.com/openzim/wp1/blob/b69e12ce9b194e87db4257044a15164bff5def07/wp1/zimfarm.py#L197-L212

audiodude:
Fixes #891 
Fixes #892

### DIFF
--- a/wp1/zimfarm.py
+++ b/wp1/zimfarm.py
@@ -9,10 +9,14 @@ import wp1.logic.builder as logic_builder
 import wp1.logic.selection as logic_selection
 from wp1.constants import WP1_USER_AGENT
 from wp1.credentials import CREDENTIALS, ENV
-from wp1.exceptions import (InvalidZimDescriptionError,
-                            InvalidZimLongDescriptionError,
-                            InvalidZimTitleError, ObjectNotFoundError,
-                            ZimFarmError, ZimFarmTooManyArticlesError)
+from wp1.exceptions import (
+    InvalidZimDescriptionError,
+    InvalidZimLongDescriptionError,
+    InvalidZimTitleError,
+    ObjectNotFoundError,
+    ZimFarmError,
+    ZimFarmTooManyArticlesError,
+)
 from wp1.logic import util
 from wp1.time import get_current_datetime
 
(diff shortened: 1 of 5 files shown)


wp1/zimfarm.py
```diff
@@ -184,12 +179,19 @@ def _get_params(s3,
   filename_prefix = '%s-%s' % (util.safe_name(
       builder.b_name.decode('utf-8')), selection_id_frag)
 
+  image = CREDENTIALS[ENV].get('ZIMFARM', {}).get('iamge')
```
benoit74, line 182: Typo in `iamge`

wp1/zimfarm.py
```diff
@@ -184,12 +179,19 @@ def _get_params(s3,
...
+  if image is None:
+    image = 'ghcr.io/openzim/mwoffliner:1.15.0'
```
benoit74, line 184: Should default rather be `latest`, just like today? I feel it odd to specify 1.15.0, which will get outdated in few months

benoit74 (approved):
Mostly OK, good to merge once you've read / taken into account my comments.

PR #897, opened 2025-06-16
WIP: Add ZIM file scheduling functionality and update builder logic

nicolapace:
- Introduced a new worker for ZIM file scheduling in supervisord configuration.
- Refactored the request_zim_file function to handle ZIM file creation requests.
- Enhanced schedule_zim_file function to support scheduled repetitions.
- Updated the create_zim_file_for_builder endpoint to accept scheduled repetitions.
- Added necessary queue management for ZIM file scheduling in queues.py.

TODO:
- run create_or_update_builder for scheduled job
- add db queries
- handle zim creations using hook (+ notify)
- other CRUD api endpoints
- frontend

### DIFF
--- a/supervisord-dev.conf
+++ b/supervisord-dev.conf
@@ -68,6 +68,30 @@ stopsignal=TERM
 autostart=true
 autorestart=true
 
+[program:wp1-zimfile-scheduling]
+; In the docker-compose world, the redis host is just 'redis'
+command=/usr/local/bin/rq worker -u redis://redis zimfile-scheduling
+; process_num is required if you specify >1 numprocs
+process_name=zimfile-scheduling-%(process_num)s
+
+; If you want to run more than one zimfile-scheduling worker, increase this
+numprocs=1
+
+; This is the directory from which RQ is run. Be sure to point this to the
+; directory where your source code is importable from
+directory=/usr/src/app
+
+redirect_stderr=true
+stdout_logfile=/var/log/wp1bot/%(program_name)s-%(process_num)s.log
+stdout_logfile_maxbytes=100MB
+stdout_logfile_backups=5
+
+; RQ requires the TERM signal to perform a warm shutdown. If RQ does not die
+; within 10 seconds, supervisor will forcefully kill it
+stopsignal=TERM
+autostart=true
+autorestart=true
+
 [program:scheduler]
 ; In the docker-compose world, the redis host is just 'redis'
 command=/usr/local/bin/rqscheduler --host redis -i 20
--- a/wp1/logic/builder.py
+++ b/wp1/logic/builder.py
@@ -385,28 +385,25 @@ def latest_selections_with_errors(wp10db, builder_id):
   return res
 
 
-def schedule_zim_file(s3,
-                      redis,
-                      wp10db,
-                      builder_id,
-                      user_id=None,
-                      title='',
-                      description='',
-                      long_description=None):
-  if isinstance(builder_id, str):
-    builder_id = builder_id.encode('utf-8')
-  builder = get_builder(wp10db, builder_id)
-  if builder is None:
-    raise ObjectNotFoundError('Could not find builder with id = %s' %
-                              builder_id)
+def request_zim_file( builder, builder_id, title, description, long_description, rebuild_selection=False):
+  """Request a ZIM file creation for the given builder.
+  """
+  # logger.error("Inputs: builder_id=%s, title=%s, description=%s, long_description=%s", builder_id, title, description, long_description)
 
-  if user_id is not None:
-    user_id = str(user_id)
-    builder_user_id = builder.b_user_id.decode('utf-8')
-    if user_id != builder_user_id:
-      raise UserNotAuthorizedError(
-          'Could not use builder id = %s for user id = %s' %
-          (builder_id, user_id))
+  wp10db = wp10_connect()
+  redis = redis_connect()
+  s3 = connect_storage()
+
+  if rebuild_selection:
+    pass
+    ## WIP 
+    # create_or_update_builder(wp10db,
+    #                         list_name,
+    #                         user_id,
+    #                         project,
+    #                         params,
+    #                         model,
+    #                         builder_id=builder_id)
 
   task_id = zimfarm.schedule_zim_file(s3,
                                       redis,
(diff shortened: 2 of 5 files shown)


wp1/queues.py
```diff
@@ -182,3 +188,39 @@ def poll_for_zim_file_status(redis, task_id):
...
+  scheduler = Scheduler(connection=queue.connection, queue=queue)
+
+  period_months = scheduled_repetitions['repetition_period_in_months']
+  num_repetitions = scheduled_repetitions['number_of_repetitions']
+  
+  first_run = datetime.now(UTC) + relativedelta(months=period_months)
+
+  cron_string = f"{first_run.minute} {first_run.hour} {first_run.day} */{period_months} *"
+
+  job = scheduler.cron(
```
audiodude, line 216: Use scheduler.schedule instead of cron

audiodude:
@kelson42 This is the summer of code project.

kelson42:
@nicolapace @audiodude I still would prefer that things run normaly, ie. :
* Problem/improvement explain in issue
* Solution approach agreed in issue comments
* Solution code proposed and explained in PR

All of this done in a way that we don't have massive/global issues and PR.

db/migrations/20250621_01_A5B6C-add-interval-and-remaining-gens-to-zim-schedules.py
```diff
@@ -0,0 +1,14 @@
...
+This migration adds 'interval_between_zim_generations' and 'remaining_generations' to the zim_schedules table.
+"""
+
+from yoyo import step
+
+__depends__ = {'20250610_02_FHB8A-create-zim-schedules-table'}
+
+steps = [
+    step(
+        'ALTER TABLE zim_schedules ADD COLUMN s_interval_between_zim_generations INTEGER NULL, ADD COLUMN s_remaining_generations INTEGER NULL',
```
audiodude, line 11: Let's just call this `s_interval`.

wp1/logic/builder.py
```diff
@@ -385,16 +387,58 @@ def latest_selections_with_errors(wp10db, builder_id):
...
+
+  wp10db = wp10_connect()
+  redis = redis_connect()
+  s3 = connect_storage()
+
+  if rebuild_selection:
+    logger.error('Rebuilding selection for builder_id=%s', builder_id)
+
+    builder : Builder = get_builder(wp10db, builder_id)
+    update_builder(wp10db, builder)
```
audiodude, line 403: Update builder saves the passed builder to the database. But you just loaded it from the database, right? Why do you need to save it?

wp1/web/builders.py
audiodude, line 1: Let's save this for another PR.

wp1/logic/zim_schedules.py
```diff
@@ -0,0 +1,101 @@
...
+  with wp10db.cursor() as cursor:
+    cursor.execute(
+      'SELECT * FROM zim_schedules WHERE s_builder_id = %s', (builder_id,)
+    )
+    rows = cursor.fetchall()
+  return [
+    ZimSchedule(**row) for row in rows
+  ]
+
+def decrement_remaining_generations(wp10db, schedule_id):
```
audiodude, line 66: Why do we need to keep track of the remaining generations? Shouldn't the `repeat` parameter of rq-scheduler handle this automatically?

wp1/queues.py
```diff
@@ -176,9 +184,66 @@ def enqueue_materialize(redis, builder_cls, builder_id, content_type):
...
+  if 'number_of_repetitions' not in scheduled_repetitions:
+    raise ValueError('scheduled_repetitions must contain number_of_repetitions')
+
+  queue = _get_zimfile_scheduling_queue(redis)
+  scheduler = Scheduler(connection=queue.connection, queue=queue)
+
+  period_months = scheduled_repetitions['repetition_period_in_months']
+  num_repetitions = scheduled_repetitions['number_of_repetitions']
+  
+  first_run = datetime.now(UTC) + relativedelta(months=period_months)
```
audiodude, line 222: Let's discuss the definition of a "month". Is it 30 days, or do we want the zim file to occur on the same "day" each month (Jan 4, Feb 4, Mar 4)?

wp1/logic/builder.py
```diff
@@ -385,16 +387,58 @@ def latest_selections_with_errors(wp10db, builder_id):
   return res
 
 
-def schedule_zim_file(s3,
+def request_zimfile_from_zimfarm( builder, builder_id, title, description, long_description, rebuild_selection=False):
+  """Request a ZIM file creation for the given builder.
+  """
+  # logger.error("Inputs: builder_id=%s, title=%s, description=%s, long_description=%s", builder_id, title, description, long_description)
+
+  wp10db = wp10_connect()
```
audiodude, line 395: This should only be done in "top level" tasks, so if this function is being called from other logic/web code, we shouldn't be doing this. I understand that it's necessary for the scheduled job.

supervisord-dev.conf
audiodude, line 1: Let's move the supervisor config changes to their own PR.

wp1/logic/builder.py
```diff
@@ -385,16 +387,58 @@ def latest_selections_with_errors(wp10db, builder_id):
   return res
 
 
-def schedule_zim_file(s3,
+def request_zimfile_from_zimfarm( builder, builder_id, title, description, long_description, rebuild_selection=False):
```
audiodude, line 390: Why do you need both `builder` and `builder_id`? Shouldn't you be able to get one from the other?

wp1/logic/builder.py
```diff
@@ -193,7 +195,7 @@ def auto_schedule_zim_file(s3, redis, wp10db, builder_id):
       'utf-8') if zim_file.z_description is not None else None
   long_description = zim_file.z_long_description.decode(
       'utf-8') if zim_file.z_long_description is not None else None
-  schedule_zim_file(s3,
+  request_zimfile_from_zimfarm(s3,
```
audiodude, line 198: This call is wrong. The first param to `request_zimfile_from_zimfarm` is `builder`, but here you're passing `s3`.

audiodude (changes requested):
Let's discuss at our meeting a better way of handling draft PRs in the future. On the one hand, I want to give you feedback as early as possible. On the other, I'm not sure which parts of the PR are ready for feedback versus ones you "know" are broken.

wp1/logic/builder.py
nicolapace: yes I can do that

wp1/logic/zim_schedules.py
nicolapace: to show something on the webpage/frontend (?)

wp1/logic/builder.py
nicolapace: ok then it's not needed, cause the materialize that it's done after will also update it.

supervisord-dev.conf
nicolapace: yes, and I will change both the dev and the non-dev one in there.

nicolapace:
This PR is a replica of smaller, fine-grained PRs for the GSoC 2025 project. Closing it.

PR #904, opened 2025-06-25
Update `zim_schedules` schema

### LINKED ISSUE #902: Correct the DB schema for `zimfile-schedules` tables
**Summary**
Update the schema of the `zim_schedules` table to support recurring schedule metadata and correct the datatype of the `s_rq_job_id` field.

**Proposed Change**

Add columns:

- `s_interval` `INTEGER NULL` – to store the interval (in days) between scheduled generations.
- `s_remaining_generations` `INTEGER NULL` – to track how many future runs remain.

Modify column:

- Change `s_rq_job_id` from `INTEGER` to `VARCHAR(36)` to support string-based UUIDs used by RQ.

The schema changes will be implemented changing the last YoYo migration script (to reduce the number of migrations script), and re-applied, and the dump will be updated accordingly.
Also `wp10_test.sql` will be updated.

### LINKED ISSUE #923: Update WP1 test db user table
add 
```
`u_email` varchar(255) DEFAULT NULL,
```
in `wp10_test.up.sql`

nicolapace:
The migration for the `zim_schedules` table was updated:

The `s_rq_job_id` column type was changed from INTEGER to VARCHAR(36) to better match job ID formats.
Two new columns were added:
- `s_interval` `INTEGER NULL` (interval between ZIM generations)
- `s_remaining_generations` `INTEGER NULL` (remaining generations to be created)

The corresponding SQL schema in `wp10_test.up.sql` and the dev database dump reference were updated to reflect these changes.

Fixes #902 
Fixes #923


wp10_test.up.sql
```diff
@@ -131,7 +131,9 @@ CREATE TABLE zim_schedules (
   s_id VARCHAR(36) NOT NULL PRIMARY KEY,
   s_builder_id VARCHAR(255) NOT NULL,
   s_zim_file_id INTEGER NULL,
-  s_rq_job_id INTEGER NOT NULL,
+  s_rq_job_id VARCHAR(36) NOT NULL,
```
audiodude, line 134: This should match.

PR #917, opened 2025-06-26
add email handling in OAuth complete flow and corresponding test

### LINKED ISSUE #916: Read user email from wikimedia account
- modify production Wikimedia OAuth consumer / create a new consumer that has access to user email.
- read the email and save/update in the database and session["user"]

nicolapace:
This PR enhances the OAuth `complete()` flow by ensuring that the user's email is properly retrieved and stored.

If the identity object from the OAuth provider includes an email, it is used directly.
If not, the email is fetched from the users table using the user ID.
The email is then stored in both the session and the users table.
Added also `test_complete_sets_email_in_session()` to verify that the email is correctly set in the session after a successful OAuth flow.

Fixes #916 (wikimedia oauth consumer needs to be updated/recreated with email permissions)

### DIFF
--- a/wp1/web/oauth.py
+++ b/wp1/web/oauth.py
@@ -44,22 +44,37 @@ def complete():
   session.pop('request_token')
   identity = handshaker.identify(access_token)
 
+  wp10db = get_db('wp10db')
+  
+  if identity.get('email') is None:
+    with wp10db.cursor() as cursor:
+      cursor.execute(
+          '''SELECT u_email FROM users WHERE u_id = %(u_id)s''',
+          {'u_id': identity['sub']}
+      )
+      result = cursor.fetchone()
+      email = result['u_email'] if result else None
+  else:
+    email = identity.get('email')
+
   session['user'] = {
       'access_token': access_token,
       'identity': {
           'username': identity['username'],
+          'email': email,
           'sub': identity['sub']
       }
   }
-  wp10db = get_db('wp10db')
+
   with wp10db.cursor() as cursor:
     cursor.execute(
-        '''INSERT INTO users (u_id, u_username)
-                      VALUES (%(u_id)s, %(u_username)s)
-                      ON DUPLICATE KEY UPDATE u_id= %(u_id)s, u_username= %(u_username)s''',
+        '''INSERT INTO users (u_id, u_username, u_email)
+                      VALUES (%(u_id)s, %(u_username)s, %(u_email)s)
+                      ON DUPLICATE KEY UPDATE u_id= %(u_id)s, u_username= %(u_username)s, u_email= %(u_email)s''',
         {
             'u_id': identity['sub'],
-            'u_username': identity['username']
+            'u_username': identity['username'],
+            'u_email': email
         })
     wp10db.commit()
   next_path = session.pop('next_path')
--- a/wp1/web/oauth.py
+++ b/wp1/web/oauth.py
@@ -61,7 +61,7 @@ def complete():
       VALUES (%(u_id)s, %(u_username)s, %(u_email)s)
       ON DUPLICATE KEY UPDATE
         u_username = VALUES(u_username),
-        u_email = COALESCE(VALUES(u_email), u_email)
+        u_email = IF(u_email IS NULL, VALUES(u_email), u_email)
       ''',
       {
         'u_id': identity['sub'],
--- a/wp1/web/oauth.py
+++ b/wp1/web/oauth.py
@@ -61,7 +61,7 @@ def complete():
       VALUES (%(u_id)s, %(u_username)s, %(u_email)s)
       ON DUPLICATE KEY UPDATE
         u_username = VALUES(u_username),
-        u_email = IF(u_email IS NULL, VALUES(u_email), u_email)
+        u_email = VALUES(u_email)
       ''',
       {
         'u_id': identity['sub'],
--- a/wp1/web/oauth.py
+++ b/wp1/web/oauth.py
@@ -45,37 +45,30 @@ def complete():
   identity = handshaker.identify(access_token)
 
   wp10db = get_db('wp10db')
-  
-  if identity.get('email') is None:
-    with wp10db.cursor() as cursor:
-      cursor.execute(
-          '''SELECT u_email FROM users WHERE u_id = %(u_id)s''',
-          {'u_id': identity['sub']}
-      )
-      result = cursor.fetchone()
-      email = result['u_email'] if result else None
-  else:
-    email = identity.get('email')
 
   session['user'] = {
       'access_token': access_token,
       'identity': {
           'username': identity['username'],
-          'email': email,
           'sub': identity['sub']
       }
   }
 
   with wp10db.cursor() as cursor:
     cursor.execute(
-        '''INSERT INTO users (u_id, u_username, u_email)
-                      VALUES (%(u_id)s, %(u_username)s, %(u_email)s)
-                      ON DUPLICATE KEY UPDATE u_id= %(u_id)s, u_username= %(u_username)s, u_email= %(u_email)s''',
-        {
-            'u_id': identity['sub'],
-            'u_username': identity['username'],
-            'u_email': email
-        })
+      '''
+      INSERT INTO users (u_id, u_username, u_email)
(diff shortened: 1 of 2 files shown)


wp1/web/oauth_test.py
```diff
@@ -168,3 +169,21 @@ def test_logout_unauthorized_user(self):
...
+  @patch('wp1.web.oauth.homepage_url',
+         TEST_OAUTH_CREDS[ENV]['CLIENT_URL']['homepage'])
+  @patch('wp1.web.oauth.handshaker', handshaker)
+  def test_complete_sets_email_in_session(self):
+    self.app = create_app()
+    with self.override_db(self.app), self.app.test_client() as client:
+      with client.session_transaction() as sess:
+        sess['request_token'] = self.REQUEST_TOKEN
+        sess['next_path'] = ''
+      rv = client.get('/v1/oauth/complete?query_string')
```
audiodude, line 184: Don't you have to put an actual legitimate query string here?

audiodude (changes requested):

nicolapace:
I rewrote so that I upsert in the DB the email, but don't save in session. CI will also fail cause this needs some changes in the testDB schema, but I tested locally.
I make sure not to update the email to a NULL value (with COALESCE) to prevent emails added manually by users on the zimfile generation page to be overwritten.

wp1/web/oauth.py
```diff
@@ -44,23 +44,31 @@ def complete():
...
-        {
-            'u_id': identity['sub'],
-            'u_username': identity['username']
-        })
+      '''
+      INSERT INTO users (u_id, u_username, u_email)
+      VALUES (%(u_id)s, %(u_username)s, %(u_email)s)
+      ON DUPLICATE KEY UPDATE
+        u_username = VALUES(u_username),
+        u_email = COALESCE(VALUES(u_email), u_email)
```
audiodude, line 64: If I read this right, it will prioritize the u_email sent in the VALUES clause, so the one returned from OAuth. Doesn't that mean that even if the user has saved a preferred email on the form, we will overwrite it with the email returned by OAuth? Or am I reading this wrong?

https://dev.mysql.com/doc/refman/8.4/en/comparison-operators.html#function_coalesce

wp1/web/oauth_test.py
```diff
@@ -92,6 +110,35 @@ def test_complete_unauthorized_user(self):
...
+      with self.app.app_context():
+        wp10db = get_db('wp10db')
+        with wp10db.cursor() as cursor:
+          cursor.execute('INSERT INTO users (u_id, u_username, u_email) VALUES (%s, %s, %s)', (self.USER_NO_EMAIL['identity']['sub'], 'overwrite', 'test@test.it'))
+          wp10db.commit()
+      with client.session_transaction() as sess:
+        sess['request_token'] = self.REQUEST_TOKEN
+        sess['next_path'] = ''
+      rv = client.get('/v1/oauth/complete?query_string')
+      self.assertEqual(
```
audiodude, line 130: This assert doesn't make any sense to me. We're asserting that a constant that we created solely for this test has a certain value?

wp1/web/oauth_test.py
```diff
@@ -92,6 +110,35 @@ def test_complete_unauthorized_user(self):
...
+  @patch('wp1.web.app.ENV', Environment.DEVELOPMENT)
+  @patch('wp1.web.app.CREDENTIALS', TEST_OAUTH_CREDS)
+  @patch('wp1.web.oauth.homepage_url',
+         TEST_OAUTH_CREDS[ENV]['CLIENT_URL']['homepage'])
+  @patch('wp1.web.oauth.handshaker', handshaker_no_email)
+  def test_complete_authorized_user_doesnt_overwrite_email(self):
+    self.app = create_app()
+    with self.override_db(self.app), self.app.test_client() as client:
+      with self.app.app_context():
+        wp10db = get_db('wp10db')
```
audiodude, line 122: When using any of the Base*Test classes, the application database is available as `self.wp10db`. There is no need to "connect" using an application context. Here and below.

wp1/web/oauth_test.py
```diff
@@ -92,6 +110,35 @@ def test_complete_unauthorized_user(self):
...
+        with wp10db.cursor() as cursor:
+          cursor.execute('INSERT INTO users (u_id, u_username, u_email) VALUES (%s, %s, %s)', (self.USER_NO_EMAIL['identity']['sub'], 'overwrite', 'test@test.it'))
```
audiodude, line 124: nit: Store `USER_NO_EMAIL['identity']['sub']` in a local variable. It's clunky and you reference it several times.

wp1/web/oauth.py
audiodude: Ah okay, reading your test, you're protecting against the case where a user deletes or removes their email address from Wikipedia (the returned OAuth email is NULL), or doesn't have one to begin with, but had added a preferred one in the GUI. The case I was thinking of was that they log in, submit user@example.com as their preferred email, then log in again and their Wikipedia email is user2@otherexample.com. So then we would overwrite their preferred email with their Wikipedia one. Which is probably not what we want.

wp1/web/oauth_test.py
```diff
@@ -109,8 +156,10 @@ def test_complete_authorized_user(self):
           rv.location)
       wp10db = get_db('wp10db')
       with wp10db.cursor() as cursor:
-        cursor.execute('SELECT * FROM users WHERE u_id = 1234')
-        self.assertNotEqual(None, cursor.fetchone())
+        cursor.execute('SELECT * FROM users WHERE u_id = %s', (self.USER['identity']['sub']))
+        row = cursor.fetchone()
+        self.assertIsNotNone(row)
+        self.assertEqual(row['u_email'].decode('utf-8'), self.USER['identity']['email'])
```
audiodude, line 162: In general, try not to use variables or expressions as the compared values of assertions. Hardcode the value (eg `"user@example.com"`) even if it leads to duplication. This makes the test easier to read and can reduce bugs.

Also, the order of assertEqual is `assertEqual(expected, actual)`. So the value you expect is the first argument, the value from the db should be the second.

audiodude (changes requested):
This makes sense, good work. Just a few questions.

wp1/web/oauth.py
nicolapace: I can do something like `u_email = IF(u_email IS NULL, VALUES(u_email), u_email)`

wp1/web/oauth.py
```diff
@@ -44,23 +44,31 @@ def complete():
...
-        {
-            'u_id': identity['sub'],
-            'u_username': identity['username']
-        })
+      '''
+      INSERT INTO users (u_id, u_username, u_email)
+      VALUES (%(u_id)s, %(u_username)s, %(u_email)s)
+      ON DUPLICATE KEY UPDATE
+        u_username = VALUES(u_username),
+        u_email = IF(u_email IS NULL, VALUES(u_email), u_email)
```
audiodude, line 64: I think we might need a column in the database like `u_is_manual_email`. There's too many edge cases here. For example, if the email `first@example.com` was set from OAuth, then their wikipedia/OAuth email gets updated to `second@example.com`, we *do* want to update the email, because it wasn't set manually by the user. So I think we need to know whether it was set manually. WDYT?

wp1/web/oauth.py
```diff
@@ -44,23 +44,31 @@ def complete():
...
-                      ON DUPLICATE KEY UPDATE u_id= %(u_id)s, u_username= %(u_username)s''',
-        {
-            'u_id': identity['sub'],
-            'u_username': identity['username']
-        })
+      '''
+      INSERT INTO users (u_id, u_username, u_email)
+      VALUES (%(u_id)s, %(u_username)s, %(u_email)s)
+      ON DUPLICATE KEY UPDATE
+        u_username = VALUES(u_username),
```
audiodude, line 63: If the username changes on Wikipedia/OAuth (I'm not sure if this is possible...), shouldn't we change it in the db?

audiodude (changes requested):
Still a few questions, thanks!

wp1/web/oauth.py
audiodude: You're right, I read it wrong!

nicolapace:
just save the email from oauth, when the user logs in

PR #926, opened 2025-06-29
Modify `builders.py` api for scheduled zim generations support

### LINKED ISSUE #912: modify zimfile generation api endpoint to support scheduled generations / emails
modify `wp1/web/builders.py`

nicolapace:
Introduce the ability to schedule repeated ZIM file generations directly from the `create_zim_file_for_builder()` endpoint.

- Extended `create_zim_file_for_builder()` to accept a `scheduled_repetitions` field from the request body.
- Added input validation to ensure `scheduled_repetitions` includes the required fields:
 - `repetition_period_in_months`
 - `number_of_repetitions`
 - `email`
- Passed the validated `scheduled_repetitions` to `handle_zim_generation()`.
- Updated `handle_zim_generation()` function signature to include the new optional parameter.

Fixes #912

### DIFF
--- a/wp1/web/builders.py
+++ b/wp1/web/builders.py
@@ -185,6 +185,8 @@ def create_zim_file_for_builder(builder_id):
 
   title = data.get('title')
   desc = data.get('description')
+  long_desc = data.get('long_description')
+  scheduled_repetitions = data.get('scheduled_repetitions')
 
   error_messages = []
   if not title:
@@ -193,10 +195,20 @@ def create_zim_file_for_builder(builder_id):
   if not desc:
     error_messages.append('Description is required for ZIM file')
 
+  if scheduled_repetitions is not None:
+    if not (
+      isinstance(scheduled_repetitions, dict) and
+      all(k in scheduled_repetitions for k in (
+          "repetition_period_in_months",
+          "number_of_repetitions",
+          "email",
+      ))
+  ):
+      error_messages.append('Invalid or missing fields in scheduled_repetitions')
+
   if error_messages:
     return flask.jsonify({'error_messages': error_messages}), 400
 
-  long_desc = data.get('long_description')
 
   try:
     logic_builder.handle_zim_generation(s3,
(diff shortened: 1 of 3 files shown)


wp1/web/builders.py
```diff
@@ -193,10 +195,20 @@ def create_zim_file_for_builder(builder_id):
...
+  if scheduled_repetitions is not None:
+    if not (
+      isinstance(scheduled_repetitions, dict) and
+      all(k in scheduled_repetitions for k in (
+          "repetition_period_in_months",
+          "number_of_repetitions",
+          "email",
+      ))
+  ):
+      error_messages.append('Invalid or missing fields in scheduled_repetitions')
```
audiodude, line 207: The `error_messages` list are user facing error messages. The user will not know what `scheduled_repetitions` is, and they will be unable to correct the error. This error will only occur if the frontend is misconfigured.

That's not to say we shouldn't return an error message. Instead, just do a:

```
return 'Invalid or missing fields in scheduled_repetitions', 400
```

So that a 400 status code will be sent with the given message as text (so it's not valid JSON and doesn't accidentally get shown to the user).

wp1/web/builders_test.py
```diff
@@ -549,6 +549,122 @@ def test_create_zim_file_for_builder_too_many_articles(
                        json={'description': 'Test description'})
       self.assertEqual('400 BAD REQUEST', rv.status)
 
+  @patch('wp1.zimfarm.request_zimfarm_task')
+  def test_create_zim_file_for_builder_scheduled_repetitions_valid(self,
```
audiodude, line 553: nit: No need to put "valid" in the name of the test function. The happy path is always test_some_function().

wp1/web/builders_test.py
```diff
@@ -549,6 +549,122 @@ def test_create_zim_file_for_builder_too_many_articles(
...
+
+    self.app = create_app()
+    with self.override_db(self.app), self.app.test_client() as client:
+      with client.session_transaction() as sess:
+        sess['user'] = self.USER
+      rv = client.post('/v1/builders/%s/zim' % builder_id,
+                       json={
+                           'title': 'Test title',
+                           'description': 'Test description',
+                           'scheduled_repetitions': {}
```
audiodude, line 636: I think when you build the frontend you might find that it's easier to allow an empty dict, and simply ignore it.

audiodude (approved):
LGTM

PR #929, opened 2025-06-30
Refactor `logic/builder.py:handle_zim_generation()` to be used in scheduled workers

### LINKED ISSUE #928: Refactor `wp1/logic/builder.py:handle_zim_generation()` to be more modular and flexible
We need to refactor the `handle_zim_generation()` function in `wp1/logic/builder.py`. The logic for requesting the ZIM file from the ZIM farm should be extracted so it can be reused.

We also need to add an option to rebuild selections related to the builder before the ZIM request is sent.

nicolapace:
This PR refactors zim file generation by decoupling `request_zim_file_for_builder()` from `handle_zim_generation()`, and creating a new `request_scheduled_zim_file_for_builder`.

The main reason for this change is to make the zim file request logic flexible for use in scheduled workers.
The new `request_scheduled_zim_file_for_builder()` now handles:
- skipping `user_id` checks, as these are not relevant for scheduled tasks
- re-opening db, redis and s3 connections independently
- rebuilding selections
- initiating the zimfarm request and updating the zim file status

Also a new `materialize_builder_with_connections()` has been added to skip repeated connections and clean the materialize process.

New tests have been added.

Fixes #928

### DIFF
--- a/wp1/logic/builder.py
+++ b/wp1/logic/builder.py
@@ -4,6 +4,7 @@ import logging
 
 import attr
 
+from typing import Type
 from kiwixstorage import KiwixStorage
 from pymysql.connections import Connection
 from redis import Redis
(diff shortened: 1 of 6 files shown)


wp1/logic/builder_test.py
```diff
@@ -1231,3 +1257,252 @@ def test_pending_zim_tasks_for(self):
...
     self.assertIn('9def', tasks)
+
+  @patch('wp1.logic.builder.zimfarm.request_zimfarm_task')
+  @patch('wp1.logic.builder.wp10_connect')
+  @patch('wp1.logic.builder.redis_connect')
+  @patch('wp1.logic.builder.connect_storage')
+  @patch('wp1.logic.builder.utcnow',
+         return_value=datetime.datetime(2022, 12, 25, 0, 1, 2))
+  def test_request_zim_file_for_builder(self, patched_utcnow,
+                                        patched_connect_storage,
```
audiodude, line 1268: nit: I think in most of the other tests I refer as mock_*, not patched_*. Please update for consistency.

audiodude (changes requested):
Good work, but I think we can restructure `materialize_builder`/`materialize_builder_with_connections`.

wp1/logic/builder_test.py
nicolapace: I also used `patched_*` in other tests, i'll fix it wherever I find it

wp1/logic/builder.py
```diff
@@ -135,8 +141,36 @@ def get_builder(wp10db, id_):
       raise ObjectNotFoundError(f'No builder with id={id_}')
     return Builder(**db_builder)
 
+def materialize_builder_with_connections(s3: KiwixStorage,
```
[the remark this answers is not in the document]
nicolapace: That's better, I also changed it to get a `builder` instead of `builder_id` so that it doesn't re-do the same query to get the `builder` multiple times.

wp1/logic/builder.py
```diff
@@ -136,31 +143,39 @@ def get_builder(wp10db, id_):
...
+                        redis: Redis = None,
+                        wp10db: Connection = None):
+  """
+  Materializes a builder selection using the given builder class and content type.
+  This is intended to be used by worker processes.
+  """
+  should_close = False
+  if wp10db is None:
+    wp10db = wp10_connect()
+    should_close = True
```
audiodude, line 159: No need for this. Just check `if wp10db is not None`.

audiodude (approved):

wp1/logic/builder.py
nicolapace: I didn't get it. We want to close the database connection if `wp10db` was not passed to the function, right?

audiodude: You're right, we only want to close it if we opened it, so the boolean is correct.

PR #934, opened 2025-07-03
Implement scheduling for future ZIM file generations

### LINKED ISSUE #911: Refactor zimfile generation to support scheduled generations

nicolapace:
- Added a new function `schedule_future_zimfile_generations()` in `queues.py that:
  - calculates next first future xim generation run and interval in months (1 month=30days)
  - schedules future zim-generations
  - inserts schedule in the DB

- Added call to `schedule_future_zimfile_generations()` function in `handle_zim_generations()`
- Added tests

Fixes #911

### DIFF
--- a/wp1/logic/builder_test.py
+++ b/wp1/logic/builder_test.py
@@ -1313,8 +1313,7 @@ class BuilderTest(BaseWpOneDbTest):
     
     with patch('wp1.logic.builder.importlib.import_module') as mock_import:
       mock_module = MagicMock()
-      mock_builder_cls = MagicMock()
-      mock_module.Builder = mock_builder_cls
+      mock_module.Builder = MagicMock()
       mock_import.return_value = mock_module
       
       actual = logic_builder.request_scheduled_zim_file_for_builder(
@@ -1410,65 +1409,6 @@ class BuilderTest(BaseWpOneDbTest):
                                                   mock_get_builder,
                                                   mock_materialize_builder):
     """Test zimfile request with rebuild_selection=True"""
-    self._insert_builder()
-    self._insert_selection(1, 'text/tab-separated-values', builder_id=self.builder.b_id)
-
-    #insert a zim_schedule
-    with self.wp10db.cursor() as cursor:
-      cursor.execute('INSERT INTO zim_schedules (s_id, s_builder_id, s_rq_job_id, s_last_updated_at) '
-                     'VALUES (%s, %s, %s, %s)',
-                     (b'schedule_123', self.builder.b_id, b'rq_job_id_123', b'20191225044444'))
-      self.wp10db.commit()
-
-    mock_wp10_connect.return_value = self.wp10db
-    redis_mock = MagicMock()
-    s3_mock = MagicMock()
-    mock_redis_connect.return_value = redis_mock
-    mock_connect_storage.return_value = s3_mock
-    
-    mock_request_zimfarm_task.return_value = 'test_task_id_456'
-    
-    mock_get_builder.return_value = self.builder
-    
-    with patch('wp1.logic.builder.importlib.import_module') as mock_import:
-      mock_module = MagicMock()
-      mock_builder_cls = MagicMock()
-      mock_module.Builder = mock_builder_cls
-      mock_import.return_value = mock_module
-      
-      actual = logic_builder.request_scheduled_zim_file_for_builder(
-          builder=self.builder,
-          title='Test Title',
-          description='Test Description',
-          long_description='Test Long Description',
-          zim_schedule_id='schedule_123'
-      )
-    
-    self.assertEqual('test_task_id_456', actual)
-    mock_get_builder.assert_called_once_with(self.wp10db, self.builder.b_id)
-    mock_materialize_builder.assert_called_once()
-    mock_request_zimfarm_task.assert_called_once()
-
-    #check that the zim_schedule was updated
-    with self.wp10db.cursor() as cursor:
-      cursor.execute('SELECT s_zim_file_id, s_last_updated_at '
-                     'FROM zim_schedules WHERE s_id = %s', ('schedule_123',))
-      schedule = cursor.fetchone()
-      self.assertIsNotNone(schedule['s_zim_file_id'])
-      self.assertEqual( b'20221225000102', schedule['s_last_updated_at'])
-
-  @patch('wp1.logic.builder.zimfarm.request_zimfarm_task')
-  @patch('wp1.logic.builder.wp10_connect')
-  @patch('wp1.logic.builder.redis_connect')
-  @patch('wp1.logic.builder.connect_storage')
-  @patch('wp1.logic.zim_schedules.utcnow',
-         return_value=datetime.datetime(2022, 12, 25, 0, 1, 2))
-  def test_request_zim_file_for_builder_empty_descriptions(self, mock_utcnow,
-                                                           mock_connect_storage,
-                                                           mock_redis_connect,
-                                                           mock_wp10_connect,
-                                                           mock_request_zimfarm_task):
-    """Test zimfile request with empty/None descriptions"""
     self._insert_builder()
     self._insert_selection(1, 'text/tab-separated-values', builder_id=self.builder.b_id)
 
--- a/wp1/logic/builder_test.py
+++ b/wp1/logic/builder_test.py
@@ -1328,32 +1328,20 @@ class BuilderTest(BaseWpOneDbTest):
     mock_materialize_builder.assert_called_once()
     mock_request_zimfarm_task.assert_called_once()
 
-  @patch('wp1.logic.builder.materialize_builder')
   @patch('wp1.logic.builder.get_builder')
-  @patch('wp1.logic.builder.zimfarm.request_zimfarm_task')
   @patch('wp1.logic.builder.wp10_connect')
   @patch('wp1.logic.builder.redis_connect')
   @patch('wp1.logic.builder.connect_storage')
-  @patch('wp1.logic.zim_schedules.utcnow',
+  @patch('wp1.logic.builder.utcnow',
          return_value=datetime.datetime(2022, 12, 25, 0, 1, 2))
-  def test_request_scheduled_zim_file_for_builder_with_zim_schedule(self,
-                                                  mock_utcnow,
-                                                  mock_connect_storage,
-                                                  mock_redis_connect,
-                                                  mock_wp10_connect,
-                                                  mock_request_zimfarm_task,
-                                                  mock_get_builder,
-                                                  mock_materialize_builder):
-    """Test zimfile request with rebuild_selection=True"""
+  def test_request_scheduled_zim_file_for_builder_missing_class(
+      self, mock_utcnow, mock_connect_storage, mock_redis_connect,
+      mock_wp10_connect, mock_get_builder):
+    """Test zimfile request with missing builder class in module"""
     self._insert_builder()
-    self._insert_selection(1, 'text/tab-separated-values', builder_id=self.builder.b_id)
-
-    #insert a zim_schedule
-    with self.wp10db.cursor() as cursor:
-      cursor.execute('INSERT INTO zim_schedules (s_id, s_builder_id, s_rq_job_id, s_last_updated_at) '
-                     'VALUES (%s, %s, %s, %s)',
-                     (b'schedule_123', self.builder.b_id, b'rq_job_id_123', b'20191225044444'))
-      self.wp10db.commit()
+    self._insert_selection(1,
+                           'text/tab-separated-values',
+                           builder_id=self.builder.b_id)
 
     mock_wp10_connect.return_value = self.wp10db
     redis_mock = MagicMock()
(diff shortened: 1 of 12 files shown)


wp1/queues.py
```diff
@@ -182,3 +191,54 @@ def poll_for_zim_file_status(redis, task_id):
...
+  if not isinstance(scheduled_repetitions, dict) or not required_keys <= scheduled_repetitions.keys():
+    raise ValueError(f'scheduled_repetitions must be a dict containing {required_keys}')
+
+  queue = _get_zimfile_scheduling_queue(redis)
+  scheduler = Scheduler(connection=queue.connection, queue=queue)
+
+  period_months = scheduled_repetitions['repetition_period_in_months']
+  num_repetitions = scheduled_repetitions['number_of_repetitions']
+  
+  # Calculate the first run time: period_months from now
```
audiodude, line 215: nit: More comments that just restate what is happening on the next line.

A good comment doesn't explain what the code is doing. The code explains what the code is doing. A good comment explains *why* the code is doing that, if it's something tricky or surprising.

wp1/queues.py
```diff
@@ -182,3 +191,54 @@ def poll_for_zim_file_status(redis, task_id):
...
+  first_run = datetime.now(UTC) + relativedelta(months=period_months)
+
+  # Calculate the interval in seconds for the repetition period in months
+  interval_seconds = period_months * 30 * 24 * 3600  # Approximate month as 30 days
```
audiodude, line 219: For example, the fact that we normalize a "month" to 30 days. That's good material for a comment.

wp1/queues.py
```diff
@@ -182,3 +191,54 @@ def poll_for_zim_file_status(redis, task_id):
...
+    scheduled_time=first_run,
+    func=logic_builder.request_scheduled_zim_file_for_builder,
+    args=[builder, title, description, long_description],
+    interval=interval_seconds,
+    repeat=num_repetitions,
+    queue_name='zimfile-scheduling',
+  )
+
+  # Insert the new schedule into the zim_schedules table using the model
+  from wp1.models.wp10.zim_schedule import ZimSchedule
```
audiodude, line 231: Don't import inside of functions. Imports at the top please.

wp1/queues.py
```diff
@@ -182,3 +191,54 @@ def poll_for_zim_file_status(redis, task_id):
...
+  period_months = scheduled_repetitions['repetition_period_in_months']
+  num_repetitions = scheduled_repetitions['number_of_repetitions']
+  
+  # Calculate the first run time: period_months from now
+  first_run = datetime.now(UTC) + relativedelta(months=period_months)
+
+  # Calculate the interval in seconds for the repetition period in months
+  interval_seconds = period_months * 30 * 24 * 3600  # Approximate month as 30 days
+
+  job = scheduler.schedule(
```
audiodude, line 221: So the way this is set up is confusing to readers of the code. The `scheduler.schedule` function takes `scheduled_time` and `repeat`/`interval`. I think it has already introduced an "off by one error" in your code. Why is the first time not now? If the user requests a ZIM file with 3 repetitions, 1 month apart, this looks like the first one is done immediately, "now", outside of this code. But then it schedules one a month from now, and 3 more, for 4 total right?

Why can't we just have the API/web method check for scheduled repetitions and if it's found, we call this function immediately and schedule the first instance here, instead of doing it outside and only doing the repetitions here? Is it because we don't want to rebuild the selection the first time?

wp1/queues_test.py
```diff
@@ -190,3 +191,84 @@ def test_poll_for_zim_file_status(self, patched_scheduler, patched_queue):
     queues.poll_for_zim_file_status(self.redis, '1234')
 
     scheduler_mock.enqueue_in.assert_called_once()
+
+  @patch('wp1.queues.Scheduler')
+  @patch('wp1.queues.uuid.uuid4')
+  @patch('wp1.queues.logic_zim_schedules.insert_zim_schedule')
+  def test_schedule_future_zimfile_generations(self, patched_insert, patched_uuid4, patched_scheduler):
+    mock_builder = Builder(
```
audiodude, line 199: I've seen this in your other PRs too, but technically this is not a mock builder, it's a real Builder object.

wp1/queues_test.py
```diff
@@ -190,3 +191,84 @@ def test_poll_for_zim_file_status(self, patched_scheduler, patched_queue):
     queues.poll_for_zim_file_status(self.redis, '1234')
 
     scheduler_mock.enqueue_in.assert_called_once()
+
+  @patch('wp1.queues.Scheduler')
+  @patch('wp1.queues.uuid.uuid4')
+  @patch('wp1.queues.logic_zim_schedules.insert_zim_schedule')
+  def test_schedule_future_zimfile_generations(self, patched_insert, patched_uuid4, patched_scheduler):
```
audiodude, line 198: nit: Same comment about `patched_*`. I think the rest of the code uses `mock_*`.

wp1/queues_test.py
```diff
@@ -190,3 +191,84 @@ def test_poll_for_zim_file_status(self, patched_scheduler, patched_queue):
...
+    patched_scheduler.return_value.schedule.return_value = job_mock
+    patched_uuid4.return_value = 'uuid-1'
+
+    result = queues.schedule_future_zimfile_generations(
+        self.redis, self.wp10db,
+        mock_builder, title, description, long_description,
+        scheduled_repetitions
+    )
+
+    # Should return the job ID
```
audiodude, line 226: nit: Unnecessary comment.

wp1/queues_test.py
```diff
@@ -190,3 +191,84 @@ def test_poll_for_zim_file_status(self, patched_scheduler, patched_queue):
     queues.poll_for_zim_file_status(self.redis, '1234')
 
     scheduler_mock.enqueue_in.assert_called_once()
+
+  @patch('wp1.queues.Scheduler')
+  @patch('wp1.queues.uuid.uuid4')
+  @patch('wp1.queues.logic_zim_schedules.insert_zim_schedule')
```
audiodude, line 197: This is a function that just writes to the database right? Why do we have to mock it? Can't we write to the test database and then query?

audiodude (changes requested):

wp1/queues.py
nicolapace: And also yes, 'number_of_repetitions' is actually number of 'future' repetitions, not counting the ZIM generation done now.

audiodude (commented):
This looks good in general. Let's discuss "now" ZIMs versus scheduled in our next meeting.

audiodude (commented):
Some notes from our meeting.

wp1/queues.py
```diff
@@ -182,3 +192,54 @@ def poll_for_zim_file_status(redis, task_id):
...
+
+  zim_schedule = ZimSchedule(
+      s_id=zim_schedule_id.encode('utf-8'),
+      s_builder_id=builder.b_id,
+      s_zim_file_id=None,
+      s_rq_job_id=job.id.encode('utf-8'),
+      s_interval=period_months,
+      s_remaining_generations=num_scheduled_repetitions,
+      # s_email=scheduled_repetitions['email'].encode('utf-8'),
+      s_last_updated_at=datetime.now(UTC).strftime(constants.TS_FORMAT_WP10).encode('utf-8'),
```
audiodude, line 241: Isn't this what `set_lasted_updated_at_now` is for, in the model class?

wp1/queues.py
```diff
@@ -182,3 +192,54 @@ def poll_for_zim_file_status(redis, task_id):
...
+    queue_name='zimfile-scheduling',
+  )
+
+  zim_schedule = ZimSchedule(
+      s_id=zim_schedule_id.encode('utf-8'),
+      s_builder_id=builder.b_id,
+      s_zim_file_id=None,
+      s_rq_job_id=job.id.encode('utf-8'),
+      s_interval=period_months,
+      s_remaining_generations=num_scheduled_repetitions,
```
audiodude, line 239: Same off by one error, but in the other direction now. If we scheduled `first_run` one month from now, and then `num_scheduled_repetitions` repeats, there will be `num_scheduled_repetitions` + 1 remaining generations, correct? We are doing one right now, and repeating N times, once at `first_future_time` and then N-1 times after.

wp1/queues.py
```diff
@@ -182,3 +192,54 @@ def poll_for_zim_file_status(redis, task_id):
...
+    raise ValueError(f'scheduled_repetitions must be a dict containing {required_keys}')
+
+  queue = _get_zimfile_scheduling_queue(redis)
+  scheduler = Scheduler(connection=queue.connection, queue=queue)
+
+  period_months = scheduled_repetitions['repetition_period_in_months']
+  # Calculate the number of scheduled repetitions minus one, 
+  # since the first run is still considered a repetition.
+  num_scheduled_repetitions = scheduled_repetitions['number_of_repetitions'] - 1
+  first_future_run = datetime.now(UTC) + relativedelta(months=period_months)
```
audiodude, line 217: Let's calculate the interval_seconds and use that for the timedelta for the first future run. It seems like we will have headaches with the "first" future run being a calendar month away (whatever that means to `relativedelta`) and then subsequent runs being exactly and then ones after that being 30 days.

wp1/queues.py
```diff
@@ -182,3 +192,54 @@ def poll_for_zim_file_status(redis, task_id):
...
+                                        builder: Builder,
+                                        title: str,
+                                        description: str,
+                                        long_description: str,
+                                        scheduled_repetitions: dict):
+  """
+  Schedule future ZIM file creations using rq-scheduler.
+  """
+  required_keys = {'repetition_period_in_months', 'number_of_repetitions', 'email'}
+  if not isinstance(scheduled_repetitions, dict) or not required_keys <= scheduled_repetitions.keys():
```
audiodude, line 207: nit: This is a bit of a weird construction. Let's extract a temporary variable to make the condition more clear:

```
has_min_required_keys = required_keys <= scheduled_repetitions.keys()
if not isinstance(...) or not has_min_required_keys:
   ...
```

wp1/queues.py
```diff
@@ -182,3 +192,54 @@ def poll_for_zim_file_status(redis, task_id):
   scheduler = Scheduler(queue=poll_q, connection=redis)
   scheduler.enqueue_in(timedelta(minutes=2),
                        logic_builder.on_zim_file_status_poll, task_id)
+
+def schedule_future_zimfile_generations(redis: Redis,
```
audiodude, line 196: Nit: #956

wp1/queues_test.py
```diff
@@ -176,3 +178,94 @@ def test_enqueue_materialize(self, mock_materialize_builder,
...
+
+    self.assertEqual('job-id', result)
+    # Verify scheduler.schedule was called with correct parameters
+    mock_scheduler.return_value.schedule.assert_called_once()
+    _, call_kwargs = mock_scheduler.return_value.schedule.call_args
+    self.assertEqual(queues.logic_builder.request_scheduled_zim_file_for_builder, call_kwargs['func'])
+
+    expected_args = [builder, title, description, long_description, ANY]
+    self.assertEqual(expected_args, call_kwargs['args'])
+    expected_interval = scheduled_repetitions['repetition_period_in_months'] * 30 * 24 * 3600
```
audiodude, line 232: nit: We don't want to calculate expected values, they should all be literals/constants as much as possible.

wp1/queues_test.py
```diff
@@ -176,3 +178,94 @@ def test_enqueue_materialize(self, mock_materialize_builder,
...
+    result = queues.schedule_future_zimfile_generations(
+        self.redis, self.wp10db,
+        builder, title, description, long_description,
+        scheduled_repetitions
+    )
+
+    self.assertEqual('job-id', result)
+    # Verify scheduler.schedule was called with correct parameters
+    mock_scheduler.return_value.schedule.assert_called_once()
+    _, call_kwargs = mock_scheduler.return_value.schedule.call_args
```
audiodude, line 227: Do we need this, or can we do `mock_scheduler.assert_called_once_with(foo="bar")`

audiodude (changes requested):
Looking better, good work. A few comments still.

wp1/logic/builder_test.py
```diff
@@ -1309,33 +1310,124 @@ def test_request_scheduled_zim_file_for_builder(
     mock_request_zimfarm_task.return_value = 'test_task_id_456'
 
     mock_get_builder.return_value = self.builder
```
audiodude, line 1312: I know this isn't from this PR, but why do we have to mock `get_builder`? We inserted the builder above, can't we get it from the actual test database?

wp1/logic/builder_test.py
```diff
@@ -1309,33 +1310,124 @@ def test_request_scheduled_zim_file_for_builder(
...
+    
+    with patch('wp1.logic.builder.importlib.import_module') as mock_import:
```
audiodude, line 1314: I think the pattern that we use is patching everything as a decorator at the top of the test function. Is there a particular reason why you're patching with a context manager?

wp1/logic/builder_test.py
```diff
@@ -1309,33 +1310,124 @@ def test_request_scheduled_zim_file_for_builder(
...
+                           'text/tab-separated-values',
+                           builder_id=self.builder.b_id)
+
+    mock_wp10_connect.return_value = self.wp10db
+    redis_mock = MagicMock()
+    s3_mock = MagicMock()
+    mock_redis_connect.return_value = redis_mock
+    mock_connect_storage.return_value = s3_mock
+    
+    mock_get_builder.return_value = self.builder
```
audiodude, line 1352: Same question here, why do we have to mock `get_builder`?

wp1/logic/builder_test.py
```diff
@@ -1309,33 +1310,124 @@ def test_request_scheduled_zim_file_for_builder(
...
+    
+    with patch('wp1.logic.builder.importlib.import_module') as mock_import:
```
audiodude, line 1354: Same question here, why are we patching with a context manager?

wp1/logic/builder_test.py
```diff
@@ -1309,33 +1310,124 @@ def test_request_scheduled_zim_file_for_builder(
...
+      mock_module.Builder = None  # Simulate missing Builder class 
+      mock_import.return_value = mock_module
+      with self.assertRaises(ImportError) as cm:
+          actual = logic_builder.request_scheduled_zim_file_for_builder(
+              builder=self.builder,
+              title='Test Title',
+              description='Test Description',
+              long_description='Test Long Description'
+          )
+      self.assertIn('Builder class not found in module', str(cm.exception))
```
audiodude, line 1365: We generally don't assert on exception messages.

wp1/logic/builder_test.py
```diff
@@ -1309,33 +1310,124 @@ def test_request_scheduled_zim_file_for_builder(
...
+      
+  @patch('wp1.logic.builder.materialize_builder')
+  @patch('wp1.logic.builder.get_builder')
+  @patch('wp1.logic.builder.zimfarm.request_zimfarm_task')
+  @patch('wp1.logic.builder.wp10_connect')
+  @patch('wp1.logic.builder.redis_connect')
+  @patch('wp1.logic.builder.connect_storage')
+  @patch('wp1.logic.zim_schedules.utcnow',
+         return_value=datetime.datetime(2022, 12, 25, 0, 1, 2))
+  def test_request_scheduled_zim_file_for_builder_with_zim_schedule(self,
```
audiodude, line 1375: Same things in this function re: mock_get_builder and the context manager

audiodude (approved):

wp1/logic/builder_test.py
nicolapace: yes, totally right, thanks!

PR #935, opened 2025-07-04
add support for updating `zim_schedule` with `zim_file_id`

nicolapace:
Add/Update `zim_file_id` to `zim_schedule` when the scheduled file is requested from the zimfarm.

### DIFF
--- a/wp1/logic/builder.py
+++ b/wp1/logic/builder.py
@@ -10,6 +10,7 @@ from pymysql.connections import Connection
 from redis import Redis
 
 import wp1.logic.selection as logic_selection
+import wp1.logic.zim_schedules as logic_zim_schedules
 import wp1.logic.util as logic_util
 from wp1 import app_logging, queues, zimfarm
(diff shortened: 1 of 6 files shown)


audiodude:
> Upon further consideration, I'm wondering whether the `s_zim_file_id` field in the `zim_schedules` table is actually necessary. Would it be sufficient to use `latest_zim_file_for()` to retrieve the corresponding `zim_file` from the latest selection? It seems like we might be able to simplify the schema unless there's an edge case I'm overlooking.
> 
> This PR is all about storing the zim_file at the time the schedule is created, but I'm starting to think this may not be required.
> 
> The only potential justification I can see for keeping the field is to handle the case where a user manually triggers a new zimfile generation shortly after a scheduled one. In that case, `latest_zim_file_for()` might point to the most recent (manual) one rather than the one generated by the scheduler.

Discussed offline, I believe it is necessary for finding the zim schedule to notify when the hook is called by the zimfarm.

audiodude (approved):
LGTM

PR #950, opened 2025-07-21
Use latest selection when creating zimfarm schedules

### LINKED ISSUE #947: Use `latest.{ext}` selection endpoint when requesting zimfiles to zimfarm

nicolapace:
Refactor schedule name and zimfile name generation.
Use latest selection from api when creating ZIMFARM schedules.

Fixes #947


audiodude:
@nicolapace I think this change broke using the production Zimfarm for dev ZIMs. For example:

image

Notice that the article list URL is `localhost`, which will obviously not work.

We need to prioritize #1008

PR #953, opened 2025-07-21
Associate zimfarm schedule with builder

### LINKED ISSUE #952: Associate ZIMFARM schedules with builder instead of selection
Associate ZIMFARM schedules with builder instead of selection, so that the zimfarm schedule can be re-used with the latest materialized selection.

nicolapace:
Associate ZIMFARM schedules with builder instead of selection, so that the zimfarm schedule can be re-used with the latest materialized selection.

Fixes #952

### DIFF
--- a/wp1/zimfarm.py
+++ b/wp1/zimfarm.py
@@ -173,17 +173,17 @@ def _validate_zim_metadata(title=None, description=None, long_description=None):
 def get_zimfarm_schedule_name(builder_id: str) -> str:
     """Generate a unique schedule name for the ZIM file based on builder"""
     if not builder_id:
-        raise ObjectNotFoundError('Builder ID cannot be None')
+        raise ValueError('Builder ID cannot be None')
     parts = builder_id.split('-')
     # Use last two parts of the UUIDv4 for more uniqueness
     short_id = ''.join(parts[-2:])
-    return f'wp1_builder_{short_id}'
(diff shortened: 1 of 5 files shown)


wp1/zimfarm.py
```diff
@@ -170,14 +170,14 @@ def _validate_zim_metadata(title=None, description=None, long_description=None):
...
 
-def get_zimfarm_schedule_name(selection_id: str) -> str:
-    """Generate a unique schedule name for the ZIM file based on builder and selection."""
-    if not selection_id:
-        raise ObjectNotFoundError('Selection ID cannot be None')
-    parts = selection_id.split('-')
+def get_zimfarm_schedule_name(builder_id: str) -> str:
+    """Generate a unique schedule name for the ZIM file based on builder"""
+    if not builder_id:
+        raise ObjectNotFoundError('Builder ID cannot be None')
```
audiodude, line 176: Technically this isn't really an `ObjectNotFoundError` because we didn't attempt to "look for" an object.

audiodude (changes requested):
So we're not changing the DB schema, we're just associating the Zimfarm "schedule name" with the builder ID, right?

audiodude (approved):
I'm not exactly sure what this PR is doing, there seems to be a lot of changes from other branches mixed in?

It looks fine though, so feel free to merge however makes sense to you.

PR #966, opened 2025-08-07
Reuse zim schedules

### LINKED ISSUE #964: ZIM schedule migration: check for schedule when requesting manual ZIM
In the previous code, we deleted the ZIM schedule from the Zimfarm after scheduling a single task. In the new version, we will be reusing ZIM schedules, both for one-off jobs and recurring jobs.

Because of this, we need to check when the user requests a manual ZIM if a schedule exists or not. If the selection doesn't have a row in the `zim_schedules` table, then we have our answer. However, if there is a migrated `zim_schedule`, then we need to make a request to the Zimfarm to check if there is an existing schedule, and if not, create it.

nicolapace:
This PR adds functionality to re-use zim schedules, updating or creating new ones depending if the exists in the zimfarm or are already used.

Fixes #964 (ZIM schedule migration: check for schedule when requesting manual ZIM)

the zimfile schedule name is changed to not use the builder id so that more then 1 schedule can be created per builder (manual + scheduled)

### DIFF
--- a/wp1/logic/builder.py
+++ b/wp1/logic/builder.py
@@ -425,7 +425,7 @@ def request_zim_file_task_for_builder(redis: Redis,
   Requests a ZIM file from the Zimfarm for the given builder.
   Returns the entire ZimTask object instead of just the z_id.
   """
-  task_id = zimfarm.request_zimfarm_task(redis, wp10db, builder)
+  task_id = zimfarm.request_zimfarm_task(redis, wp10db, builder, zim_schedule_id)
   selection = latest_selection_for(wp10db, builder.b_id,
            'text/tab-separated-values')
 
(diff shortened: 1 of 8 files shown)


wp1/zimfarm.py
```diff
@@ -336,10 +359,68 @@ def create_zimfarm_schedule(redis,
   return zim_schedule
 
 
+def update_zimfarm_schedule(redis,
```
audiodude, line 362: Maybe we can reuse the logic and do something like `create_or_update_zimfarm_schedule`?

wp1/logic/builder.py
```diff
@@ -499,17 +515,30 @@ def handle_zim_generation(redis,
           'Could not use builder id = %s for user id = %s' %
           (builder_id, user_id))
   
-  zim_schedule : ZimSchedule = zimfarm.create_zimfarm_schedule(redis,
-                                  wp10db,
-                                  builder,
-                                  title=title,
-                                  description=description,
-                                  long_description=long_description)
+  existing_schedule_in_db = find_existing_schedule_in_db(wp10db, builder.b_id, scheduled_repetitions)
```
audiodude, line 518: Put this into the zimfarm.create_or_update method.

wp1/zimfarm.py
```diff
@@ -174,11 +174,11 @@ def _validate_zim_metadata(title=None, description=None, long_description=None):
         f"Long description must be different from the description.")
 
 
-def get_zimfarm_schedule_name(builder_id: str) -> str:
-    """Generate a unique schedule name for the ZIM file based on builder"""
-    if not builder_id:
-        raise ValueError('Builder ID cannot be None')
-    parts = builder_id.split('-')
+def get_zimfarm_schedule_name(schedule_id: str) -> str:
```
audiodude, line 177: We can leave this associated with the Builder because there should only ever be 0 or 1 schedule per Builder.

wp1/logic/builder.py
```diff
@@ -471,6 +471,22 @@ def request_scheduled_zim_file_for_builder(builder: Builder, zim_schedule_id: by
...
+def find_existing_schedule_in_db(wp10db, builder_b_id, scheduled_repetitions):
+  """
+  Returns an existing schedule based on scheduled_repetitions.
+  If scheduled_repetitions is set, looks for a manual schedule (s_remaining_generations is None).
+  Otherwise, looks for a schedule with s_remaining_generations == 0.
+  """
+  
+  # Get all ZIM schedules for this builder
+  schedules = logic_zim_schedules.list_zim_schedules_for_builder(wp10db, builder_b_id)
+  print(f"Existing schedules in DB: {schedules}, scheduled_repetitions={scheduled_repetitions}")
```
audiodude, line 484: nit: remove prints, here and elsewhere.

audiodude (commented):
Just a few questions.

PR #968, opened 2025-08-12
update zimfarm api calls

elfkuzco:
# Rationale
The zimfarm API is upgraded to v2 with major differences revolving around:
- the naming of API routes
- Shape of payload/data
- Modification of shape of data in DB

This PR serves as the base for the modifications of the wp1 code to conform to the contract of the new API.

### DIFF
--- a/wp1/zimfarm_test.py
+++ b/wp1/zimfarm_test.py
@@ -1053,7 +1053,8 @@ class ZimFarmTest(BaseWpOneDbTest):
 
   @patch('wp1.zimfarm.get_zimfarm_token')
   @patch('wp1.zimfarm.requests.delete')
-  def test_delete_zimfarm_schedule_by_builder_id_success(self, patched_delete, patched_get_zimfarm_token):
+  def test_delete_zimfarm_schedule_by_builder_id_success(
+      self, patched_delete, patched_get_zimfarm_token):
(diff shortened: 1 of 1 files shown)


audiodude (approved):

benoit74 (approved):
Still OK for me

benoit74:
This is now ready/mandatory to be deployed, Zimfarm has been upgrade to new v2 endpoint.

@audiodude can you please have a look at why we have two failing tests and merge and deploy this once fixed?

Thank you

audiodude:
Okay I'm getting this on production:

image

It obviously means that we either didn't get the token properly, or we're not sending it correctly.

audiodude:
> @audiodude , is it possible the env has not been updated to `v2`?

I made that mistake originally, but you can see in my screenshot that the url that is mentioned has v2 in it.

elfkuzco:
created #1010 to fix the issues in the payload

audiodude:
@elfkuzco  Still broken:

image

Does the Zimfarm return useful content in the body of these 400 messages? If so I can look in the logs if we are logging them, or add that logging if it will help.

elfkuzco:
Okay. Do you kmow any selections that could yield small zims? I tried doing some e2e tests with this and the task took a while and I interrupted.  So, it never really ran e2e but the manager was able to pick up a task

PR #982, opened 2025-08-22
Zim long desc missing

### LINKED ISSUE #946: Refine ZIM metadata
Image

We should allow users to not input any long description, this metadata is optional

We should make it clearer from the beginning that title and description are mandatory. For now if you do not enter the combobox, you are not displayed with an error, you just have a disabled "Request ZIM file" button.

audiodude:
Fixes #946

Although we had already fixed the fronted part of the issue, it was still failing with a backend error.

Basically, the original design assumed that most/all ZIM metdata was "optional", however this later turned out to be an incorrect assumption. This lead to fields like `title` and `description` being keyword/optional arguments when calling the entrypoint for creating ZIM files, and had a knock on effect of making it optional in other places. However, because of this, the validation process was unnecessarily confusing and complex.

This PR makes title and description required, while leaving long_description optional, which is more semantically in line with the actual requirements.

We've also added an exception handler in the API/web endpoint so that ZIMs with improper metadata result in 400 errors to the client.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
         # Some unit tests also try to write to Redis
         image: redis
         ports:
-          - 6379:6379
+          - 9777:6379
         options: --health-cmd "redis-cli ping" --health-interval 10s --health-timeout 5s --health-retries 5
 
     steps:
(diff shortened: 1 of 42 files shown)


audiodude (commented):
Apologies for the formatting changes @nicolapace. It seems like we never got you properly set up with `yapf` while you were doing your development.

audiodude:
@nicolapace ready for a look.

audiodude:
Sorry I was trying to fix merge conflicts and accidentally merged this PR!

PR #992, opened 2025-08-29
Manually delete zim schedule

### LINKED ISSUE #976: Manage schedules already in use
- Show an error saying that the schedule is already in use
- Add an option to delete the schedule

nicolapace:
image
Fixes #976

### DIFF
--- a/wp1-frontend/src/components/ZimFile.vue
+++ b/wp1-frontend/src/components/ZimFile.vue
@@ -42,7 +42,7 @@
             </button>
           </div>
           
-          <div v-if="status === 'NOT_REQUESTED'">
+          <div v-else-if="status === 'NOT_REQUESTED'">
             <p>
               Use this form to create a ZIM file from your selection, so that
               you can browse the articles it contains offline. The
@@ -94,9 +94,11 @@
       </div>
       <div
         v-if="
-          status === 'NOT_REQUESTED' ||
-          status === 'FAILED' ||
-          (status != 'REQUESTED' && isDeleted)
+          !activeSchedule && (
+            status === 'NOT_REQUESTED' ||
+            status === 'FAILED' ||
+            (status != 'REQUESTED' && isDeleted)
+          )
         "
         class="row"
       >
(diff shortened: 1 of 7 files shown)


wp1/logic/builder.py
```diff
@@ -587,10 +593,38 @@ def zim_file_status_for(wp10db, builder_id):
       'utf-8') if zim_schedule and zim_schedule.s_description else None
   data['long_description'] = zim_schedule.s_long_description.decode(
       'utf-8') if zim_schedule and zim_schedule.s_long_description else None
+  active_schedule = logic_zim_schedules.find_active_recurring_schedule_for_builder(wp10db, builder_id)
```
audiodude, line 596: Can you refactor it so that `active_schedule` is populated unconditionally before the `not zim_file` check?

wp1/web/builders.py
```diff
@@ -292,3 +292,39 @@ def latest_zim_file_for_builder(builder_id):
...
+    active_schedule = logic_zim_schedules.find_active_recurring_schedule_for_builder(wp10db, builder_id.encode('utf-8'))
+    if not active_schedule:
+      return flask.jsonify({'error_messages': ['No active recurring schedule found']}), 404
+      
+    # Delete the schedule
+    deleted = logic_zim_schedules.delete_zim_schedule(redis, wp10db, active_schedule.s_id)
+    if not deleted:
+      return flask.jsonify({'error_messages': ['Failed to delete schedule']}), 500
+      
+  except Exception as e:
```
audiodude, line 326: No need to catch non-specific exceptions. The server will automatically return a 500 if an exception is raised.

wp1-frontend/src/components/ZimFile.vue
```diff
@@ -18,6 +18,30 @@
       <div class="row">
         <div class="col-lg-6 col-md-9 mx-4">
           <h3>Create ZIM file</h3>
+          
+          <!-- Active Schedule Warning -->
+          <div v-if="activeSchedule" class="alert alert-warning" role="alert">
```
audiodude, line 23: I can't see the action button from your screenshot, but it seems from the changes here that a user would be able to interact with the "Create a ZIM" form, even when they have an active schedule. That could cause problems. Let's hide the form and only show this message if there is an active schedule.

audiodude (changes requested):

PR #1001, opened 2025-09-05
Show warning only if there is no zim-file currently requested

### LINKED ISSUE #995: The warning for "You have an actively scheduled ZIM" appears too early
When requesting a ZIM, and setting up a schedule, after clicking the confirmation button, the page "reloads" with a message about status updates. However, the warning message about having a scheduled ZIM is present and already hiding parts of the page, which is confusing.

Image

We should only show this message if there is no active ZIM being currently created.

nicolapace:
Fixes #995


audiodude (approved):
LGTM

audiodude:
Deploying

PR #1010, opened 2025-09-09
fix recipe payload and routes

elfkuzco:
## Changes
- remove null webhook from recipe payload
- remove leading slashes in urls


audiodude (approved):
LGTM

Issue #491, opened 2022-08-27
Check the return value of s3 object deletes


audiodude:
Currently, s3 object deletes in wp1/logic/selection.py doesn't check the return value of the s3 RPC and always returns true. It would be nice to check the return value, though it seems excessively complicated to do so. Either way, it's not particularly important because it's not clear what we would do if any of them failed besides log it (we don't want to fail the whole deletion process since s3 objects can be cleaned up later).

its-me-maady:
I'm seeing lot of alpha-go issues are solved already but not closed @audiodude

PR #1055, opened 2025-12-13
Check  s3 delete_object response and return failure on errors

rajvi0106:
**DESCRIPTION**

This PR fixes the behavior of  'delete_keys_from_storage' so that it correctly interprets the response returned by S3 'delete_objects' and returns an accurate success status.

Specifically, it preserves the intended behavior of existing tests: when delete_objects returns no response or a response without an Errors key, the operation is treated as successful. Additionally, the function was updated to safely handle cases where the S3 response is not a dictionary, ensuring consistent behavior across all scenarios.

**CHANGES IMPLEMENTED**

1. Parse the 'Errors' field when present in S3 response.
2. Return 'False' where one or more deletion operation fails.
3. Log S3 deletion errors with key,error code, and message.
4. Handle cases where 'delete_objects' returns a non dictionary response.

**TESTS**

All existing tests in selection_test.py now pass.
Added tests for new behavior:
 1. delete_objects returns none.
 2. delete_objects returns non dictionary.

### DIFF
--- a/wp1/logic/selection.py
+++ b/wp1/logic/selection.py
@@ -144,7 +144,7 @@ def delete_keys_from_storage(keys):
     )
 
     fully_successful = True
-    errors = []
+
     # TODO: Check for errors in the response. For now, just pretend
     # it's always successful.
     # for e in (resp if resp is not None else {}).get('Errors', []):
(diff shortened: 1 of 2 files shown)


audiodude:
I believe this is a fix for #491

audiodude (changes requested):
Thanks for submitting a PR!

I looked at this, and it seems it just takes the code from the comment and implements it. However, if I remember correctly (and it's been a long time), the main issue was that I didn't actually know how to properly check for error conditions in the S3 API. The code in the comment was my best guess from looking at sample responses, but I decided to leave it unimplemented and ignore errors rather than create false confidence by implementing incorrect error handling code.

Before we merge this, can you provide an explanation of your understanding of the S3 API or what documents you referenced to land on this solution? I'd like to make sure the error checking is comprehensive and robust, and if not, leave it unimplemented. Thanks!

audiodude (approved):

PR #1057, opened 2025-12-19
Add random article button

### LINKED ISSUE #393: Random article in WikiProject (optionally with given quality/importance)
This was suggested on [English Wikipedia](https://en.wikipedia.org/wiki/Wikipedia_talk:Version_1.0_Editorial_Team/Index#Suggestion:_random_article_from_these) and isn't a bad idea for a feature for the website.

The three inputs would be: WikiProject, Quality, and Importance, with all of them being optional. We then redirect the user to a the requested wikipedia article.

rao107:
This PR adds a button to randomly select an article next to the "Update View" button.

Fixes #393 

I'm not sure if the Cypress test properly tests "randomness" but I also didn't want to create callback hell by clicking on the random button multiple times. Let me know if you have better ideas for more concisely testing the "randomness" in Cypress.

### DIFF
--- a/wp1/logic/rating.py
+++ b/wp1/logic/rating.py
@@ -248,8 +248,8 @@ def get_random_article(
         query += " AND r_importance = %(r_importance)s"
         params["r_importance"] = importance
 
-    # TODO: This is not very efficient since it will generate a random number for all rows that
-    # match the criteria and sort them all.
+    # This is not very efficient since it will generate a random number for all rows that match
+    # the criteria and sort them all.
     query += " ORDER BY RAND() LIMIT 1"
 
     with wp10db.cursor() as cursor:


wp1/logic/rating.py
```diff
@@ -232,6 +232,32 @@ def get_project_rating_by_type(
...
+    params = {"r_project": project_name}
+
+    if quality is not None:
+        query += " AND r_quality = %(r_quality)s"
+        params["r_quality"] = quality
+    if importance is not None:
+        query += " AND r_importance = %(r_importance)s"
+        params["r_importance"] = importance
+
+    query += " ORDER BY RAND() LIMIT 1"
```
audiodude: So there's a whole debate in general about methods of getting random items out of a SQL database, but it's generally understood that `ORDER BY RAND()` is the worst performing option. I believe the database will use the indexes to limit the number of articles based on the project/quality/importance. However it will then generate a random number for each article (let's say there's 500,000 of them) and sort them all, just to return one.

This is probably not a practical problem for us, since we have so little usage. If this button was getting clicked multiple times per minute, it would be more of an issue.

That said, it would probably be good to put a comment acknowledging the issue, potentially with a link to a bug that describes why this works "for now" but to keep an eye out for it.

wp1/web/projects.py
```diff
@@ -184,6 +184,34 @@ def articles(project_name):
     return flask.jsonify(output)
 
 
+@projects.route("/<project_name>/articles/random")
+def random_article(project_name):
```
audiodude: Flask tests for this new method?

PR #1059, opened 2025-12-24
Fix: Can't regenerate failed/expired ZIMs (#1036)

### LINKED ISSUE #1036: Can't create ZIMs from existing selections with expired/failed ZIMs
Image

Previously, we would create a schedule for a ZIM, kick off the task, and then immediately delete the schedule.

After GSoC 2025, we are reusing ZIM schedules. We're supposed to have code that checks if a schedule with the name we want to use exists already. The name is generated from the ID of the Builder and should be stable.

Either that code isn't working, or something went wrong when migrating existing ZIM selections/builders.

ziaddevv:
The bug had some minor issues :

Schedule existence check was using the wrong ID (s_id instead of builder_id) causing the code to think schedules didn't exist when they actually did , This led to 409 Conflict errors when trying to create "new" schedules

after fixing the ID issue , I hit a 405 Method Not Allowed error because the PATCH request was going to /v2/schedules instead of /v2/schedules/{schedule_name} . I have investigated The Zimfarm API requires the schedule name in the URL path 

after fixing both API issues the request was made and the ZIM was created but download URLs were still broken because the task_id wasn't being updated properly when selection versions changed, and the version pointer (b_selection_zim_version) still pointed to the old selection

Changes made:

zimfarm.py : Use builder_id instead of s_id for schedule checks, and fix PATCH URL to include schedule name in the path
builder.py: Add three layer update logic (try current selection - - > try by schedule_id --> insert new) and update b_selection_zim_version to point to current selection when requesting ZIMs

I have tested this locally , by reproducing the same issue , it worked fine , i have successfully regenerated the failed ZIMs and verified the download links works fine 
 
I tried to explain the issue well , please review it , maybe I found the causes of the issue , but actually I'm not sure if this is the correct way to solve it   
fixes #1036

### DIFF
--- a/wp1/logic/builder_test.py
+++ b/wp1/logic/builder_test.py
@@ -273,6 +273,48 @@ class BuilderTest(BaseWpOneDbTest):
                 )
         self.wp10db.commit()
 
+    def _setup_failed_zim_regeneration_scenario(
+        self,
+        zim_schedule_id=b"schedule-123",
+        old_task_id="old-task-id",
+        create_new_selection=False,
+    ):
+        self._insert_zim_schedule(zim_schedule_id, self.builder.b_id)
+
+        with self.wp10db.cursor() as cursor:
+            # Insert selection v1 with failed ZIM
+            cursor.execute(
+                """INSERT INTO selections
+                (s_id, s_builder_id, s_updated_at, s_content_type, s_version,
+                 s_object_key, s_article_count)
+                VALUES (%s, %s, '20250102000000', 'text/tab-separated-values', 1,
+                        'old.tsv', 100)""",
+                (1, self.builder.b_id),
+            )
+            cursor.execute(
+                """INSERT INTO zim_tasks
+                (z_selection_id, z_zim_schedule_id, z_status, z_task_id)
+                VALUES (%s, %s, 'FAILED', %s)""",
+                (1, zim_schedule_id, old_task_id),
+            )
+            # Optionally create selection v2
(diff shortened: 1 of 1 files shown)


audiodude:
I don't have time to review the code right now, I only read the description. But I wanted to leave immediate feedback to tell you that, based on your description, I really appreciate the problem solving and effort that went into this PR. And I apologize for the state of the code that caused this bug.

I _am_ interested in how the tests passed for this deficient code, but I expect that they were also deficient in the sense that they were testing the incorrect behavior instead of the behavior we actually desired. Please make sure to run and update tests as necessary.

Again, thank you so much! I will review this soon.

audiodude (changes requested):
This logic seems sound, so good work. All that's left is to add tests for the 3 new cases (selection version has changed, new zim_task id required, update b_selection_zim_version).

While you're in there, check if the existing tests are even still useful, and be sure to delete or amend as necessary.

Thanks!

ziaddevv:
> This logic seems sound, so good work. All that's left is to add tests for the 3 new cases (selection version has changed, new zim_task id required, update b_selection_zim_version).
> 
> While you're in there, check if the existing tests are even still useful, and be sure to delete or amend as necessary.
> 
> Thanks!

Apologies for the delay on this; I was unavailable last week.

I've added the tests now, and they cover all the cases as requested. I checked the codebase for existing or related tests to reference but couldn't find any, so I implemented these to ensure full coverage of the new logic.

Issue #1069, opened 2026-01-31
Builder API should return builder ID, to be more usefull


audiodude:
https://github.com/openzim/wp1/blob/main/wp1/web/builders.py#L80-L88

It currently just returns a JSON object with `"success": true` but it would be more useful to potential API clients to return the ID.

PR #1071, opened 2026-02-02
return builder_id in create/update response

lumpy72006:
this addresses  #1069


lumpy72006:
I've done that now 🙂

audiodude (approved):
Let me know whether you'd like to fix the last nit. If not I'll run CI.

PR #1080, opened 2026-02-22
Fix #819: Handle expired ZIM file download gracefully

### LINKED ISSUE #819: Unable to download zim

harshavardhanreddy06:
This commit addresses the issue where an expired ZIM file download throws an S3 NoSuchKey error. The backend /zim/latest endpoint now returns a 410 Gone status when the file is expired (older than 2 weeks). The frontend ZimFile component intercepts this by fetching the URL with redirect: 'manual' and provides a user-friendly error instructing the user to re-create their ZIM.

### DIFF
--- a/wp1-frontend/src/components/ZimFile.vue
+++ b/wp1-frontend/src/components/ZimFile.vue
@@ -30,7 +30,7 @@
               You already have a ZIM scheduled for every
               <strong
                 >{{ activeSchedule.interval_months }} month{{
-                  activeSchedule.interval_months > 1 ? 's' : ''
+                  activeSchedule.interval_months > 1 ? "s" : ""
                 }}</strong
               >.
               <span v-if="activeSchedule.next_generation_date">
@@ -201,7 +201,7 @@
                 {{
                   displayGraphemeLimitText(
                     longDescription,
-                    maxLongDescriptionLength
+                    maxLongDescriptionLength,
                   )
                 }}
               </small>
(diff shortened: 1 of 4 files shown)


wp1-frontend/src/components/ZimFile.vue
```diff
@@ -524,6 +529,44 @@ export default {
...
+      // deleted from storage) before the browser falls through to S3's
+      // confusing NoSuchKey error page.
+      this.downloading = true;
+      this.downloadError = null;
+      try {
+        const response = await fetch(this.zimPathFor(), {
+          redirect: "manual",
+          credentials: "include",
+        });
+        if (response.type === "opaqueredirect") {
```
audiodude: Why does this have to be handled separately from the default 200 path?

wp1/logic/builder.py
```diff
@@ -512,6 +512,22 @@ def zim_file_for_latest_selection(wp10db, builder_id) -> ZimTask:
...
 
+def is_zim_file_expired(wp10db, builder_id):
+    """Returns True if the ZIM file for the given builder has expired (been deleted from storage).
+
+    ZIM files are deleted from storage after ZIM_FILE_TTL seconds (2 weeks).
+    This is a time-based check; it does not verify actual S3 existence.
+    """
+    zim = latest_zim_file_for(wp10db, builder_id)
+    if zim is None or zim.z_status != b"FILE_READY":
+        return False
```
audiodude: This logic is incorrect. The ZIM could have a non-`FILE_READY` status and still be processing, not necessarily expired.

wp1-frontend/src/components/ZimFile.vue
```diff
@@ -524,6 +529,44 @@ export default {
...
+          // The server returned a 302 redirect — let the browser navigate.
+          window.location.href = this.zimPathFor();
```
audiodude: I don't think this will work. You are manually handling redirect, but the `.../zim/latest` endpoint inherently returns a redirect. Shouldn't you retrieve and query the actual wasabi endpoint?

audiodude:
In fact, looking at my own comments in your PR, I see the real bug is:

```
                # Older than 2 weeks are deleted.
                data["z_is_deleted"] = logic_selection.is_zim_file_deleted(
                    data["z_updated_at"]
                )
```

Which is incorrect, because as stated the update time is not necessarily the time when the ZIM file itself was created.

Issue #101, opened 2019-09-04
Update error, duplicate key issue


audiodude:
When updating project 'Football', we got the following stack trace:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/rq/worker.py", line 822, in perform_job
    rv = job.perform()
  File "/usr/local/lib/python3.7/site-packages/rq/job.py", line 605, in perform
    self._result = self._execute()
  File "/usr/local/lib/python3.7/site-packages/rq/job.py", line 611, in _execute
    return self.func(*self.args, **self.kwargs)
  File "./wp1/logic/project.py", line 68, in update_project_by_name
    update_project(wikidb, wp10db, project)
  File "./wp1/logic/project.py", line 468, in update_project
    extra_assessments['extra'])
  File "./wp1/logic/project.py", line 247, in update_project_assessments
    process_unseen_articles(wikidb, wp10db, project, old_ratings, seen)
  File "./wp1/logic/project.py", line 357, in process_unseen_articles
    move_data['timestamp_dt'])
  File "./wp1/logic/page.py", line 53, in update_page_moved
    logic_move.insert(wp10db, new_move)
  File "./wp1/logic/move.py", line 34, in insert
    ''', attr.asdict(move))
  File "/usr/local/lib/python3.7/site-packages/pymysql/cursors.py", line 170, in execute
    result = self._query(query)
  File "/usr/local/lib/python3.7/site-packages/pymysql/cursors.py", line 328, in _query
    conn.query(q)
  File "/usr/local/lib/python3.7/site-packages/pymysql/connections.py", line 517, in query
    self._affected_rows = self._read_query_result(unbuffered=unbuffered)
  File "/usr/local/lib/python3.7/site-packages/pymysql/connections.py", line 732, in _read_query_result
    result.read()
  File "/usr/local/lib/python3.7/site-packages/pymysql/connections.py", line 1075, in read
    first_packet = self.connection._read_packet()
  File "/usr/local/lib/python3.7/site-packages/pymysql/connections.py", line 684, in _read_packet
    packet.check_error()
  File "/usr/local/lib/python3.7/site-packages/pymysql/protocol.py", line 220, in check_error
    err.raise_mysql_exception(self._data)
  File "/usr/local/lib/python3.7/site-packages/pymysql/err.py", line 109, in raise_mysql_exception
    raise errorclass(errno, errval)
[...]

PR #1098, opened 2026-02-28
added duplicate clause in move.py

arnchlmcodes:
Added an ON DUPLICATE KEY UPDATE clause to move.py which updates m_new_namespace and m_new_article

as discussed in #101

### DIFF
--- a/wp1/logic/selection.py
+++ b/wp1/logic/selection.py
@@ -179,13 +179,12 @@ def update_zimfarm_task(wp10db, task_id, status, set_updated_now=False):
     with wp10db.cursor() as cursor:
         if set_updated_now:
             updated_at = utcnow().strftime(TS_FORMAT_WP10).encode("utf-8")
-            with wp10db.cursor() as cursor:
-                cursor.execute(
-                    """UPDATE zim_tasks SET z_status = %s, z_updated_at = %s
+            cursor.execute(
+                """UPDATE zim_tasks SET z_status = %s, z_updated_at = %s
                WHERE z_task_id = %s""",
-                    (status, updated_at, task_id),
-                )
-                found = bool(cursor.rowcount)
+                (status, updated_at, task_id),
+            )
+            found = bool(cursor.rowcount)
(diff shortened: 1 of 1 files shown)


wp1/logic/move_test.py
```diff
@@ -0,0 +1,54 @@
...
+        self.move1 = Move(
+            m_timestamp=b"20200101000000\x00\x00\x00\x00\x00\x00",
+            m_old_namespace=0,
+            m_old_article=b"Old_Article_1",
+            m_new_namespace=0,
+            m_new_article=b"New_Article_1",
+        )
+
+    def test_insert_duplicate_key_update(self):
+        logic_move.insert(self.wp10db, self.move1)
```
audiodude: In general, it is best practice to avoid using codebase code in tests, besides the specific code being tested.

This is because of the possibility that there is a bug in `logic_move.insert` which causes this test to fail. It would then be hard to determine if the bug is in insert itself, or in the code being tested by this function.

This is more of a problem with something like `create_or_update_builder`. We almost _never_ want to use that function to create builders in the db for tests. Instead we prefer something like https://github.com/openzim/wp1/blob/main/wp1/logic/builder_test.py#L165.

Imagine if there was a bug in `create_or_update_builder`, which has 1 test for itself, but is used (from this perspective, incorrectly) in 50 other tests. If we accidentally introduce a bug in `create_or_update_builder`, now we have 51 failing tests to sort through and it's not clear where or what is causing the failures. If we isolate the other tests from that method, however, the theoretical bug now causes 1 precise test failure.

The tradeoff is that if and when the schema changes, we have 50 more places to update the tests.

All that to say: please insert the initial data using SQL statements directly. Since this is the only test in this file, there's no need to write a `_insert_move` helper, though you are welcome to.

audiodude (changes requested):
THanks for the test, definitely the right idea!

Unfortunately you have accidentally made more work for yourself, but the end result wil be much stronger.

PR #1102, opened 2026-02-28
Stage 1: Add dotenv-based config system with backward-compatible credentials adapter

### LINKED ISSUE #1016: Replace `credentials.py` config system with a more robust solution, like `dotenv`
The ad-hoc config system that we currently have relies on each deployment (development, development-workers, production, production-workers, CI and maybe some I'm missing) to have their own copies of a `credentials.py` file in the python source directory/module tree. Additionally, within this system there is a further "split" between PRODUCTION, DEVELOPMENT. and TEST environments. This means that keys are duplicated over and over, and it's unclear which ones are necessary or need to be edited for each environment.

This system creates friction when adding new services or dependencies that need to be configured. It's also error prone when updating workers or CI. Finally, it is nearly impossible to explain succinctly to potential new developers who wish to work on the project.

I propose a migration to `dotenv` (https://pypi.org/project/python-dotenv/), which is more of an industry standard solution. We can use `--env-file .env` as an argument to our development docker containers. In CI, we can just load a .env file as a set of standard environment variables. Even in testing, we can manually configure the enviornment by loading a specific `.env.test` file in the Pytest runtime, to avoid [this awful hack](https://github.com/openzim/wp1/blob/main/wp1/credentials.py.example#L21).

ziaddevv:
**Summary**

Replace the scattered credentials.py configuration system with a centralized dotenv-based approach.
This PR introduces the foundation (config.py + .env files) and a backward-compatible adapter so all existing consumers continue working without changes

**Changes**

New: wp1/config.py

Central Config class that reads settings from environment variables via python-dotenv. Includes type-safe helpers (_getenv, _getenv_int, _getenv_list), 
auto-detects pytest and loads .env.test, and validates WP1_ENV values with clear error messages

New: .env.example, .env.test, .env.docker

`.env.example`
Developer setup template that combines all previous credential example files into one organized file with clear sections and setup options.

Comments were rewritten to be more accurate and actionable rather than copied blindly from the multiple old files (needs more review)

`.env.test` 
 Test overrides auto-loaded when pytest is detected. All keys explicitly overridden to prevent leakage from Pipenv’s .env auto-load

`.env.docker`  

Docker hostname overrides for example  minio, redis, dev-database instead of localhost

Modified: wp1/credentials.py (un-gitignored)

now it is a backward-compatible adapter importing from config.py and reconstructing the CREDENTIALS[ENV] nested dict expected by consumers. Since it no longer contains secrets, it is now committed

**Migration Plan**
This is PR 1 of the dotenv migration . No consumer files are changed in this PR . the adapter ensures full backward compatibility.

fixes #1016


wp1/config.py
```diff
@@ -0,0 +1,164 @@
...
+
+from wp1.environment import Environment
+
+# Load .env from project root (parent of wp1/ directory).
+# When running under pytest, load .env.test instead so that WP1_ENV=test
+# is set automatically — matching the old credentials.py.e2e behaviour.
+# We check both 'pytest' in sys.modules AND the PYTEST_CURRENT_TEST env var
+# because config.py may be imported before pytest registers in sys.modules.
+_PROJECT_ROOT = Path(__file__).resolve().parent.parent
+_is_testing = "pytest" in sys.modules or "PYTEST_CURRENT_TEST" in os.environ
```
audiodude: Is this even needed anymore? Can't we just do this on the command line:

```
WP1_ENV=test pipenv run pytest
```

wp1/config.py
```diff
@@ -0,0 +1,164 @@
...
+            )
+        return default
+    return value
+
+
+def _getenv_int(key, *, default=None, required=False):
+    value = _getenv(key, default=default, required=required)
+    if value is None:
+        return None
+    return int(value)
```
audiodude: We should check for `ValueError` and provide a helpful error message.

.env.example
```diff
@@ -0,0 +1,171 @@
...
+# Secret key for Flask session encryption. If this ever changes, all currently
+# logged-in users will be logged out (their session cookies become invalid).
+# To generate a secure random key for production:
+#   python3 -c "import os; print(os.urandom(24).hex())"
+SESSION_SECRET_KEY=WP1_secret_key    # EDIT: set a random secret in production
+
+
+# --- Client URLs ---
+# URLs used for redirections, CORS, and internal service communication.
+CLIENT_DOMAINS=http://localhost:5173
```
audiodude, line 107: nit: Document that this is a comma-separated list. Also do so for any other potential lists.

audiodude (approved):

wp1/config.py
ziaddevv: > Is this even needed anymore? Can't we just do this on the command line:
> 
> ```
> WP1_ENV=test pipenv run pytest
> ```


Actually Pipenv calls load_dotenv internally .That means it **overwrites** any environment variables that were set from the command line.

So `WP1_ENV=test pipenv run pytest`
unfortunately doesn’t work as expected:

Pipenv loads the .env file (which contains WP1_ENV=development) and overrides the command-line value before Python even starts.

So in order to make it happen those are the flag we can use 

- using `PIPENV_DOTENV_LOCATION=.env.test` every time ([check this](https://pipenv.pypa.io/en/latest/shell.html#environment-specific-configuration))
- Using `PIPENV_DONT_LOAD_ENV=1 WP1_ENV=test ` to disable Pipenv’s dotenv loading completely 


The auto approach (_is_testing) avoids all of this it still works  for any developer running pipenv run pytest without needing to know about Pipenv's behavior.

What do you think?

wp1/config.py
ziaddevv: Done ,  added a try/except that catches ValueError and raises a RuntimeError with the key name and the invalid value in the message

ziaddevv:
Hi @audiodude  , I’ve added tests for the config file and addressed your review comments. I’m still waiting for your input on whether we should handle test mode automatically using or rely on manually setting the environment.

wp1/config.py
audiodude: You're right! The trick is to pass the ENV variable to the python process, not the pipenv process:

```
tmoney@tmoney-linux:~/code/pipenv-dotenv-test$ cat Pipfile
[[source]]
url = "https://pypi.org/simple"
verify_ssl = true
name = "pypi"

[packages]
dotenv = "*"

[dev-packages]

[requires]
python_version = "3.12"
tmoney@tmoney-linux:~/code/pipenv-dotenv-test$ cat .env 
WP1_ENV=production
DB_HOST=localhost
tmoney@tmoney-linux:~/code/pipenv-dotenv-test$ cat main.py 
import os
import dotenv

dotenv.load_dotenv()

print(f"WP1_ENV={os.environ.get('WP1_ENV')}")
print(f"DB_HOST={os.environ.get('DB_HOST')}")
tmoney@tmoney-linux:~/code/pipenv-dotenv-test$ pipenv run python main.py
Loading .env environment variables...
WP1_ENV=production
DB_HOST=localhost
tmoney@tmoney-linux:~/code/pipenv-dotenv-test$ WP1_ENV=test pipenv run python main.py
Loading .env environment variables...
WP1_ENV=production
DB_HOST=localhost
tmoney@tmoney-linux:~/code/pipenv-dotenv-test$ pipenv run WP1_ENV=test python main.py 
Loading .env environment variables...
WP1_ENV=test
DB_HOST=localhost
tmoney@tmoney-linux:~/code/pipenv-dotenv-test$
```

audiodude: Automatic seems better, or at least easier. The problem is that it's "magic code". It's secret configuration. The value is silently changed in runtime code, and because developers aren't passing it, they don't see or unerstand where and how it's set. What do you think?

audiodude (changes requested):

wp1/config.py
ziaddevv: Nice find, yes it works with just an environment check in config.py , but I've just tried the exact format  `pipenv run WP1_ENV=test pytest` it fails
`Error: the command WP1_ENV=test could not be found within PATH or Pipfile's [scripts].`

What Pipenv version and shell are you using? I'm on 2025.1.1 with zsh.
The only format that works for me is `pipenv run bash -c 'WP1_ENV=test pytest'`, which actually feels pretty clunky for daily use.

that's why I'm still leaning towards keeping the  auto-detection . regarding requiring devs to remember that wrapper every time risks forgetting and tests silently using the dev config
I know it is a little "magic" behind , we could add a log message plus clear docs in the code  to make it transparent
personally `pipenv run pytest` is way easier to use and remember

audiodude: > What Pipenv version and shell are you using? I'm on 2025.1.1 with zsh. The only format that works for me is `pipenv run bash -c 'WP1_ENV=test pytest'`, which actually feels pretty clunky for daily use.

```
tmoney@tmoney-linux:~/code/wp1$ pipenv --version
pipenv, version 2026.5.1
tmoney@tmoney-linux:~/code/wp1$ $SHELL --version
GNU bash, version 5.3.9(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2025 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

It's probably a bash/zsh thing that I don't have much interest in debugging.
 
> that's why I'm still leaning towards keeping the auto-detection . regarding requiring devs to remember that wrapper every time risks forgetting and tests silently using the dev config I know it is a little "magic" behind , we could add a log message plus clear docs in the code to make it transparent personally `pipenv run pytest` is way easier to use and remember

We already have: https://github.com/openzim/wp1/blob/main/wp1/base_db_test.py#L64

What other information from the dev environment would be a disaster to use during tests?

Can the tests just use a ./run_tests.sh script, in bash, that does it my way. It could handle any other arcane environment munging or setup we might need in the future? Hell, it could do the docker compose test.yml as well right? It could even check if the docker graph was running before doing _that_ .

Are we too focused on one specific way of getting this to work? I'm not married to `WP1_ENV=`, but I really would like to remove automatic test environment detection as part of this work.

ziaddevv: I rechecked the existing test suite , it is actually fairly well protected so running on dev won't hurt 

If we decided to remove the auto detection , I checked those solutions and they work fine 

first just by using a bash wrapper as you suggested all we need is to  `exports WP1_ENV=test `and `PIPENV_DONT_LOAD_ENV=1 `before calling `pipenv run pytest "$@"`
and it works fine and also as you mentioned  it can do some extra work beyond the env setting 
also I found this cool plugin [pytest-env](https://github.com/pytest-dev/pytest-env)
It's basically a tiny wrapper that just adds a line like  `WP1_ENV=test` in pytest.ini, and it handles the rest .It sets the environment variable very early, before any test modules are imported.
Adds one small dev dependency, but `pipenv run pytest `works perfectly.

and even using conftest.py directly works ,using the official pytest hook API, it runs before collection/imports. Zero new dependencies. but I can't assure it is reliable or not 

I found this relevant discussion where the same problem was discussed [pytest](https://github.com/pytest-dev/pytest/discussions/10027) 

So as far as I get , if we want to just set the env test pytest-env is fair solution and if we want to go further than that we can make our own bash script .

ziaddevv: And yes a bash wrapper can do some extra work
- Check that db test containers are running before tests start with a clear error message if they're not
- Start/stop test containers automatically 
- Clean dirty tables from interrupted test runs (I face this a lot) 


image

ziaddevv: Also I have just figured out why the command `pipenv run WP1_ENV=test pytest` didn't work for me , it was the pipenv version not a bash/zsh .The old version (2025.1.1) didn't support inline env vars , it treated WP1_ENV=test as a literal command name .After upgrading to Pipenv 2026.5.2 worked as expected as they fixed the issue

audiodude: Okay let's merge this with a update to the README to make sure your pipenv is up to date and run `pipenv run WP1_ENV=test pytest`. Can you file an issue to make a ./run_tests.sh in a follow up? I really like the idea of running docker and cleaning the dev db.

BTW I added this no-op test for "resetting" the DB after you've done ^C on the tests: https://github.com/openzim/wp1/blob/main/wp1/base_db_test.py#L176

```
$ pipenv run pytest wp1/base_db_test.py
```

.env.example
audiodude: I don't see the update.

wp1/config.py
audiodude: Feel free to resolve this conversation once you've done those tasks.

audiodude (changes requested):

audiodude (approved):

.env.example
ziaddevv: I actually added it in the second commit, along with tests for `config.py`

audiodude:
Awesome, very nice and clean. Thank you!

audiodude:
I resolved a merger error in Pipfile, but it looks like we need to regenerate Pipfile.lock. Can you do that real quick and I'll merge?

ziaddevv:
> Awesome, very nice and clean. Thank you!

Thank you,  I really appreciate it. It wouldn’t have been possible without your guidance and help.

Issue #1100, opened 2026-02-28
[Enhancement] Refactoring the z_status for zim_tasks table to include DELETED flag


Best2Two:
Why are we treating the file’s deletion status on the serverless side as detached from the database? This introduces inconsistency and adds extra logic throughout the codebase. For example, we now have conditions like isDeleted && status == 'FILE_READY'. Shouldn’t we refactor this to include a DELETED status in the database, especially if the record is going to remain in the database after it's deletion anyway?

PR #1103, opened 2026-03-03
Refactor: Implement DELETED status in frontend to resolve inconsistency (#1100)

arybhatt4533:
Description:
This PR addresses the issue where the ZIM file deletion status was detached from the database, causing logic inconsistencies in the frontend.

Key Changes:

Status Integration: Replaced the fragile isDeleted && status == 'FILE_READY' condition with a robust status === 'DELETED' check.

UI Updates: Added a specific user-friendly error message for deleted files, informing users that links expire after 2 weeks.

Workflow Fix: Simplified the v-if conditions to ensure the "Request ZIM" form correctly reappears when a file is marked as DELETED in the database.

Code Cleanup: Removed redundant variables to align with the "Single Source of Truth" approach suggested in #1100.

Fixes: #1100

### DIFF
--- a/docker/zimfarm/create_worker.sh
+++ b/docker/zimfarm/create_worker.sh
@@ -53,7 +53,6 @@ curl -s -X 'POST' \
   -d '{
   "role":"worker",
   "username": "test_worker",
-  "email":"test_worker@acme.com",
   "password":"test_worker"
 }'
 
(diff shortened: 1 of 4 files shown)


audiodude:
This is unnecessary, per the resolution of #1100, and is the wrong approach anyways. The files referred to in `scores.py` are cache files and have nothing to do with ZIMs created by users.

PR #1107, opened 2026-03-06
Fix stale DB redirects appearing as new assessments in quality logs

### LINKED ISSUE #462: Bot marking articles from 2016 as "new" in quality logs
Message from https://en.wikipedia.org/wiki/Wikipedia_talk:Version_1.0_Editorial_Team/Index :

> This has been cropping up for a few weeks, but was particularly bad this week. As an example:

> On the March 25 run, the bot [added](https://en.wikipedia.org/w/index.php?title=Wikipedia:Version_1.0_Editorial_Team/Video_game_articles_by_quality_log&diff=1079297341&oldid=1079106457&diffmode=source) [Harvest Moon: The Tale of Two Towns](https://en.wikipedia.org/wiki/Harvest_Moon:_The_Tale_of_Two_Towns) to the quality log page, saying it was a new Stub.
This is odd, because the article was created and got a video games talk page template in 2010, was last edited in September 2021, the last action on the talk page was a page move in 2016.
This is consistent, though, because that same day the bot also listed e.g. [HyperZone](https://en.wikipedia.org/wiki/HyperZone), [Team Dragon Knights](https://en.wikipedia.org/wiki/Team_Dragon_Knights), [Tempo Storm](https://en.wikipedia.org/wiki/Tempo_Storm), and [Ty the Tasmanian Tiger 3: Night of the Quinkan](https://en.wikipedia.org/wiki/Ty_the_Tasmanian_Tiger_3:_Night_of_the_Quinkan) - all of which received no edits and have been tagged for years, but their last talk page action was a 2016 page move. And not just articles, but templates too: {{[Free Radical Design](https://en.wikipedia.org/wiki/Template:Free_Radical_Design)}}, last talk page action a 2016 page move.

> I don't know it it's actually a bug in the bot or if the underlying data is throwing in a bunch of stale stuff from 6 years ago, but it's very strange. --[Pres](https://en.wikipedia.org/wiki/User:PresN)[N](https://en.wikipedia.org/wiki/User_talk:PresN) 13:46, 28 March 2022 (UTC)

arnchlmcodes:
page_touched was being parsed into timestamp_dt directly, overwriting the project's last run cutoff before any comparison could be made. Now it Stores the parsed value in row_timestamp_dt and return None if not newer than last run

Fixes #462

### DIFF
--- a/wp1/logic/page.py
+++ b/wp1/logic/page.py
@@ -4,7 +4,7 @@ from datetime import datetime
 import requests
 
 import wp1.logic.util as logic_util
-from wp1.constants import GLOBAL_TIMESTAMP, TS_FORMAT
+from wp1.constants import GLOBAL_TIMESTAMP, TS_FORMAT, TS_FORMAT_WP10
 from wp1.logic import log as logic_log
 from wp1.logic import move as logic_move
 from wp1.logic.api import page as api_page
@@ -90,7 +90,7 @@ def _get_redirects_from_db(wikidb, namespace, title, timestamp_dt):
     args_dict = {
         "title": wiki_db_title,
         "namespace": namespace,
-        "timestamp": timestamp_dt.strftime("%Y%m%d%H%M%S")
+        "timestamp": timestamp_dt.strftime(TS_FORMAT_WP10)
     }
     with wikidb.cursor() as cursor:
         cursor.execute(
@@ -108,7 +108,7 @@ def _get_redirects_from_db(wikidb, namespace, title, timestamp_dt):
                 "dest_ns": row["rd_namespace"],
                 "dest_title": row["rd_title"],
                 "timestamp_dt": datetime.strptime(
-                    row["page_touched"].decode("utf-8"), "%Y%m%d%H%M%S"
+                    row["page_touched"].decode("utf-8"), TS_FORMAT_WP10
                 ),
             }
         return None
--- a/wp1/logic/page.py
+++ b/wp1/logic/page.py
@@ -99,13 +99,15 @@ def _get_redirects_from_db(wikidb, namespace, title, timestamp_dt):
         )
         row = cursor.fetchone()
         if row:
-            timestamp_dt = datetime.strptime(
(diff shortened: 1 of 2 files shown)


audiodude (changes requested):
Shouldn't you just add a comparison between `page_touched` and `timestamp_dt` to the SQL query itself? Why try to compare after the row has already been retrieved?

wp1/logic/page.py
```diff
@@ -87,29 +87,32 @@ def update_page_moved(
 def _get_redirects_from_db(wikidb, namespace, title, timestamp_dt):
     wiki_db_title = title.decode("utf-8").replace(" ", "_")
     wikidb.ping()
-    args_dict = {"title": wiki_db_title, "namespace": namespace}
+    args_dict = {
+        "title": wiki_db_title,
+        "namespace": namespace,
+        "timestamp": timestamp_dt.strftime("%Y%m%d%H%M%S")
```
audiodude: You should not use a string literal for the timestamp format.

As you may have noticed, WP1 has very strange, unorthodox, and confusing ways of storing timestamps in MariaDB. It comes down to the fact that:

1) Timestamps are stored as strings, in a format such that string comparison is equivalent to comparison of  corresponding TIMESTAMP types.
2) The WP1 application database, often referenced with the variable `wp10db` uses a different such string fromat than the Wikipedia replica database (often referenced as `wikidb`).

The formatting constants are stored in constants.py: https://github.com/openzim/wp1/blob/main/wp1/constants.py#L16-L17

See an example here:
https://github.com/openzim/wp1/blob/main/wp1/models/wp10/log.py

Not that the "WP1" model of log stores one of its timestamp fields in the WP1 format, because it is native data generated by our application, and the other in the Wikipedia replica format, because it is data taken from the replica db.

Since you are querying the replica db here (`wikidb`), it seems you have in fact introduced a bug, because the formatting you chose looks more like the WP1 format. But please confirm.

audiodude (changes requested):

audiodude (approved):

Issue #1017, opened 2025-09-24
Use Python type hinting


audiodude:
It's 2025 and the Python world has now embraced type hinting. I won't lay out the case for this practice in general, because plenty has been written about it.

For WP1, the most immediate benefits I see are:

1. Knowing when you're working with `b'bytes'` from the DB (as opposed to a `'string'`, because the DB stores only bytes for legacy reasons
2. Knowing when you're dealing with a UUID versus a "regular" string.

audiodude:
@benoit74 @rgaudin Would love to get your opinions on this. I think it would definitely be beneficial, but might be a lot of work. But I have done typing migrations in the past (Javascript -> Typescript) and I've actually found that you can find some bugs when you do this work.

Not sure if it's worth the effort though.

benoit74:
There is definitely a huge added value, this is our default requirement for new projet. 

Regarding whether if it is worth the effort, it is a bit hard to say since it depends on the effort which depends on current WP1 code.

We've done the migration of python-scraperlib for instance and some parts have required very significant changes but even that was deemed worth the effort (and I'm still sure it is) because it allows to trap so many mistakes and errors, especially as code evolves.

kelson42:
I can only encourage you to implement this issue!

audiodude:
@abhiraj75 that sounds good. I especially encourage you to work incrementally as you suggested. I'm not sure if you can use a type checking tool like mypy or PyCharm on a partially migrated codebase, so that would be something to look into.

audiodude:
On the other hand, now that I think of it, this can be done in one massive PR because type hinting doesn't affect runtime behavior right? The only problem with that is based on how long it takes it could have merge conflicts with things we add along the way.

rgaudin:
Type checkers can be given files or folders to restrict its scope.

We use pyright on all other project so it might be wise to go with it

abhiraj75:
@rgaudin @audiodude  I'll start with a module and open a PR soon , will make sure the type hints are pyright compatible.

abhiraj75:
@rgaudin @benoit74 Do give it a look once and let me know if you have any suggestions regarding the approach. Thanks!

abhiraj75:
@audiodude Do give the PR a look and provide any suggestions if you have. Thanks!

PR #1109, opened 2026-03-08
Add type hints to db and constants modules

abhiraj75:
Adds type annotations to wp1/db.py and wp1/constants.py as part of the incremental type hinting effort (#1017).

Changes:

wp1/db.py: Added type hints for connect() function signature, module-level variables, and local variables. 

wp1/constants.py: Added type hints for all module-level constants, including bytes annotations for GLOBAL_TIMESTAMP and GLOBAL_TIMESTAMP_WIKI, and dict[str | bytes, str] for CONTENT_TYPE_TO_EXT.

### DIFF
--- a/wp1/constants.py
+++ b/wp1/constants.py
@@ -17,8 +17,8 @@ TS_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
 TS_FORMAT_WP10 = "%Y%m%d%H%M%S"
 LOG_DATE_FORMAT = "%B %-d, %Y"
 
-GLOBAL_TIMESTAMP = time.strftime(TS_FORMAT_WP10, time.gmtime()).encode("utf-8")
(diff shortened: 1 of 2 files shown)


audiodude (commented):
Thanks for the PR, but I feel like most of these are superfluous.

Can't mypy (or whatever) figure out that:

```
ZIM_FILE_TTL: int = 2 * 7 * 24 * 60 * 60  # 2 weeks
```

Is an int? Won't we already get a type failure if we try to reassign a string to it? I don't know the answer, I'm genuinely asking.

I can definitely see the usefulness for complex types like dicts:

```
CONTENT_TYPE_TO_EXT: dict[str | bytes, str] = {
    "text/tab-separated-values": "tsv",
    b"text/tab-separated-values": "tsv",
    "application/vnd.ms-excel": "xls",
    b"application/vnd.ms-excel": "xls",
}
```

Since "anything" can go into a dict and we want to preserve a strict mapping.

audiodude (approved):

PR #1124, opened 2026-03-22
Added Cache-Control headers for CSS/JS assets (#241)

### LINKED ISSUE #241: Css/js resources should be cached in browser
proper Http response header should be set

its-me-maady:
Added cache.conf to nginx config to serve CSS and JS assets with
proper Cache-Control headers, allowing browsers to cache them
for 1 year. Follows same pattern as existing gzip.conf.

Fixes #241


audiodude:
> That said, happy to shorten to 1w or 1m if you prefer a more conservative approach regardless.

Thanks for the explanation, that makes sense. Since the file names are unique to each release, we could theoretically cache indefinitely, with no expiration, right? 

Either way this is fine, thanks!

Issue #72, opened 2019-08-22
Logs not properly read/written when article gets moved


audiodude:
For context, see the discussion [on English Wikipedia](https://en.wikipedia.org/wiki/Wikipedia_talk:Version_1.0_Editorial_Team/Index#Logs_will_be_back_soon!)

Basically, when a page gets moved, the new page gets written to the logs as moved, but then the next day it appears as a "new" page which has "just been assessed". This could either be a problem with writing the logs (aka the actual logging table in wp10), or a problem with reading them and generating the logging page.

kelson42:
@audiodude What is the status on this ticket? Do we have a strategy to implement?

audiodude:
@kelson42 It needs investigation, there's no root cause identified currently.

kelson42:
@walkerma @audiodude Does someone is in position to confirm this is still an occuring problem?

PR #1126, opened 2026-03-23
Fix spurious new-assessment log entries for renamed articles

arnchlmcodes:
Heyy @audiodude During update_project_assessments_by_kind, any article whose ref is absent from old_ratings gets treated as new. The problem is that a renamed article will always look new at this stage because the old name is what's stored in old_ratings, not the new one. Previously, store_new_ratings would immediately call add_log_for_rating for these, writing a spurious log entry with old_rating_value defaulting to NotA-Class.

The fix defers those log entries into a list instead of writing them immediately. Only articles that already existed in old_ratings but had a rating change get logged right away, since an existing ref changing its rating cannot be a rename.

process_unseen_articles runs after both quality and importance passes complete. For each old ref that wasn't seen, it calls get_move_data() to check if the article was renamed. If a move is detected, the destination ref is added to a moved_articles set. Back in update_project_assessments, the deferred logs are then written only for refs not present in moved_articles, ensuring that renamed articles don't get a spurious new-assessment log entry while genuinely new articles are logged correctly.

fixes issue #72

let me know if the fix is correct, I'll start working on the test cases next.


audiodude (approved):
This looks great! Please add the tests now.

arnchlmcodes:
hey @audiodude , sorry for taking time with the tests, the coverage was tricky and kind of confusing hence the delay,
I've added 3 test classes

StoreNewRatingsTest checks the branch logic in store_new_ratings directly  new refs defer their log, existing changed refs log immediately, unchanged refs do nothing. Also asserts the rating is persisted even when the log is deferred since that wasn't obvious from reading the code.

ProcessUnseenArticlesMovedTest covers the new return value from process_unseen_articles (it used to return None). Had to add a dummy article to seen just to get past the early-return guard. Also explicitly assert that get_move_data and update_page_moved were called, otherwise the test passes vacuously if the mock path is never hit.

DeferredLogMovedArticleTest is the end-to-end regression check for the original bug, renamed article shouldn't produce a spurious new-assessment log. Covers three scenarios: pure rename, purely new articles, and a mix of both.

audiodude (approved):
Great work, these tests look good!

PR #1128, opened 2026-03-24
fix time calculations and redundant save in logs.py

Best2Two:
fixes #1127

### DIFF
--- a/wp1/logs.py
+++ b/wp1/logs.py
@@ -87,8 +87,9 @@ def calculate_logs_to_update(redis, project_name, from_dt=None):
     those that are newer than that date. Otherwise, the date of the most recently
     uploaded log, as determined by querying the wiki replica, is used.
     """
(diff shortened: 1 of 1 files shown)


db/migrations/20250330_01_zaRQa-migrate-last-7-days-of-logs.py
```diff
@@ -22,7 +22,7 @@ def migrate_logs(conn):
     redis = redis_connect()
 
     from_dt = get_current_datetime() - timedelta(days=7)
-    from_dt.replace(hour=23, minute=59, second=59)
+    from_dt = from_dt.replace(hour=23, minute=59, second=59)
```
audiodude: As I've said before, the migrations are immutable. Even though this might be considered a bug, it should not be changed.

audiodude (changes requested):

audiodude (approved):

Issue #1117, opened 2026-03-13
Allow for customization of selections


Popolechien:
A nice feature to add would be the possibility to allow users to generate a ZIM file based off existing flavours (ie, `nopic`, `mini` and `maxi`).

kashif7230:
I am still getting familiar with the codebase, i can help test the integration once the PR is ready.

"This is a great addition for ZIM customization. I've been following the progress on the flavor selection logic; once the UI in ZimFile.vue and the API passing are drafted, I’d be happy to help test the various flavor combinations to ensure the parameters are being passed correctly to the offliner."

audiodude:
Fixed by #1133

PR #1133, opened 2026-03-25
Add zim flavour selection for content format choice

### LINKED ISSUE #919: can I make nopic zimfile with wp1 ?
can I make nopic zimfile with wp1 ?
If it doesn't. could this be done?

ziaddevv:
Allow users to choose a content flavour when creating a ZIM file, controlling what content is included (nopic,novid, lead sections , mini,maxi , nopdf).

changes:

Add flavour selection support across the system (database , backend , Zimfarm integration)
Add a dropdown in the UI to choose the desired flavour when creating a ZIM

here is some outputs 
full
image

nopic
image

fixes #919 , #1117

### DIFF
--- a/db/migrations/20260324_01_add_flavour_to_zim_schedules.py
+++ b/db/migrations/20260324_01_add_flavour_to_zim_schedules.py
@@ -1,0 +1,14 @@
+"""
+Add flavour column to zim_schedules table for ZIM format selection
+"""
+
+from yoyo import step
+
+__depends__ = {"20250821_01_add_email_confirmation_token_to_zim_schedules"}
+
+steps = [
+    step(
+        "ALTER TABLE zim_schedules " "  ADD COLUMN s_flavour VARBINARY(50) NULL",
+        "ALTER TABLE zim_schedules " "  DROP COLUMN s_flavour",
+    )
+]
--- a/wp1-frontend/src/components/ZimFile.vue
+++ b/wp1-frontend/src/components/ZimFile.vue
@@ -106,6 +106,17 @@
             this was a transient or external error, you can try requesting your
             ZIM file again below.
           </p>
+          <p
+            v-if="
+              flavour &&
+              (status === 'FILE_READY' ||
+                status === 'REQUESTED' ||
+                status === 'FAILED')
+            "
+            class="text-muted"
+          >
+            <strong>Content:</strong> {{ selectedFlavourDescription }}
+          </p>
         </div>
       </div>
       <div
@@ -212,6 +223,18 @@
               <div class="invalid-feedback">
                 Long description must differ from description and not be shorter
               </div>
+            </div>
+            <div id="flavour-group" class="form-group">
+              <label for="flavour">What to include</label>
+              <select id="flavour" class="form-control" v-model="flavour">
+                <option
+                  v-for="opt in flavourOptions"
+                  :key="opt.value"
+                  :value="opt.value"
+                >
+                  {{ opt.description }}
+                </option>
+              </select>
             </div>
             <!-- Scheduling options -->
             <div class="form-group form-check mt-3">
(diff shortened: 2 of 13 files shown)


audiodude:
> hi @audiodude sorry for pinging but is that copilot thing configured on the repo or I request it or what?

Looks like you requested it: 
image

db/migrations/20260324_01_add_flavour_to_zim_schedules.py
```diff
@@ -0,0 +1,14 @@
...
+Add flavour column to zim_schedules table for ZIM format selection
+"""
+
+from yoyo import step
+
+__depends__ = {"20250821_01_add_email_confirmation_token_to_zim_schedules"}
+
+steps = [
+    step(
+        "ALTER TABLE zim_schedules " "  ADD COLUMN s_flavour VARCHAR(50) NULL",
```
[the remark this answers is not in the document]
audiodude: Please fix. We do not use `VARCHAR`, only `VARBINARY`.

I believe this comes from a time when Unicode support was less reliable and the original creators of the database schema didn't "trust" the MySQL collation.

wp1-frontend/src/components/ZimFile.vue
audiodude, line 1: Let's not include every single option, they are confusing and probably mostly unnecessary.

@Popolechien: which options should we include?

Also I think we should hide the enums completely, and the `label` should be what you currently have as the `description`. No one knows or cares what `maxi` is besides ZIM nerds.

wp1/logic/builder.py
```diff
@@ -694,7 +701,7 @@ def handle_zim_generation(
             )
 
     zim_schedule = zimfarm.create_or_update_zimfarm_schedule(
-        redis, wp10db, builder, title, description, long_description
+        redis, wp10db, builder, title, description, long_description, flavour=flavour
```
audiodude: This should not be "optional", see https://github.com/openzim/wp1/pull/982#issue-3346272210

wp1/zimfarm.py
```diff
@@ -233,6 +256,28 @@ def _get_params(
         )
     image_name, image_tag = image.split(":")
 
+    offliner_config = {
```
audiodude, line 254: Nice refactor!

wp1-frontend/src/components/ZimFile.vue
```diff
@@ -213,6 +224,21 @@
                 Long description must differ from description and not be shorter
               </div>
             </div>
+            <div id="flavour-group" class="form-group">
+              <label for="flavour">Flavour</label>
```
audiodude: I don't think we want to refer to this in the UI as "Flavour". I would just call it "What to include". It is not at all intuitive what a flavour of a ZIM would constitute (even though that's what it's called in the API).

audiodude (changes requested):
Good work. A few UI tweaks, but this is in really good shape!

wp1-frontend/src/components/ZimFile.vue
benoit74: > besides ZIM nerds

I would be even more restrictive: "besides mwoffliner ZIM nerds" 🤣 

I'm kinda surprised we have all these values in mwoffliner, looks like I've oversight something in mwoffliner/zimfarm reviews ; I've now proposed to restrict the list to values we've really use in production and use clearer labels: https://github.com/openzim/mwoffliner/pull/2687 ; probably this is the list you wanna use (and then you can directly fetch and display values from zimfarm APIs).

ziaddevv:
Thanks @audiodude @benoit74.  I've applied all the requested changes .Really appreciate your feedback

wp1-frontend/src/components/ZimFile.vue
ziaddevv: > > besides ZIM nerds
> 
> I would be even more restrictive: "besides mwoffliner ZIM nerds" 🤣
> 
> I'm kinda surprised we have all these values in mwoffliner, looks like I've oversight something in mwoffliner/zimfarm reviews ; I've now proposed to restrict the list to values we've really use in production and use clearer labels: [openzim/mwoffliner#2687](https://github.com/openzim/mwoffliner/pull/2687) ; probably this is the list you wanna use (and then you can directly fetch and display values from zimfarm APIs).

Yeah, I agree :laughing:  Initially I checked mwoffliner as well and noticed there are quite a lot of variations there . I even spent some time trying to understand what each one does , thanks for helping

audiodude (approved):

PR #1141, opened 2026-03-31
change zimfarm schedules routes to recipes

elfkuzco:
## Rationale
This PR replaces calls to the zimfarm schedules routes with the `recipes` route

This closes #1132


benoit74 (commented):
@audiodude as I suspected, the "wrong" term "schedule" has already leaked to WP1... this is "wrong" because it is not the business term chosen by Kiwix (and is a bit confusing technically speaking). Do you want to take this opportunity to fix this in the whole WP1 codebase or is what @elfkuzco done sufficient for the time being?

PR #1142, opened 2026-03-31
Fix Redis Token Decoding for Zimfarm Authentication

dkconnect:
So i was going through this file and I felt this is a big missing part here. Please have a look

Problem
1. Incorrect Token Lookup
- Redis returns data as "bytes" in many clients
- Code was accessing using "str" keys ("refresh_token")
- so the Result: token was never found

2.. Repeated Unnecessary Authentication Calls
- Because "refresh_token" was never detected, the system will continue requesting new tokens This will mean:
- Extra API calls
- Inefficient performance

3. Runtime Errors 
- Direct ".decode()" could fail if Redis already returned strings

### DIFF
--- a/wp1/zimfarm.py
+++ b/wp1/zimfarm.py
@@ -109,8 +109,13 @@ def refresh_zimfarm_token(redis, refresh_token):
 
 
 def get_zimfarm_token(redis):
-    data = redis.hgetall(REDIS_AUTH_KEY)
-    if data is None or data.get("refresh_token") is None:
+    data = redis.hgetall(REDIS_AUTH_KEY) or {}
+    data = {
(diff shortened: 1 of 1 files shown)


kelson42:
@dkconnect Thank you for your PR. It will be reviewed by a maintainer within a few days.

dkconnect:
Hey @kelson42 @benoit74 I dont what am I missing here

do i need to do something different to pass the "Continuous Integration / black-format-check" test case

audiodude:
Just run `black .` to format all files with [black](https://github.com/psf/black) in place.

I will do a proper review soon.

audiodude:
Preempted by #1146

PR #1145, opened 2026-04-02
Add setup/seed.py for WP1 database seeding

### LINKED ISSUE #262: Seed the Arabic WP1 database
We have [schema.sql](https://github.com/openzim/wp1/blob/main/setup/schema.sql) which I believe does work to seed a wp1 database. However, it is missing some of the data that is required for the bot to run, like the namespace mappings and global class weights.

We should create a seed.py script that can run to seed the database from schema.sql and some optional data source the specifies the namespaces and class weights (the latter probably comes from conf.json).

Best2Two:
fixes #262 
added setup/seed.py to seed a WP1 database. It creates the tables from schema.sql, seeds global_rankings from conf/<lang>/conf.json, and fetches namespace mappings from the Wikimedia API into namespacename. The database is created automatically if it doesn't exist. Tested against a fresh Arabic WP1 database and got 27 rankings and 61 namespace rows.

it's language agnostic though so you can actually populate whatever table you want by changing the --lang option, i believe this requires a documentation but not sure where to put.

image
image

### DIFF
--- a/setup/seed.py
+++ b/setup/seed.py
@@ -1,0 +1,320 @@
+#!/usr/bin/env python3
+"""Seed the WP1 database with schema, global rankings, and namespace data.
+
+This script performs three seeding steps (each can be skipped individually):
+
+  1. Apply setup/schema.sql to create the required database tables.
+  2. Seed ``global_rankings`` from the QUALITY/IMPORTANCE maps in
+     ``conf/<lang>/conf.json``.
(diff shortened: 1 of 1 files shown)


Best2Two:
@audiodude hello, kindly I need a review here

audiodude:
I don't think this is a sound approach, see comment.

PR #1146, opened 2026-04-03
auth via oauth with zimfarm

### LINKED ISSUE #1144: Add capability to authenticate with zimfarm based on OAuth
Companion issue to https://github.com/openzim/zimfarm/issues/1803

@elfkuzco will work on this since he knows well how to do it ^^

elfkuzco:
## Rationale
This PR enhances WP1 to optionally authenticate with Zimfarm using ouath (via Ory) instead of just username and password.

## Changes
- create token provider class that generates oauth tokens based on specified auth mode
- directly raise `ZimfarmError` when no token can be generated instead of delegating it to callers

This closes #1144

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -154,12 +154,21 @@ services:
       JWT_SECRET: DH8kSxcflUVfNRdkEiJJCn2dOOKI3qfw
       POSTGRES_URI: postgresql+psycopg://zimfarm:zimpass@zimfarm-db:5432/zimfarm
       ALEMBIC_UPGRADE_HEAD_ON_START: true
+      DEBUG: true
       INIT_USERNAME: admin
       INIT_PASSWORD: admin
       ALLOWED_ORIGINS: http://localhost:8003
       ARTIFACTS_UPLOAD_URI: s3+http://minio:9000/?keyId=minio_key&secretAccessKey=minio_secret&bucketName=org-kiwix-dev-artifacts
       LOGS_UPLOAD_URI: s3+http://minio:9000/?keyId=minio_key&secretAccessKey=minio_secret&bucketName=org-kiwix-dev-logs
       ZIM_UPLOAD_URI: s3+http://minio:9000/?keyId=minio_key&secretAccessKey=minio_secret&bucketName=org-kiwix-dev-zims
+      OAUTH_JWKS_URI: https://ory.login-staging.kiwix.org/.well-known/jwks.json
+      OAUTH_ISSUER: https://ory.login-staging.kiwix.org
+      OAUTH_OIDC_CLIENT_ID: 38302485-7f3d-4e88-b1b3-ec02fb92ec8c
+      OAUTH_OIDC_LOGIN_REQUIRE_2FA: true
+      OAUTH_SESSION_AUDIENCE_ID: 309693e7-ad5e-4379-bf93-ba89314230fd
+      OAUTH_SESSION_LOGIN_REQUIRE_2FA: true
+      CREATE_NEW_OAUTH_ACCOUNT: false
+      AUTH_MODES: local,oauth-session
     networks:
       - wp1bot-dev
     depends_on:
--- a/docker/zimfarm/zimfarm_ui_dev/config.json
+++ b/docker/zimfarm/zimfarm_ui_dev/config.json
@@ -1,3 +1,6 @@
 {
-  "ZIMFARM_WEBAPI": "http://localhost:8004/v2"
+  "ZIMFARM_WEBAPI": "http://localhost:8004/v2",
+  "OAUTH_BASE_URL": "https://ory.login-staging.kiwix.org",
+  "OAUTH_MODE": "session",
+  "LOGIN_MODES": ["local"]
 }
--- a/wp1/zimfarm.py
+++ b/wp1/zimfarm.py
@@ -1,10 +1,12 @@
 import logging
 import urllib.parse
 import uuid
-from datetime import UTC, datetime
+from typing import cast, Any
+from datetime import datetime, UTC, timedelta
 
 import regex
 import requests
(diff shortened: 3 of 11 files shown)


elfkuzco:
@audiodude , do you think we should remove storing the tokens via Redis given there's always going to be one instance of the class and it maintains state. There could be restarts of the application or multiple instances but they can all generate newer tokens and continue to function independently as generating a new token doesn't invalidate the other.

benoit74:
@audiodude shall we merge and deploy this soon? WP1 seems to be the last system using "old" username/password auth to Zimfarm (cannot say for sure, since all I see is logs using "old" API endpoints).

audiodude:
Yes I generally feel good about the concept and am eager to merge. However, I do want to do a detailed code review first and I haven't had the time. Mostly just to make sure I understand the changes and the new flow.

wp1/zimfarm.py
```diff
@@ -45,87 +46,158 @@ def store_zimfarm_token(redis, data):
...
-        json={"username": user, "password": password},
-    )
-    try:
-        r.raise_for_status()
-    except requests.exceptions.HTTPError as e:
-        logger.exception(r.text)
-        raise ZimFarmError("Error getting authentication token for Zimfarm") from e
+def getnow():
+    """naive UTC now"""
+    return datetime.now(UTC).replace(tzinfo=None)
```
audiodude: We've already got something like this:

https://github.com/openzim/wp1/blob/main/wp1/timestamp.py

But I've been getting warnings about it forever. Can you move your implementation to that location and use it?

wp1/zimfarm.py
```diff
@@ -45,87 +46,158 @@ def store_zimfarm_token(redis, data):
...
+
+
+class ZimfarmClientTokenProvider:
+    """Client to generate access tokens to authenticate with Zimfarm API"""
+
+    def __init__(self):
+        self._access_token: str | None = None
+        self._refresh_token: str | None = None
+        self._expires_at: datetime = datetime.fromtimestamp(0, UTC).replace(tzinfo=None)
+        self._zimfarm_creds: dict[str, Any] = CREDENTIALS[ENV].get("ZIMFARM", {})
```
audiodude, line 74: I don't see corresponding changes to the credentials.*.example.py fles. Please add.

docker/zimfarm/zimfarm_ui_dev/config.json
```diff
@@ -1,3 +1,6 @@
 {
-  "ZIMFARM_WEBAPI": "http://localhost:8004/v2"
+  "ZIMFARM_WEBAPI": "http://localhost:8004/v2",
+  "OAUTH_BASE_URL": "https://ory.login-staging.kiwix.org",
```
audiodude, line 3: We're not actually hitting this in development right? it's just a URL for signing a JWT or something?

docker-compose-dev.yml
```diff
@@ -154,12 +154,21 @@ services:
...
       POSTGRES_URI: postgresql+psycopg://zimfarm:zimpass@zimfarm-db:5432/zimfarm
       ALEMBIC_UPGRADE_HEAD_ON_START: true
+      DEBUG: true
       INIT_USERNAME: admin
       INIT_PASSWORD: admin
       ALLOWED_ORIGINS: http://localhost:8003
       ARTIFACTS_UPLOAD_URI: s3+http://minio:9000/?keyId=minio_key&secretAccessKey=minio_secret&bucketName=org-kiwix-dev-artifacts
       LOGS_UPLOAD_URI: s3+http://minio:9000/?keyId=minio_key&secretAccessKey=minio_secret&bucketName=org-kiwix-dev-logs
       ZIM_UPLOAD_URI: s3+http://minio:9000/?keyId=minio_key&secretAccessKey=minio_secret&bucketName=org-kiwix-dev-zims
+      OAUTH_JWKS_URI: https://ory.login-staging.kiwix.org/.well-known/jwks.json
```
audiodude, line 164: Same question here, are we hitting live URLs? If so, why?

audiodude:
> @audiodude , do you think we should remove storing the tokens via Redis given there's always going to be one instance of the class and it maintains state. There could be restarts of the application or multiple instances but they can all generate newer tokens and continue to function independently as generating a new token doesn't invalidate the other.

That's certainly a valid design. But this PR already has it continuing to use Redis, and we need Redis for other parts of the system, so I think it's fine to leave (though you could add an issue if you'd like for the future).

audiodude (changes requested):
This looks great, thank you so much!

just a few comments.

docker/zimfarm/zimfarm_ui_dev/config.json
elfkuzco: No, default auth_mode should be `local`

```py
        if self._zimfarm_creds.get("auth_mode", "local") == "local":
            if not (
                self._zimfarm_creds.get("user") and self._zimfarm_creds.get("password")
            ):
```

Unless dev sets the `auth_mode` to `oauth`. Just added the variables there for completeness

docker-compose-dev.yml
elfkuzco: No. Just added them for completeness. Default should be local

wp1/zimfarm.py
elfkuzco: Added.

benoit74:
@audiodude did you got the answers you need and shall we merge? Should we setup a meeting so that we answer quickly and precisely to all your points in a more interactive manner?

We really wanna merge this soon to close the current SSO project phase.

audiodude (approved):

audiodude:
Thanks for the poke, merging now

benoit74:
@elfkuzco can you have a quick look at failing tests?

elfkuzco:
> @elfkuzco can you have a quick look at failing tests?

fixed

Issue #1009, opened 2025-09-09
Update to Python 3.14


audiodude:
Python 3.14 has few if any breaking changes and potentially offers [up to 30% performance increase](https://docs.python.org/3.14/whatsnew/3.14.html#a-new-type-of-interpreter).

We should upgrade when we get a chance. It should just be a matter of bumping the versions in the Docker images and double checking that tests pass, and smoke testing the website.

rgaudin:
Yes to upgrading to 3.14 when it's released. Depending on your other python dependencies, that might not be possible immediately and you might have to upgrade only to 3.13 for a while.

As for using free-threaded python, I'm not in favor. I doubt it would have visible beneficial impact but the potential for issues is huge and given there's an effort to upgrade a lot of things, that sound like a bad timing.

audiodude:
> [@audiodude](https://github.com/audiodude) I will try working on that Should i update different packages and commit them in a single PR or should i make them atomic?

I think we should upgrade individual packages in separate PRs, to help with isolation if any issues arise. Of course, it would be best if we could upgrade package py-foo directly to the "final" version that is going to be used with whatever latest version of Python we end up on, whether it's 3.13 or 3.14

> Also, do you see any other dependencies that must be updated other than those listed above?

My preference is to not update _any_ dependency that is not strictly required for the Python upgrade.

its-me-maady:
Hi, I'd like to contribute to this. @Best2Two are you still actively working on the dependency upgrades, or is this open for someone else to pick up? I don't want to duplicate efforts.

its-me-maady:
@audiodude No response from @Best2Two — is it okay for me to work on?

audiodude:
Sure, we don't assign issues. I've changed the title and goal of the issue though. 

After following this discussion, I think it's more important to upgrade the specific packages we use than to move to Python 3.14. That will put us in a better place to upgrade Python later anyways.

its-me-maady:
@audiodude could you review #1147 when you get a chance? Both the oauthlib and yoyo-migrations upgrades are blocked on this since they also require pytest ~=9.0 — if I open those PRs now they'll have conflicting Pipfile.lock changes. Merging this first keeps the remaining PRs clean. Thanks!

audiodude:
#1147 is in the merge queue, sorry for the delay.

its-me-maady:
Hi @audiodude PR #1154 is open for review. Take a look when you get a chance

its-me-maady:
[...]

PR #1162, opened 2026-05-03
chore: upgrade Python from 3.12 to 3.14 in Docker images

its-me-maady:
Update all Dockerfiles and Pipfile to use Python 3.14. Part of #1009.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,10 +15,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up Python 3.12.0
+      - name: Set up Python 3.14.4
         uses: actions/setup-python@v5
         with:
-          python-version: '3.12.0'
+          python-version: '3.14.4'
 
       - name: Install Black
         run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -15,10 +15,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up Python 3.12.0
+      - name: Set up Python 3.14
```
audiodude: I think it's best to put the patch version, like we had, to avoid surprise version bumps on the Docker end. Otherwise LGTM

audiodude (changes requested):

its-me-maady:
Updated the Python version to 3.14.4 (latest patch release) across all Dockerfiles and the CI workflow, as requested.

audiodude (approved):

audiodude:
Thank you! In the future you can put `fixes #1234` in your PR description and it will auto close the issue, which we prefer.

its-me-maady:
Got it, thanks for the tip! I was unsure if the issue would need further work beyond the three dependency PRs, so I played it safe with 'Part of' — but I'll use 'Fixes' when I'm confident the PR completes the goal.

PR #1173, opened 2026-05-24
Add type hints to selection builders

ziaddevv:
add types to 

- wp1/selection/abstract_builder.py
- wp1/selection/models/simple.py
- wp1/selection/models/petscan.py
- wp1/selection/models/sparql.py
- wp1/selection/models/wikiproject.py
- wp1/selection/models/book.py

### DIFF
--- a/wp1/selection/models/book.py
+++ b/wp1/selection/models/book.py
@@ -1,5 +1,6 @@
 import logging
-import urllib
+import urllib.parse
+from typing import Any
 
 import mwparserfromhell
 import requests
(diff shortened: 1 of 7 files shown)


wp1/selection/models/book.py
```diff
@@ -59,7 +60,9 @@ def build(self, content_type, **params):
 
         return "\n".join(titles).encode("utf-8")
 
-    def validate(self, **params):
+    def validate(
+        self, **params: Any
+    ) -> tuple[list[str] | str, list[str] | str, list[str]]:
```
audiodude, line 65: I don't know if I mentioned this before, but typing not only gives us confidence when we write new code, it also highlights junk like this, where we have an opportunity to tighten up/refacto the signatures.

Though in this case, isn't `| str, list[str] | str, list[str]]` redundant? Here and below.

audiodude (changes requested):

audiodude (approved):

PR #1176, opened 2026-05-29
 Add combinator selection builder core functions

ziaddevv:
Adds a new combinator selection builder that combines existing materialized builder selections using set operations.

  ## Changes

  - Add `combinator.py`
  - Add `combinator_test.py`
  - Support include/exclude groups with `union` and `intersection`
  - Read referenced builders' latest TSV selections from storage
  - Normalize article titles before combining
  - Produce deterministic sorted TSV output
  
  - Add save-time validation for:
    - required include group
    - valid operations
    - referenced builder existence
    - same user
    - same project
    - no meataBuilder-to-metaBuilder references
  - Add combinator tests

 Build-time checks are limited to materialization state, such as missing selections, failed selections, missing object keys, S3 download failures and empty final results.

### DIFF
--- a/wp1/selection/models/combinator.py
+++ b/wp1/selection/models/combinator.py
@@ -16,20 +16,17 @@ from wp1.selection.abstract_builder import AbstractBuilder
 logger = logging.getLogger(__name__)
 
 
-# as  abstract Metabuilder should be added later
-# some functions will be moved to it
+# TODO: move shared metabuilder helpers into a metaBuilder base class when that abstraction is introduced
 
 META_BUILDER_MODELS = {"wp1.selection.models.combinator"}
 
 
-# metabuilder
 def _as_text(value: bytes | str | int | None) -> str:
     if isinstance(value, bytes):
         return value.decode("utf-8")
     return str(value)
 
 
-# metabuilder
 def _builder_label(builder: Wp10Builder) -> str:
     name = getattr(builder, "b_name", None)
     builder_id = _as_text(getattr(builder, "b_id", ""))
@@ -38,18 +35,14 @@ def _builder_label(builder: Wp10Builder) -> str:
     return builder_id
 
 
-# metabuilder
 def _builder_model(builder: Wp10Builder) -> str:
     return _as_text(getattr(builder, "b_model", ""))
 
 
-# metabuilder
 def _is_meta_builder(builder: Wp10Builder) -> bool:
     return _builder_model(builder) in META_BUILDER_MODELS
 
 
-# metabuilder
-# just to preserve order
 def _dedupe(builder_ids: list[str]) -> list[str]:
     seen: set[str] = set()
     unique_ids: list[str] = []
@@ -72,25 +65,25 @@ def _validate_group_shape(
         return [], errors
 
     if not isinstance(group, dict):
-        return [], [f"The {name} group configuration is invalid"]
+        return [], [f"The {name} group must be a dictionary"]
 
     builder_values = group.get("builders", [])
     if builder_values is None:
         builder_values = []
     if not isinstance(builder_values, list):
-        errors.append(f"The {name} group builder list is invalid")
-        builder_values = []
-
-    builders = [
-        builder_id for builder_id in builder_values if isinstance(builder_id, str)
-    ]
+        return [], [f"The {name} group builder list must be a list"]
 
     if required and not builder_values:
         errors.append(f"Please add at least one builder to the {name} group")
 
-    if len(builders) != len(builder_values) or any(
-        not builder_id for builder_id in builders
-    ):
+    builders: list[str] = []
+    has_invalid_builder_id = False
+    for builder_id in builder_values:
+        if not isinstance(builder_id, str) or not builder_id:
+            has_invalid_builder_id = True
+            continue
+        builders.append(builder_id)
+    if has_invalid_builder_id:
         errors.append(f"The {name} group contains an invalid builder ID")
 
     operation = group.get("operation")
@@ -102,6 +95,7 @@ def _validate_group_shape(
     return builders, errors
 
 
+# TODO: extractt TSV title parsing helpers if more selection builders needs this behavior
 def _normalize(line: bytes) -> str | None:
     s = line.decode("utf-8", errors="replace").strip()
     if not s or s.startswith("#"):
@@ -131,7 +125,6 @@ def _apply_operation(operation: str, sets: list[set[str]]) -> set[str]:
     raise ValueError(f"Unsupported operation: {operation}")
 
 
-# metabuilder
 def _fetch_selection_data(wp10db, s3, builder_id: str) -> bytes:
     """Fetch the latest materialized TSV snapshot for a referenced builder."""
     selection = logic_builder.latest_selection_for(
@@ -144,23 +137,21 @@ def _fetch_selection_data(wp10db, s3, builder_id: str) -> bytes:
             f"(no selection found)"
         )
 
-    status = selection.s_status
-    if status == b"FAILED" or status == "FAILED":
+    status = _as_text(selection.s_status)
+    if status == "FAILED":
         raise Wp1FatalSelectionError(
-            f"Referenced builder {builder_id} has no usable selection "
-            f"(status={status!r})"
+            f"Referenced builder {builder_id} latest selection failed"
         )
 
-    if status != b"OK" and status != "OK":
+    if status != "OK":
         raise Wp1RetryableSelectionError(
-            f"Referenced builder {builder_id} has no usable selection "
+            f"Referenced builder {builder_id} latest selection is not ready "
             f"(status={status!r})"
         )
 
     if selection.s_object_key is None:
         raise Wp1RetryableSelectionError(
-            f"Referenced builder {builder_id} has no usable selection "
-            f"(s_object_key is None)"
+            f"Referenced builder {builder_id} latest selection has no object key"
         )
 
     object_key = selection.s_object_key
@@ -175,11 +166,6 @@ def _fetch_selection_data(wp10db, s3, builder_id: str) -> bytes:
         raise Wp1RetryableSelectionError(
             f"Failed to download selection for builder {builder_id} "
             f"from S3 key {object_key!r}: {code}"
-        ) from e
-    except Exception as e:
-        raise Wp1RetryableSelectionError(
-            f"Failed to download selection for builder {builder_id} "
-            f"from S3 key {object_key!r}: {e}"
         ) from e
 
     return buffer.getvalue()
@@ -225,11 +211,21 @@ class Builder(AbstractBuilder):
         if errors:
             return ([], [], errors)
 
-        if "wp10db" not in params or "user_id" not in params or "project" not in params:
+        missing_params = []
+        if params.get("wp10db") is None:
+            missing_params.append("wp10db")
+        if params.get("user_id") is None:
+            missing_params.append("user_id")
+        if params.get("project") is None:
+            missing_params.append("project")
+        if missing_params:
             return (
                 [],
                 [],
-                ["Configuration could not be validated. Please try saving again."],
+                [
+                    "Missing required validation parameters: "
+                    + ", ".join(missing_params)
+                ],
             )
 
         wp10db = params["wp10db"]
@@ -249,6 +245,7 @@ class Builder(AbstractBuilder):
             try:
                 builder = logic_builder.get_builder(wp10db, reference_id)
             except ObjectNotFoundError:
+                # TODO: https://github.com/openzim/wp1/issues/1178
                 errors.append(
                     f"Builder {reference_id!r} no longer exists. Please remove it "
                     "from this combinator."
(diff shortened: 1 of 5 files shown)


wp1/selection/models/combinator_test.py
```diff
@@ -0,0 +1,331 @@
...
+
+
+def _reference_builder(
+    id_="builder-a",
+    name="Builder A",
+    user_id="1234",
+    project="en.wikipedia.org",
+    model="wp1.selection.models.simple",
+):
+    return SimpleNamespace(
```
audiodude: Why can't we return an actual Builder model?

wp1/selection/models/combinator_test.py
```diff
@@ -0,0 +1,331 @@
...
+            return data[builder_id]
+
+        params = dict(self.params)
+        params.update(
+            include={"builders": ["builder-a", "builder-b"], "operation": "union"},
+            exclude={"builders": ["builder-c"], "operation": "union"},
+            s3=MagicMock(),
+        )
+
+        with patch(
```
audiodude: I've generally tried to keep all the patches at the top method level. Is there a reason you're doing it differently here?

wp1/selection/models/combinator.py
```diff
@@ -0,0 +1,317 @@
...
+    errors: list[str] = []
+
+    # can we have empty groups?
+    if group is None:
+        if required:
+            errors.append(f"Please add at least one builder to the {name} group")
+        return [], errors
+
+    if not isinstance(group, dict):
+        return [], [f"The {name} group configuration is invalid"]
```
audiodude: nit: "...must be a dictionary"

wp1/selection/models/combinator.py
```diff
@@ -0,0 +1,317 @@
...
+
+    builder_values = group.get("builders", [])
+    if builder_values is None:
+        builder_values = []
+    if not isinstance(builder_values, list):
+        errors.append(f"The {name} group builder list is invalid")
```
audiodude: nit: "...must be a list".

In general, it is infuriating when an API returns "Blah is invalid". Tell me WHY it's invalid!

wp1/selection/models/combinator.py
```diff
@@ -0,0 +1,317 @@
...
+        builder_values = []
+
+    builders = [
+        builder_id for builder_id in builder_values if isinstance(builder_id, str)
+    ]
+
+    if required and not builder_values:
+        errors.append(f"Please add at least one builder to the {name} group")
+
+    if len(builders) != len(builder_values) or any(
```
audiodude: I think you can simplify this and just immediately throw if you detect an empty or non-str builder_id.

Makes the intent more obvious.

```
for builder_id in builders:
   if not (builder_id and isinstance(builder_id,` str))):
      raise(...)
```

wp1/selection/models/combinator.py
```diff
@@ -0,0 +1,317 @@
...
+    operation = group.get("operation")
+    if builder_values and operation not in ("union", "intersection"):
+        errors.append(
+            f"Please select a valid operation (union or intersection) for the {name} group"
+        )
+
+    return builders, errors
+
+
+def _normalize(line: bytes) -> str | None:
```
audiodude, line 105: We should think about refactoring a utiltiy method like this. Maybe add a TODO?

wp1/selection/models/combinator.py
```diff
@@ -0,0 +1,317 @@
...
+    s = line.decode("utf-8", errors="replace").strip()
+    if not s or s.startswith("#"):
+        return None
+    return s.replace(" ", "_")
+
+
+def _parse_tsv_to_set(data: bytes) -> set[str]:
```
audiodude, line 112: Same here. Potential future refactor.

wp1/selection/models/combinator.py
```diff
@@ -0,0 +1,317 @@
...
+    titles: set[str] = set()
+    for line in data.split(b"\n"):
+        normalized = _normalize(line)
+        if normalized is not None:
+            titles.add(normalized)
+    return titles
+
+
+def _apply_operation(operation: str, sets: list[set[str]]) -> set[str]:
```
audiodude, line 122: I think WikiProject does something similar right? Maybe we can push all of these methods as _helpers in AbstractBuilder?

wp1/selection/models/combinator.py
```diff
@@ -0,0 +1,317 @@
...
+    )
+
+    if selection is None:
+        raise Wp1RetryableSelectionError(
+            f"Referenced builder {builder_id} has no usable selection "
+            f"(no selection found)"
+        )
+
+    status = selection.s_status
+    if status == b"FAILED" or status == "FAILED":
```
audiodude: Isn't this redundant, unless you give a more specific error message (aka "...is FAILED")

wp1/selection/models/combinator.py
```diff
@@ -0,0 +1,317 @@
...
+            f"(status={status!r})"
+        )
+
+    if status != b"OK" and status != "OK":
+        raise Wp1RetryableSelectionError(
+            f"Referenced builder {builder_id} has no usable selection "
+            f"(status={status!r})"
+        )
+
+    if selection.s_object_key is None:
```
audiodude, line 164: We should think more carefully about if this is possible, or if the object_key is always populated when the selection status is `OK`.

wp1/selection/models/combinator.py
```diff
@@ -0,0 +1,317 @@
...
+    buffer = io.BytesIO()
+    try:
+        s3.download_fileobj(object_key, buffer)
+    except ClientError as e:
+        code = e.response.get("Error", {}).get("Code", "Unknown")
+        raise Wp1RetryableSelectionError(
+            f"Failed to download selection for builder {builder_id} "
+            f"from S3 key {object_key!r}: {code}"
+        ) from e
+    except Exception as e:
```
audiodude: We shouldn't `except Exception` unless absolutely necessary.

wp1/selection/models/combinator.py
```diff
@@ -0,0 +1,317 @@
...
+        errors.extend(group_errors)
+
+        if errors:
+            return ([], [], errors)
+
+        if "wp10db" not in params or "user_id" not in params or "project" not in params:
+            return (
+                [],
+                [],
+                ["Configuration could not be validated. Please try saving again."],
```
audiodude: I don't know if the other builders do this, but it would be good to metnion exactly which param is missing. This will make your life easier when building the frontend.

wp1/selection/models/combinator.py
```diff
@@ -0,0 +1,317 @@
...
+            )
+
+        wp10db = params["wp10db"]
```
audiodude, line 246: Replace the check above with `.get()` calls here, if you want to implement the above suggestion.

wp1/selection/models/combinator.py
```diff
@@ -0,0 +1,317 @@
...
+
+        if errors:
+            return ([], [], errors)
+
+        for reference_id in referenced_ids:
+            try:
+                builder = logic_builder.get_builder(wp10db, reference_id)
+            except ObjectNotFoundError:
+                errors.append(
+                    f"Builder {reference_id!r} no longer exists. Please remove it "
```
audiodude, line 265: Okay for now, but this is the kind of nuance that we will have to think more about later. Please put a TODO referencing a specific issue (that you create).

wp1/selection/models/combinator.py
```diff
@@ -0,0 +1,317 @@
...
+
+    def build(self, content_type: str, **params: Any) -> bytes:
+        """Build the combinator selection from referenced builders."""
+        if content_type != "text/tab-separated-values":
+            raise Wp1FatalSelectionError(f"Unrecognized content type: {content_type!r}")
+
+        wp10db = params.get("wp10db")
+        s3 = params.get("s3")
+
+        if wp10db is None:
```
audiodude, line 303: nit: group these with the get calls that they reference.

wp1/selection/models/combinator.py
```diff
@@ -0,0 +1,317 @@
...
+            raise Wp1FatalSelectionError("Combinator build requires 'wp10db' parameter")
+        if s3 is None:
+            raise Wp1FatalSelectionError("Combinator build requires 's3' parameter")
+
+        include_group = params.get("include")
+        exclude_group = params.get("exclude")
+
+        if include_group is None:
```
audiodude, line 311: nit: same thing, validate it as soon as you retrieve it.

audiodude (changes requested):
Great first attempt, good work! I think most of this is solid, just a few nits and some TODOs for later.

wp1/selection/models/combinator_test.py
ziaddevv: yes we can , switched it

wp1/selection/models/combinator_test.py
ziaddevv: No strong reason ,  i forgot to standardize them

wp1/selection/models/combinator.py
ziaddevv: yes wikiproject has a similar include/exclude set behavior  but still they are a more combinator-specific for now , I'll add a TODO so I can revisit

wp1/selection/models/combinator.py
ziaddevv: yes currently redundant , i made like this so maybe if i wanted to add more handling inside for each  later 
updated now to include the status

wp1/selection/models/combinator.py
audiodude: Okay, add a comment about when this is possible.

wp1/selection/models/combinator.py
```diff
@@ -0,0 +1,314 @@
...
+        exclude_builders, group_errors = _validate_group_shape(
+            "Exclude", exclude, False
+        )
+        errors.extend(group_errors)
+
+        if errors:
+            return ([], [], errors)
+
+        missing_params = []
+        if params.get("wp10db") is None:
```
audiodude, line 230: I don't think validate gets a `wp10db` param.

wp1/selection/models/combinator.py
audiodude: Don't see the TODO.

wp1/selection/models/combinator.py
```diff
@@ -0,0 +1,314 @@
...
+            errors.append("This combinator cannot reference itself")
+
+        if errors:
+            return ([], [], errors)
+
+        for reference_id in referenced_ids:
+            try:
+                builder = logic_builder.get_builder(wp10db, reference_id)
+            except ObjectNotFoundError:
+                # TODO: https://github.com/openzim/wp1/issues/1178
```
audiodude: format as `TODO: #1178 - Think more about this.`

wp1/selection/models/combinator.py
```diff
@@ -0,0 +1,314 @@
...
+    Wp1FatalSelectionError,
+    Wp1RetryableSelectionError,
+)
+from wp1.models.wp10.builder import Builder as Wp10Builder
+from wp1.selection.abstract_builder import AbstractBuilder
+
+logger = logging.getLogger(__name__)
+
+
+# TODO: move shared metabuilder helpers into a metaBuilder base class when that abstraction is introduced
```
audiodude: `TODO: #1181 - Move shared helpers into AbstractBuilder`

wp1/selection/models/combinator.py
ziaddevv: i checked the materialization flow and this case can exist. AbstractBuilder.materialize() only uploads when selection.data is truthy, so yes  an OK selection with empty data could be inserted without an s_object_key. I’ll keep the defensive check for now

wp1/selection/models/combinator.py
ziaddevv: I added a TODO  #1178

wp1/selection/models/combinator.py
ziaddevv: agreed , i removed the broad except Exception and now only catch ClientError

wp1/selection/models/combinator.py
audiodude: Here's another comment that needs addressing.

audiodude (changes requested):
Looks like some of my comments weren't actually sent last time, apologies. It's mostly nits, should only take a few minutes to clean it al lup. Thanks!

wp1/selection/models/combinator.py
audiodude: This if statement doesn't make sense. At first glance, it seems to say "compare status to 'FAILED', whether status is `str` or `bytes`." But if status is `str` the first check will throw an exception (cannot compare str to bytes) and if status is `bytes` and not equal to `b"FAILED"`, the second clause will throw a similar exception.

Please fix.

ziaddevv: actually this is an outdated version , I Fixed it in [refine combinator validation and tests](https://github.com/openzim/wp1/pull/1176/commits/a976a8c09fdada024c0d0ff2f3164958f48c9a39)  already by normalizing selection.s_status with _as_text() before comparing, so the comparisons are now string-to-string , Also python doesn't throw an exception here, it just returns False.

wp1/selection/models/combinator.py
ziaddevv: `_create_or_update_builder()` does pass wp10db into validate(), and both combinator/WikiProject needs it for validation

audiodude (approved):

PR #1186, opened 2026-06-10
Combinator builder frontend

ziaddevv:
Adds the frontend create/edit flow for Combinator selections.

Summary:
- Adds a Combinator builder page for selecting include/exclude builders.
- Supports union/intersection operations for each group.
- Wires Combinator routes and navigation.
- Adds frontend validation 
- Adds Cypress coverage for create, edit, validation, error states, and My Selections navigation.

### DIFF
--- a/wp1-frontend/src/components/CombinatorBuilder.vue
+++ b/wp1-frontend/src/components/CombinatorBuilder.vue
@@ -83,6 +83,13 @@
                 Loading builders...
               </div>
               <div
+                v-else-if="buildersLoadError"
+                class="alert alert-danger mt-2 mb-0"
+                role="alert"
+              >
+                {{ buildersLoadError }}
+              </div>
+              <div
                 v-else-if="
                   availableIncludeBuilders(builder.project).length === 0
                 "
@@ -225,6 +232,7 @@ export default {
   data: function () {
     return {
       allBuilders: [],
+      buildersLoadError: '',
       buildersLoaded: false,
       invalidItems: '',
       includeBuilders: [],
@@ -294,6 +302,8 @@ export default {
           }
         );
         if (!response.ok) {
+          this.buildersLoadError =
+            'Unable to load builders. Please try again later.';
           this.buildersLoaded = true;
           return;
         }
@@ -303,8 +313,11 @@ export default {
           ...builder,
           id: String(builder.id),
         }));
+        this.buildersLoadError = '';
         this.buildersLoaded = true;
       } catch (e) {
+        this.buildersLoadError =
+          'Unable to load builders. Please try again later.';
         this.buildersLoaded = true;
       }
     },
--- a/wp1-frontend/src/components/CombinatorBuilder.vue
+++ b/wp1-frontend/src/components/CombinatorBuilder.vue
@@ -90,12 +90,11 @@
                 {{ buildersLoadError }}
               </div>
               <div
-                v-else-if="
-                  availableIncludeBuilders(builder.project).length === 0
-                "
-                class="text-muted small mt-2"
+                v-else-if="hasNoEligibleBuilders(builder.project)"
+                class="alert alert-warning mt-2 mb-0"
+                role="alert"
               >
-                No builders are available for
+                No eligible builders are available for
                 {{ builder.project }}.
               </div>
             </div>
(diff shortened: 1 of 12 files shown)


wp1-frontend/src/components/CombinatorBuilder.vue
```diff
@@ -0,0 +1,539 @@
...
+      try {
+        const response = await fetch(
+          `${import.meta.env.VITE_API_URL}/selection/simple/lists`,
+          {
+            headers: { 'Content-Type': 'application/json' },
+            credentials: 'include',
+          }
+        );
+        if (!response.ok) {
+          this.buildersLoaded = true;
```
audiodude: Add a TODO here. If we can't fetch builders because of a server error, we need to distinguish that from just "you don't have any eligible builders".

wp1-frontend/src/components/CombinatorBuilder.vue
```diff
@@ -0,0 +1,539 @@
...
+    operationText: function (operation) {
+      return operation === 'intersection' ? ' INTERSECTION ' : ' UNION ';
+    },
+    addInclude: function () {
+      this.addSelectedBuilder('include');
+    },
+    removeInclude: function (builderId) {
+      this.removeSelectedBuilder('include', builderId);
+    },
+    addExclude: function () {
```
audiodude: These are one line functions that are only used once. I feel it would actually be more readable just to inline them, so that we keep the list of functions on the component shorter.

wp1-frontend/src/components/SecondaryNav.vue
audiodude, line 1: This effectively "launches" Combinator builders. Do you want to do this now, or do you think we should wait a bit longer, at least until kelson takes a look?

wp1-frontend/src/components/CombinatorBuilder.vue
audiodude, line 1: Use Claude to add tests before we submit this.

audiodude (changes requested):
This looks really good! I think the main thing from an architecture perspective is thinking through when helper methods in the component make sense, versus when they're distracting. Not every @click handler needs to use a helper, but that also doesn't mean that none of them should!

wp1-frontend/src/components/CombinatorBuilder.vue
ziaddevv: no need for TODO ,added a message to  handle  server error

wp1-frontend/src/components/CombinatorBuilder.vue
ziaddevv: yeah that's true , I made it inline and removed the one line wrappers

wp1-frontend/src/components/SecondaryNav.vue
ziaddevv: Good point. I’d like to keep the nav link for now and wait to merge until Kelson reviews and agrees with exposing it. Since this PR adds the frontend flow, I think it is useful to keep the normal entry point available for review/testing , and I have no problem removing it for now if we need to merge

WDYT?

wp1-frontend/src/components/CombinatorBuilder.vue
```diff
@@ -0,0 +1,540 @@
...
+        this.allBuilders = data.builders.map((builder) => ({
+          ...builder,
+          id: String(builder.id),
+        }));
+        this.buildersLoadError = '';
+        this.buildersLoaded = true;
+      } catch (e) {
+        this.buildersLoadError =
+          'Unable to load builders. Please try again later.';
+        this.buildersLoaded = true;
```
audiodude: This doesn't make sense. There's an error, but `this.buildersLoaded = true`? I think `buildersLoaded` needs a rename.

wp1-frontend/src/components/CombinatorBuilder.vue
audiodude: You need a separate error state for "You have no eligible builders" which is different from the server throwing a 500 or whatever error.

ziaddevv:
I rebased the branch onto latest `origin/main`,removed Combinator from the nav for now and  renamed the loading flag to `buildersFetchFinished`.

audiodude (approved):

wp1-frontend/src/components/SecondaryNav.vue
audiodude: Discussed offline, let's hold off on the nav link. We can still access via URL.

PR #1188, opened 2026-06-11
Call validate() in AbstractBuilder before building

### LINKED ISSUE #1180: AbstractBuilder should call `validate()` before building
If a `Builder` is not valid, we can't build it right?

Therefore, the `AbstractBuilder` should call `validate()` before building/materializing.

`SimpleBuilder`  already validates on `build()` so refactor that out.

audiodude:
AbstractBuilder.materialize now validates the builder params before calling build(), raising Wp1RetryableSelectionError (saved as a CAN_RETRY selection with error messages) when validation fails. The equivalent error check is refactored out of SimpleBuilder.build, which now uses validate() only for list normalization.

Fixes #1180

### DIFF
--- a/wp1/selection/abstract_builder.py
+++ b/wp1/selection/abstract_builder.py
@@ -70,8 +70,7 @@ class AbstractBuilder:
             )
             if invalid or errors:
                 raise Wp1RetryableSelectionError(
-                    "The selection contained invalid parameters: %s"
-                    % "; ".join(errors)
+                    "The selection contained invalid parameters: %s" % "; ".join(errors)
                 )
 
             selection.data = self.build(
--- a/wp1/selection/abstract_builder.py
+++ b/wp1/selection/abstract_builder.py
@@ -61,6 +61,19 @@ class AbstractBuilder:
         )
         selection.set_id()
         try:
+            _, invalid, errors = self.validate(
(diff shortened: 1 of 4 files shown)


wp1/selection/abstract_builder.py
```diff
@@ -61,6 +61,18 @@ def materialize(
         )
         selection.set_id()
         try:
+            _, invalid, errors = self.validate(
```
ziaddevv, line 64: There seems to be a gap when persisted builder params are malformed. validators that assume required params are present/well-shaped can raise raw `KeyError`/`TypeError` before `materialize()` saves a selection with errors.

From what I can tell, Combinator is the only builder that currently handles malformed params defensively in `validate()`. Should each builder’s `validate()` be made similarly defensive for missing/wrong-shaped params, or should `AbstractBuilder.materialize()` catch validation-time exceptions and convert them into a selection error?

audiodude: I don't think that's true, but I'm wondering where you got that impression. Let's discuss.

audiodude: Okay as discussed offline, there remain some inconsistencies with the implementations of `validate()/build()`. This change is safe as is, so let's table any refactors for now.

PR #1195, opened 2026-06-13
Meta builder abstraction

### LINKED ISSUE #1181: Move metabulider helpers out of CombinatorBuilder
Currently, the helpers are direclty in the CombinatorBuilder class. We should move them when that abstraction is introduced

ziaddevv:
Introduces a `MetaBuilder` abstraction for builders that reference other builders.

New structure:
- `wp1/selection/abstract_builder.py`
  - `AbstractBuilder`
  - generic builder lifecycle
  - `_as_text()`

- `wp1/selection/meta_builder.py`
  - `MetaBuilder`
  - `_builder_model()`
  - `_builder_label()`
  - `_reference_label()`
  - `_is_meta_builder()`
  - `_dedupe()`
  - `_fetch_selection_data()`

- `wp1/selection/models/combinator.py`
  - `Builder(MetaBuilder)`
  - Combinator-specific validation/build logic
  - include/exclude group handling
  - TSV-to-set parsing
  - union/intersection/difference behavior

- `wp1/selection/meta_builder_test.py`
  - tests for shared MetaBuilder referenced-selection behavior

- `wp1/selection/models/combinator_test.py`
  - updated to use the new MetaBuilder structure
  
  fixes #1181

### DIFF
--- a/wp1/models/wp10/builder.py
+++ b/wp1/models/wp10/builder.py
@@ -6,6 +6,7 @@ from typing import Any
 
 import attr
 
+from wp1.logic import util as logic_util
 from wp1.constants import TS_FORMAT_WP10
 from wp1.timestamp import utcnow
 
@@ -32,6 +33,32 @@ class Builder:
     b_updated_at: bytes | None = attr.ib(default=None)
     b_current_version: int = attr.ib(default=0)
     b_selection_zim_version: int = attr.ib(default=0)
+
+    @property
+    def id(self) -> str:
+        return logic_util.as_text(self.b_id)
+
+    @property
+    def name(self) -> str:
+        return logic_util.as_text(self.b_name)
+
+    @property
+    def user_id(self) -> str:
+        return logic_util.as_text(self.b_user_id)
+
+    @property
+    def project(self) -> str:
+        return logic_util.as_text(self.b_project)
+
+    @property
+    def model(self) -> str:
+        return logic_util.as_text(self.b_model)
+
+    @property
+    def label(self) -> str:
+        if self.b_name is not None:
+            return f"{self.name} ({self.id})"
+        return self.id
 
     @property
     def created_at_dt(self) -> datetime.datetime:
@@ -75,10 +102,10 @@ class Builder:
 
     def to_web_dict(self) -> dict[str, Any]:
         return {
-            "name": self.b_name.decode("utf-8"),
-            "project": self.b_project.decode("utf-8"),
+            "name": self.name,
+            "project": self.project,
             "params": json.loads(self.b_params.decode("utf-8")),
-            "model": self.b_model.decode("utf-8"),
+            "model": self.model,
         }
 
     def set_id(self) -> None:
--- a/wp1/selection/meta_builder.py
+++ b/wp1/selection/meta_builder.py
@@ -1,0 +1,97 @@
+import io
+
+from botocore.exceptions import ClientError
+
+import wp1.logic.builder as logic_builder
+from wp1.exceptions import (
+    ObjectNotFoundError,
+    Wp1FatalSelectionError,
+    Wp1RetryableSelectionError,
+)
+from wp1.models.wp10.builder import Builder as Wp10Builder
+from wp1.selection.abstract_builder import AbstractBuilder
+
+
+class MetaBuilder(AbstractBuilder):
+    """Base class for builders that reference other builders."""
+
+    META_BUILDER_MODELS = {"wp1.selection.models.combinator"}
+
+    def _builder_model(self, builder: Wp10Builder) -> str:
+        return self._as_text(getattr(builder, "b_model", ""))
+
+    def _builder_label(self, builder: Wp10Builder) -> str:
+        name = getattr(builder, "b_name", None)
+        builder_id = self._as_text(getattr(builder, "b_id", ""))
+        if name is not None:
+            return f"{self._as_text(name)} ({builder_id})"
+        return builder_id
+
+    def _reference_label(self, wp10db, builder_id: str) -> str:
+        try:
+            builder = logic_builder.get_builder(wp10db, builder_id)
+        except ObjectNotFoundError:
+            return builder_id
+        return self._builder_label(builder)
+
+    def _is_meta_builder(self, builder: Wp10Builder) -> bool:
+        return self._builder_model(builder) in self.META_BUILDER_MODELS
+
+    def _dedupe(self, builder_ids: list[str]) -> list[str]:
+        seen: set[str] = set()
+        unique_ids: list[str] = []
+        for builder_id in builder_ids:
+            if builder_id not in seen:
+                seen.add(builder_id)
+                unique_ids.append(builder_id)
+        return unique_ids
+
+    def _fetch_selection_data(
+        self, wp10db, s3, builder_id: str, reference_label: str | None = None
+    ) -> bytes:
+        """Fetch the latest materialized TSV snapshot for a referenced builder."""
+        label = reference_label or builder_id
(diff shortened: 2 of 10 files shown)


ziaddevv:
@audiodude  can you review this ? I feel like i'm being blocked by it  so we can move forward

wp1/selection/meta_builder.py
```diff
@@ -0,0 +1,97 @@
...
+from wp1.models.wp10.builder import Builder as Wp10Builder
+from wp1.selection.abstract_builder import AbstractBuilder
+
+
+class MetaBuilder(AbstractBuilder):
+    """Base class for builders that reference other builders."""
+
+    META_BUILDER_MODELS = {"wp1.selection.models.combinator"}
+
+    def _builder_model(self, builder: Wp10Builder) -> str:
```
audiodude: Shouldn't this just be a property on Builder?

wp1/selection/meta_builder.py
```diff
@@ -0,0 +1,97 @@
...
+        return self._as_text(getattr(builder, "b_model", ""))
+
+    def _builder_label(self, builder: Wp10Builder) -> str:
```
audiodude: Same here. This is really something about a Builder, not anything to do with at MetaBuilder right?

wp1/selection/meta_builder.py
```diff
@@ -0,0 +1,97 @@
...
+        name = getattr(builder, "b_name", None)
+        builder_id = self._as_text(getattr(builder, "b_id", ""))
+        if name is not None:
+            return f"{self._as_text(name)} ({builder_id})"
+        return builder_id
+
+    def _reference_label(self, wp10db, builder_id: str) -> str:
```
audiodude: I would say this is just `builder_label_by_id(wp10db, builder_id: str)` and should be in logic/builders and call the property methods as described above.

wp1/selection/meta_builder.py
```diff
@@ -0,0 +1,97 @@
...
+        try:
+            builder = logic_builder.get_builder(wp10db, builder_id)
+        except ObjectNotFoundError:
+            return builder_id
+        return self._builder_label(builder)
+
+    def _is_meta_builder(self, builder: Wp10Builder) -> bool:
```
audiodude: nit: This probably shouldn't be hidden (_*) because it might be useful in other places.

wp1/selection/meta_builder.py
```diff
@@ -0,0 +1,97 @@
...
+        return self._builder_model(builder) in self.META_BUILDER_MODELS
+
+    def _dedupe(self, builder_ids: list[str]) -> list[str]:
```
audiodude: I don't think this is necessary. Cant you de-dupe a list with just `set(some_list)` or `list(set(some_list))` if you need a list back?

wp1/selection/meta_builder.py
```diff
@@ -0,0 +1,97 @@
...
+        self, wp10db, s3, builder_id: str, reference_label: str | None = None
+    ) -> bytes:
+        """Fetch the latest materialized TSV snapshot for a referenced builder."""
+        label = reference_label or builder_id
+        selection = logic_builder.latest_selection_for(
+            wp10db, builder_id, "text/tab-separated-values"
+        )
+
+        if selection is None:
+            raise Wp1RetryableSelectionError(
```
audiodude, line 28: We're saying this is retryable because the selection might be in process or could be re-triggered? Need to make sure we add logic to actually do that.

wp1/selection/meta_builder.py
```diff
@@ -0,0 +1,97 @@
...
+                f"Referenced builder {label} latest selection is not ready "
+                f"(status={status!r})"
+            )
+
+        # OK selections can have no object key when materialization produced empty
+        # data, since AbstractBuilder only uploads filled selection.data.
+        if selection.s_object_key is None:
+            raise Wp1RetryableSelectionError(
+                f"Referenced builder {label} latest selection has no object key"
```
audiodude: nit: "...no stored data"

audiodude (changes requested):
Sorry for the delay. This is a really good refactor. Most of my comments are getting into the territory of "I'm sure this works, but....". Trying to make sound design decisions and make sure the code is readable and resuable. Thanks!

wp1/selection/meta_builder.py
ziaddevv: yes they are , basically we only needed them in Meta Builders , but no problem for having them as properties in builder

wp1/selection/meta_builder.py
ziaddevv: should we move it to logic/builder also?

wp1/selection/meta_builder.py
ziaddevv: yes but that won't preserve order ,  i'll use `list(dict.fromkeys(builder_ids))`

wp1/selection/meta_builder.py
ziaddevv: yes this is a case we have to handle , I opened #1196 to track that

wp1/models/wp10/builder.py
```diff
@@ -16,6 +16,12 @@ def builder_id() -> bytes:
     return str(uuid.uuid4()).encode("utf-8")
 
 
+def _as_text(value: bytes | str | int | None) -> str:
```
audiodude: nit: This can go in util.py, it's not specific to Builders

wp1/selection/models/combinator.py
```diff
@@ -180,9 +180,12 @@ def _process_group(self, wp10db, s3, group: dict[str, Any]) -> set[str]:
 
         sets: list[set[str]] = []
         # TODO: #1184 - Handle multiple bad referenced selections in combinator build.
-        for builder_id in self._dedupe(builder_ids):
+        for builder_id in dict.fromkeys(builder_ids):
```
audiodude: You don't need to create a `dict` here. Just use `lset(builder_ids)`. You don't even need a `list()` because you can just iterate over the set.

wp1/selection/models/combinator.py
```diff
@@ -124,7 +124,7 @@ def validate(
         project = params["project"]
         builder_id = params.get("builder_id")
 
-        referenced_ids = self._dedupe(include_builders + exclude_builders)
+        referenced_ids = list(dict.fromkeys(include_builders + exclude_builders))
```
audiodude: Same as below, no need for `fromkeys`, just iterate over `referenced_ids = set(include_builders + exclude_builders)`

wp1/selection/models/combinator.py
```diff
@@ -143,20 +143,20 @@ def validate(
                 )
                 continue
 
-            label = self._builder_label(builder)
+            label = builder.label
 
-            if self._as_text(builder.b_user_id) != self._as_text(user_id):
+            if builder.user_id != self._as_text(user_id):
```
audiodude: `_as_text` got moved to builder logic right? So I don't think this is going to work, but I might be missing something.

audiodude (changes requested):
Good work refactoring those methods, I think they make more sense now.

wp1/selection/models/combinator.py
ziaddevv: good point , I noticed I forgot to move it from abtract builder so we had  two version of as_text , now we have one in logic/util

wp1/selection/models/combinator.py
ziaddevv: I don't remember why I initially wanted to keep the order so I made it like this, but for now its no difference and it works fine with just a set()

audiodude (approved):

PR #1204, opened 2026-06-26
handle referenced combinators during list deletion

### LINKED ISSUE #1178: behavior for combinators with deleted referenced builders
Combinators currently fail validation if one of their referenced builders no longer exists

  - Should saving a combinator be blocked until the missing builder is removed?
  - Should the UI show broken references and allow repair
  - Should deleting a builder be blocked if a combinator references it 

current implementation just returns a validation error 

`builder-name  no longer exists. Please remove it from this combinator.`

ziaddevv:
**Problem**

Deleting a list that is used by Combinators could leave broken references or make Combinators invalid when they have no included lists left.

**Summary**
- Adds a delete-impact API endpoint to preview affected Combinators before deleting a list.
- Extends list deletion to accept selected Combinator IDs and optional name confirmation.
- Re-checks Combinator references during final delete so stale preview data is not trusted.
- Deletes selected Combinators and auto-deletes Combinators that would have no included lists left.
- Updates kept Combinators by removing the deleted list reference and rematerialize them.

fixed https://github.com/openzim/wp1/issues/1178

### DIFF
--- a/wp1-frontend/src/components/BaseBuilder.vue
+++ b/wp1-frontend/src/components/BaseBuilder.vue
@@ -160,6 +160,7 @@
                   id="deleteListButton"
                   type="button"
                   class="btn btn-danger ml-4"
+                  :disabled="deleteProcessing"
                 >
                   Delete List
                 </button>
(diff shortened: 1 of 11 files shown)


audiodude:
If there are no combinators depending on the Builder, I don't think we need to confirm anything beyond just "Cancel/Delete". There's nothing to really understand or acknowledge, so just let them delete with a confrim step:

image

EDIT:
So the idea is that in this case only, we only need to confirm deletion the same as we currently do, except with the modal instead of the Javascript alert.

audiodude (commented):
So far so good. I haven't looked at the code yet, but the functionality works well!

audiodude:
image

I have an idea for a better interface for this, let's chat soon.

audiodude:
> I have an idea for a better interface for this, let's chat soon.

Here's what we discussed:

image

ziaddevv:
@audiodude  any updates?

audiodude:
Looking good!

image

I will start looking at the code now.

wp1-frontend/src/components/BaseBuilder.vue
```diff
@@ -349,32 +522,121 @@ export default {
...
+          "Could not delete this list. Check that the list still exists and you're logged in as its owner.";
+        return;
+      }
+      if (!response.ok) {
+        this.deleteSuccess = false;
+        this.errors = 'Could not check delete impact. Please try again.';
+        return;
+      }
+
+      this.deleteImpact = await response.json();
```
audiodude: You should await the `.json()` in a try catch as well, since the response could not be valid JSON and the page would just break.

wp1/web/builders.py
```diff
@@ -135,18 +135,50 @@ def get_builder(builder_id):
...
+            user_id,
+            builder_id,
+            delete_combinator_ids=delete_combinator_ids,
+            confirm_builder_name=confirm_builder_name,
+        )
     except UserNotAuthorizedError as e:
         flask.abort(403)
     except ObjectNotFoundError:
         flask.abort(404)
+    except ValueError as e:
```
audiodude: Where does the `ValueError` come from? Could we replace with a more specific application error?

audiodude (changes requested):
Looking good, just a few questions

wp1/web/builders.py
ziaddevv: It comes from this  `wp1/logic/builder.py` in `delete_builder()`


```
if confirm_builder_name is not None and confirm_builder_name != builder.name:
    raise ValueError("Builder name confirmation did not match")
```
    
I agree that is a little broad. I’ll replace it with a specific one instead

wp1-frontend/src/components/BaseBuilder.vue
ziaddevv: thanks , good catch

ziaddevv:
@audiodude  just a ping , I guess it is ready to merge

audiodude (approved):

PR #1205, opened 2026-06-29
Auto update combinator

ziaddevv:
**Problem**

Updating a source list did not automatically refresh Combinators that depend on it, so those Combinators could keep showing selections based on an older version of the referenced list.

 **Solution**

- After a list finishes materializing, find Combinators that reference it
- Enqueue rebuild jobs for those Combinators and mark each successfully enqueued one as pending
- If rebuilding a dependent Combinator fails, log the error but do not fail the original list update

### DIFF
--- a/wp1/logic/builder.py
+++ b/wp1/logic/builder.py
@@ -310,9 +310,27 @@ def _update_builder_params(
     return rowcount > 0
 
 
-def _enqueue_combinator_materializations(
-    redis: Redis, combinators: list[Builder]
-) -> list[str]:
+def _mark_combinator_pending_rebuild(wp10db: Connection, combinator: Builder) -> bool:
+    # The list UI treats builders newer than their latest selection as pending.
+    combinator = attr.evolve(combinator)
+    combinator.set_updated_at_now()
+    if combinator.b_id is None:
+        raise ValueError("Cannot update combinator without b_id")
+
+    with wp10db.cursor() as cursor:
+        cursor.execute(
+            """UPDATE builders
+           SET b_updated_at = %s
+           WHERE b_id = %s AND b_user_id = %s
+        """,
+            (combinator.b_updated_at, combinator.b_id, combinator.b_user_id),
+        )
+        rowcount = cursor.rowcount
+    wp10db.commit()
+    return rowcount > 0
+
+
+def _enqueue_combinator_rebuilds(redis: Redis, combinators: list[Builder]) -> list[str]:
     combinator_cls = None
     enqueued_combinator_ids = []
     for combinator in combinators:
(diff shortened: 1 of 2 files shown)


ziaddevv:
note : this branch is stacked on top of the delete-builder PR, so the diff may look a bit mixed. The intended change here is only the automatic rebuild of dependent Combinators after a referenced list is updated.

wp1/logic/builder.py
```diff
@@ -549,6 +583,14 @@ def materialize_builder(
             # version was updated, because that indicates that the ZIM file
             # was never requested or errored and should remain in that state.
             auto_handle_zim_generation(redis, wp10db, builder.b_id)
+        if content_type == "text/tab-separated-values" and not is_meta_builder(builder):
+            try:
+                _rebuild_referencing_combinators(redis, wp10db, builder)
+            except Exception:
```
audiodude: It's never really a good idea to except `Exception` because it can all kinds of logic errors and the like (eg `KeyError` or `ValueError`).

You need to figure out the actual exceptions that could be raised, then what to do in each situation.

wp1/logic/builder.py
```diff
@@ -305,9 +305,27 @@ def _update_builder_params(
     return rowcount > 0
 
 
-def _enqueue_combinator_materializations(
-    redis: Redis, combinators: list[Builder]
-) -> list[str]:
+def _mark_combinator_pending_rebuild(wp10db: Connection, combinator: Builder) -> bool:
```
audiodude: I would just inline this below, without updating the builder timestamp and just setting it directly in the SQL. This is a 15 line function that updates a single value in a single row, and it makes the function that calls it harder to read because the instinct is "Oh, `_mark_combinator_pending_rebuild`, that must be a complex operation".

wp1/logic/builder.py
```diff
@@ -321,6 +339,24 @@ def _enqueue_combinator_materializations(
...
+        wp10db, builder.user_id, builder.id
+    )
+
+    enqueued_combinator_ids = []
+    for record in referencing_records:
+        combinator = record["builder"]
+        enqueued_combinator_ids.extend(
+            _enqueue_combinator_rebuilds(redis, [combinator])
+        )
+        _mark_combinator_pending_rebuild(wp10db, combinator)
```
audiodude:
```
# Update the combinator udpated_at so that the web UI displays it correctly.
        cursor.execute(
            """UPDATE builders
           SET b_updated_at = %s
           WHERE b_id = %s AND b_user_id = %s
        """,
            (##utils.wp10_timestamp_now()##, combinator.b_id, combinator.b_user_id),
        )
        rowcount = cursor.rowcount
```

Obviously getting a WP10 compatible "now" timestamp is something that, I think, is in the codebase somewhere?

And then at the end a `wp10db.commit()`

audiodude (changes requested):
Simple enough, but I think it could be even simpler!

wp1/logic/builder.py
audiodude: Hmm maybe not. It should be:

```
time.strftime(TS_FORMAT_WP10, time.gmtime()).encode("utf-8")
```

wp1/logic/builder.py
ziaddevv: I'm thinking it should be a redisError

ziaddevv:
@audiodude  now it is much simpler :smile:

audiodude (approved):

PR #1209, opened 2026-07-04
Improve Combinator dependency error handling

### LINKED ISSUE #1184: handle multiple bad referenced selections in combinator build
when a combinator references multiple builders and more than one has a bad selection status, the backend currently stops at the first error it encounters. This means a retryable reference can hide a later failed reference depending on builder order. 

so I  guess we have to collect all referenced selection errors or making fatal failures take priority over retryable ones + showing an actionable message for each referenced builder

ziaddevv:
**Problem**
- Combinator builds only surfaced the first failed referenced list, so users had to fix errors one at a time and rerun the Combinator to discover the next failure.

**Summary**
- Added structured dependency error metadata for Combinator referenced-list failures.
- Collected all failed referenced lists instead of stopping at the first failure.
- Added clear user-facing reasons and actions for each dependency failure.
- Linked each failed referenced list from the Combinator error message.

fixes #1184

### DIFF
--- a/wp1/exceptions.py
+++ b/wp1/exceptions.py
@@ -3,7 +3,9 @@ class Wp1Error(Exception):
 
 
 class Wp1SelectionError(Wp1Error):
-    pass
+    def __init__(self, *args, extra=None):
+        super().__init__(*args)
+        self.extra = extra or {}
 
 
 class Wp1RetryableSelectionError(Wp1SelectionError):
--- a/wp1/exceptions.py
+++ b/wp1/exceptions.py
@@ -1,9 +1,12 @@
+from typing import Any
+
+
 class Wp1Error(Exception):
     pass
 
 
 class Wp1SelectionError(Wp1Error):
-    def __init__(self, *args, extra=None):
+    def __init__(self, *args: Any, extra: dict[str, Any] | None = None):
         super().__init__(*args)
         self.extra = extra or {}
 
@@ -14,6 +17,74 @@ class Wp1RetryableSelectionError(Wp1SelectionError):
 
 class Wp1FatalSelectionError(Wp1SelectionError):
     pass
+
+
+class Wp1MetaSelectionError(Wp1SelectionError):
+    def __init__(
+        self,
+        message: str,
+        *,
+        code: str,
+        reason: str,
+        action: str,
+        **details: str | None,
+    ):
+        super().__init__(message)
+        self.code = code
+        self.reason = reason
+        self.action = action
+        self.details = {
+            key: value for key, value in details.items() if value is not None
+        }
+
+
+class Wp1RetryableMetaSelectionError(Wp1MetaSelectionError, Wp1RetryableSelectionError):
+    pass
+
+
+class Wp1FatalMetaSelectionError(Wp1MetaSelectionError, Wp1FatalSelectionError):
+    pass
+
+
+class Wp1MetaBuilderProcessError(Wp1SelectionError):
+    def __init__(self, failures: list[tuple[dict[str, Any], Wp1SelectionError]]):
+        super().__init__("One or more referenced selections failed")
+        self.failures = failures
+
+    @property
+    def has_fatal_errors(self) -> bool:
+        return any(
+            isinstance(error, Wp1FatalSelectionError)
+            for _reference, error in self.failures
+        )
+
+    def to_user_messages(self) -> list[dict[str, Any]]:
+        messages = []
+        for reference, error in self.failures:
+            message = str(error)
+            data = {
+                "builder_id": reference["id"],
+                "builder_name": reference["name"],
+                "builder_model": reference["model"],
+                "message": message,
+                "reason": message,
+                "status": (
+                    "FAILED"
+                    if isinstance(error, Wp1FatalSelectionError)
+                    else "CAN_RETRY"
+                ),
+            }
+            if isinstance(error, Wp1MetaSelectionError):
+                data.update(
+                    {
+                        "code": error.code,
+                        "reason": error.reason,
+                        "action": error.action,
+                    }
+                )
+                data.update(error.details)
+            messages.append(data)
+        return messages
 
 
 class ZimFarmError(Wp1Error):
--- a/wp1/selection/meta_builder.py
+++ b/wp1/selection/meta_builder.py
@@ -11,6 +11,19 @@ from wp1.exceptions import (
 from wp1.selection.abstract_builder import AbstractBuilder
 
 
+# Most dependdency errors only need code/reason/action, details are optional debug fields
+def _dependency_error_extra(
+    code: str, reason: str, action: str, **details: str
+) -> dict[str, str]:
+    extra = {
+        "dependency_code": code,
+        "dependency_reason": reason,
+        "dependency_action": action,
+    }
+    extra.update({key: value for key, value in details.items() if value is not None})
+    return extra
+
+
 class MetaBuilder(AbstractBuilder):
     """Base class for builders that reference other builders."""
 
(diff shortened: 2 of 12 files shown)


audiodude:
I'm not sure I understand. Shouldn't this be patched on top of #1204 ?

ziaddevv:
> I'm not sure I understand. Shouldn't this be patched on top of #1204 ?


I doesn't seem to be a hard conflict with #1204. I'd prefer not to rebase unless you think stacking it would make review clearer. Do you want me to rebase it , or keep it based on main for now?

audiodude:
Sorry I see that they're independent now.

wp1/selection/models/combinator.py
```diff
@@ -169,30 +171,104 @@ def validate(
...
+        retryable_errors: list[dict[str, Any]],
+        fatal_errors: list[dict[str, Any]],
+    ) -> set[str]:
 
         builder_values = group.get("builders", [])
 
-        builder_ids = [
-            builder_id for builder_id in builder_values if isinstance(builder_id, str)
-        ]
+        builder_ids = []
```
audiodude: I think you lost the de-dupe. This should be a set?

wp1/selection/models/combinator.py
```diff
@@ -169,30 +171,104 @@ def validate(
...
-                wp10db,
-                s3,
-                builder_id,
-                logic_builder.builder_label_by_id(wp10db, builder_id),
-            )
+        for builder_id in builder_ids:
+            reference: dict[str, Any]
+            try:
+                builder = logic_builder.get_builder(wp10db, builder_id)
+                reference = {
```
audiodude: nit: Could probably refactor to build this in a more concise way.

wp1/selection/models/combinator.py
```diff
@@ -169,30 +171,104 @@ def validate(
...
+                continue
+            except Wp1RetryableSelectionError as e:
+                retryable_errors.append(
+                    self._referenced_builder_error(reference, e, "CAN_RETRY")
+                )
+                continue
             title_set = _parse_tsv_to_set(data)
             sets.append(title_set)
 
         return _apply_operation(operation, sets)
```
audiodude: If there are errors, we just proceed with the operation with empty lists for the impacted Builders?

wp1/selection/models/combinator.py
```diff
@@ -169,30 +171,104 @@ def validate(
 
         return ([], [], [])
 
-    def _process_group(self, wp10db, s3, group: dict[str, Any]) -> set[str]:
+    def _process_group(
+        self,
+        wp10db,
+        s3,
+        group: dict[str, Any],
+        retryable_errors: list[dict[str, Any]],
```
audiodude: I don't think this is a good pattern, to mutate a list of errors. It is very difficult to reason about and follow. I will comment on a different approach in my overall comment.

wp1/selection/models/combinator.py
```diff
@@ -169,30 +171,104 @@ def validate(
...
+    def _referenced_builder_error(
+        self, reference: dict[str, Any], error: Wp1SelectionError, status: str
+    ) -> dict[str, Any]:
+        message = str(error)
+        extra = error.extra if isinstance(error.extra, dict) else {}
+        reason = extra.get("dependency_reason")
+        if not reason:
+            label = reference["label"] or reference["id"] or ""
+            prefix = f"Referenced builder {label} "
+            if message.startswith(prefix):
```
audiodude: This is very clunky and brittle. From what I can tell, you're string matching against a string that you control, in order to branch logic and concatentate a user friendly string?

Prefer a specific ErrorType enum, or something similar. We could also create new types of `Wp1SelectionError`, like `Wp1MetaBuilderRetryableError` or things of that nature, and they could contain `user_message()` methods.

audiodude (changes requested):
I don't think this approach is quite right. I understand that you want to reuse the `_process_group` method across the include and exclude set, but passing a mutable list is not a good pattern IMO. And it's not just because it's difficult to reason about the list's value at any given point in the code. When you catch an Exception and "convert" it into a string (by storing a string for each exception), you are losing information, as you see in my comment below.

This is fine, this is what we usually do when there's only one potential error. However in this case, there is a matrix of errors. I would recommend saving/storing/passing the actual exception objects. They're just normal objects. Then construct your user messages at the end. Not only does this simplify your error message construction, it preserves information.

If someday, the Wp1RetryableError has a subclass of like Wp1RetryableRateLimitError, it might have information on the number of seconds to wait before retrying. Your current approach discards that potential information.

Happy to discuss more on Tuesday!

wp1/selection/models/combinator.py
```diff
@@ -212,13 +288,33 @@ def build(self, content_type: str, **params: Any) -> bytes:
 
         exclude_group = params.get("exclude")
 
-        include_set = self._process_group(wp10db, s3, include_group)
+        retryable_errors: list[dict[str, Any]] = []
+        fatal_errors: list[dict[str, Any]] = []
+
+        include_set = self._process_group(
+            wp10db, s3, include_group, retryable_errors, fatal_errors
```
audiodude: It is hard to reason about a list that is mutated inside of the other method.

If you really want to do it this way, you should do:

```
include_set, retryable, fatal = self._process_group(...)
```

wp1/selection/models/combinator.py
```diff
@@ -169,30 +171,104 @@ def validate(
...
+                    "model": None,
+                    "label": builder_id,
+                }
+
+            try:
+                data = self._fetch_selection_data(
+                    wp10db, s3, builder_id, reference["label"]
+                )
+            except Wp1FatalSelectionError as e:
+                fatal_errors.append(
```
audiodude: Instead of adding a new data structure that you create (`_referenced_builder_error`), just add the exception itself to the list.

wp1/selection/models/combinator.py
```diff
@@ -212,13 +288,33 @@ def build(self, content_type: str, **params: Any) -> bytes:
 
         exclude_group = params.get("exclude")
 
-        include_set = self._process_group(wp10db, s3, include_group)
+        retryable_errors: list[dict[str, Any]] = []
+        fatal_errors: list[dict[str, Any]] = []
+
+        include_set = self._process_group(
```
audiodude: Maybe we should use only one `_process_groups()` method, that takes include and exclude, so that we can collect all errors at once.

```
try:
   include_set, exclude_set = self._process_groups(wp10db, s3, include_group, exclude_group)
except Wp1MetaBuilderProcessError:
    ...
```

wp1/selection/models/combinator.py
```diff
@@ -212,13 +288,33 @@ def build(self, content_type: str, **params: Any) -> bytes:
...
             exclude_builders = exclude_group.get("builders", [])
             if exclude_builders:
-                exclude_set = self._process_group(wp10db, s3, exclude_group)
+                exclude_set = self._process_group(
+                    wp10db, s3, exclude_group, retryable_errors, fatal_errors
+                )
+
+        if retryable_errors or fatal_errors:
+            referenced_builder_errors = fatal_errors + retryable_errors
+            message = (
```
audiodude, line 270: Once you have the new Error class, you can have a `.to_user_message()` method that does something similar to this.

wp1/selection/models/combinator.py
```diff
@@ -212,13 +288,33 @@ def build(self, content_type: str, **params: Any) -> bytes:
...
+                "Could not build Combinator because referenced selections failed: "
+                + "; ".join(
+                    error["message"] or "" for error in referenced_builder_errors
+                )
+            )
+            extra = {"referenced_builder_errors": referenced_builder_errors}
+            if fatal_errors:
+                raise Wp1FatalSelectionError(message, extra=extra)
```
audiodude: This part is good, but you should try to wrap the original MetaBuilderProcessError

You can do a `throw Wp1FaltaSelectionError(message) from e`

https://docs.python.org/3/tutorial/errors.html#exception-chaining

wp1/selection/models/combinator.py
```diff
@@ -212,13 +288,33 @@ def build(self, content_type: str, **params: Any) -> bytes:
...
         exclude_set: set[str] = set()
         if exclude_group is not None:
             exclude_builders = exclude_group.get("builders", [])
             if exclude_builders:
-                exclude_set = self._process_group(wp10db, s3, exclude_group)
+                exclude_set = self._process_group(
+                    wp10db, s3, exclude_group, retryable_errors, fatal_errors
+                )
+
+        if retryable_errors or fatal_errors:
```
audiodude: This should now be the `except`

wp1/selection/meta_builder.py
```diff
@@ -11,6 +11,19 @@
 from wp1.selection.abstract_builder import AbstractBuilder
 
 
+# Most dependdency errors only need code/reason/action, details are optional debug fields
+def _dependency_error_extra(
+    code: str, reason: str, action: str, **details: str
+) -> dict[str, str]:
+    extra = {
+        "dependency_code": code,
```
audiodude: Maybe create a `Wp1RetryableMetaSelectionError` that subclasses the original, and promote these fields to first class properties on that object.

audiodude (changes requested):

ziaddevv:
Hi @audiodude , I just finished refactor based on your feedback. Could you take another look when you have a chance?

wp1/selection/models/combinator.py
ziaddevv: Just like before I  kept the list of errors order-preserving so the output is deterministic and matches the user’s configured builder order. If we use a set, the order can change between runs, which makes the UI/error message less predictable and makes tests inconsistent.

wp1/selection/models/combinator.py
```diff
@@ -176,98 +177,70 @@ def _process_group(
...
-                reference = {
-                    "id": builder.id,
-                    "name": builder.name,
-                    "model": builder.model,
-                    "label": builder.label,
-                }
             except ObjectNotFoundError:
                 reference = {
                     "id": builder_id,
                     "name": builder_id,
```
audiodude: If there is really no builder, we should just leave these `None` instead of falsely populating them with the id.

wp1/selection/models/combinator.py
```diff
@@ -286,35 +259,22 @@ def build(self, content_type: str, **params: Any) -> bytes:
...
-                )
-
-        if retryable_errors or fatal_errors:
-            referenced_builder_errors = fatal_errors + retryable_errors
+        try:
+            include_set, exclude_set = self._process_groups(
+                wp10db, s3, include_group, exclude_group
+            )
+        except Wp1MetaBuilderProcessError as e:
+            referenced_builder_errors = e.to_user_messages()
```
audiodude, line 269: You get all the user messages, but then you don't use them on line 272, is that intentional?

wp1/exceptions.py
```diff
@@ -16,6 +19,74 @@ class Wp1FatalSelectionError(Wp1SelectionError):
...
 
+class Wp1MetaSelectionError(Wp1SelectionError):
+    def __init__(
+        self,
+        message: str,
+        *,
+        code: str,
+        reason: str,
+        action: str,
+        **details: str,
```
audiodude: How can this be `str`? Doesn't it have to be `dict` to capture the remaning keyword args?

audiodude (changes requested):
Just a few remaining comments. Great work implementing the stuff we talked about!

wp1/selection/models/combinator.py
ziaddevv: they are the same , but I  agree using the user messages  is much consistent

wp1/exceptions.py
ziaddevv: to annotate the  kwargs object we specify the value itself not the whole `dict`
check here 
https://typing.python.org/en/latest/spec/callables.html#annotating-args-and-kwargs

audiodude: Oh interesting! TIL.

audiodude (approved):
LGTM, great work!

Issue #1207, opened 2026-06-30
Combinator: Don't allow the same Builder in both positive and negative lists


audiodude:
<img width="1093" height="1238" alt="Image" src="https://github.com/user-attachments/assets/a4403c58-8b26-45e6-b0eb-975910aff876" />

PR #1216, opened 2026-07-17
improve Combinator builder selection UI

### LINKED ISSUE #1206: Combinator: Don't display "no builders" message if I've already added some
Image

ziaddevv:
solves some error messages 
+
fixes #1206 #1207 #1208

### DIFF
--- a/wp1-frontend/src/components/CombinatorBuilder.vue
+++ b/wp1-frontend/src/components/CombinatorBuilder.vue
@@ -51,6 +51,9 @@
                   v-model="includeDropdown"
                   class="custom-select my-list"
                   :class="{ 'is-invalid': includeValidationMessage }"
+                  :disabled="
+                    availableIncludeBuilders(builder.project).length === 0
+                  "
                 >
                   <option value="">Select a builder to add</option>
                   <option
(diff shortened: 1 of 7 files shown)


wp1-frontend/src/components/CombinatorBuilder.vue
```diff
@@ -371,23 +379,24 @@ export default {
...
-      this.setSelectValidity('includeSelect', this.includeValidationMessage);
     },
-    // TODO : just refactor into one function
     availableIncludeBuilders: function (project) {
-      return this.availableBuildersFor(project, this.includeBuilders);
+      return this.availableBuildersFor(project, this.selectedBuilderIds());
     },
     availableExcludeBuilders: function (project) {
-      return this.availableBuildersFor(project, this.excludeBuilders);
+      return this.availableBuildersFor(project, this.selectedBuilderIds());
```
audiodude: These functionas are now identical, which indicates a design flaw of some sort. Or possibly a logic error? What code is calling these that expects separate *IncludeBuilders and *ExcludeBuilders, but both lists are the same? I would refactor that code.

audiodude (changes requested):
Just one question, looks good otherwise. Good idea to bundle these fixes together.

audiodude (approved):

PR #1271, opened 2026-08-23
Add dbname column to builders, cached from the sitematrix

### LINKED ISSUE #1268: Update the Builder data model so that it contains dbname alongside project name
Currently the project name is the canonical column that defines whether a Selection is for English Wikipedia or Spanish Wiktionary. We should also add the dbname, cached from the same sitemap table.

audiodude:
Fixes #1268

Adds `b_dbname` to the `builders` table, holding the database name (`enwiki`) of the wiki `b_project` refers to — real data from the meta.wikimedia.org sitematrix, not derived from the domain.

- New `wp1/logic/sites.py` fetches the sitematrix and caches the project list + domain→dbname map in redis for a day. `/v1/sites/` now delegates to it, which also fixes a latent bug where the endpoint wrote its cache under `sites_data` but read `sites`, so the cache never hit.
- `b_dbname` is resolved and persisted at the start of every materialization, so existing builders backfill on their next materialize and project edits are picked up. Resolution failures are logged and skipped — materialization never blocks on the sitematrix.
- Migration `20260823_01_add_dbname_to_builders` (nullable VARBINARY(255)); test schema updated.

Smoke-tested against the live sitematrix: 948 projects / 1072 dbnames parsed; `en.wikipedia.org→enwiki`, `commons.wikimedia.org→commonswiki`, `be-tarask.wikipedia.org→be_x_oldwiki`. Full backend suite passes (847).

#1267 will be rebased on top of this to use `b_dbname` for the served TSV filename instead of a heuristic.

Written with AI assistance (Claude).

### DIFF
--- a/wp1/logic/builder.py
+++ b/wp1/logic/builder.py
@@ -556,7 +556,7 @@ def _refresh_builder_dbname(redis: Redis, wp10db: Connection, builder: Builder) 
         dbname = logic_sites.dbname_for_project(
             redis, builder.b_project.decode("utf-8")
         )
-    except Exception:
+    except logic_sites.FETCH_ERRORS:
         logger.exception("Could not resolve dbname for project=%s", builder.b_project)
         return
     if dbname is None:
--- a/wp1/logic/builder.py
+++ b/wp1/logic/builder.py
@@ -13,6 +13,7 @@ from pymysql.connections import Connection
 from redis import Redis
 from redis.exceptions import RedisError
 import wp1.logic.selection as logic_selection
+import wp1.logic.sites as logic_sites
 import wp1.logic.util as logic_util
 import wp1.logic.zim_files as logic_zim_tasks
 import wp1.logic.zim_schedules as logic_zim_schedules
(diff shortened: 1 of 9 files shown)


wp1/logic/builder.py
```diff
@@ -544,6 +545,35 @@ def get_builder(wp10db: Connection, id_: str | bytes) -> Builder:
...
+
+    The dbname (eg 'enwiki') is looked up in the redis-cached sitematrix. Any
+    failure to resolve it is logged and ignored, so that materialization can
+    proceed without it.
+    """
+    try:
+        dbname = logic_sites.dbname_for_project(
+            redis, builder.b_project.decode("utf-8")
+        )
+    except Exception:
```
audiodude: Prefer specific exceptions to `except Exception`

audiodude: Done in a4f30d7 — now catches `FETCH_ERRORS` exported from `wp1.logic.sites` (`RedisError`, `requests.exceptions.RequestException`, `mwclient.errors.MwClientError`), so the guard covers exactly the service failures the sitematrix fetch can raise, and a shape/code bug would surface instead of being swallowed.

PR #1285, opened 2026-08-27
use different payloads for recipe create and update

elfkuzco:
## Rationale
zimfarm uses different payloads for recipe creation and update but currently wp1 uses same payload for both operations. this fails silently because api schema models allow extra fields and pydantic proceeds with only with what it needs. as a consequence, recipes aren't exactly getting updated in the way wp1 expects. This PR attempts to rectify it by using two different functions for creating the payloads for each operation.

https://github.com/openzim/zimfarm/blob/41e6fa8eae723d2dfcbdfe3f49c6c0a34d4ee272/backend/src/zimfarm_backend/api/routes/recipes/models.py#L36-L46

https://github.com/openzim/zimfarm/blob/41e6fa8eae723d2dfcbdfe3f49c6c0a34d4ee272/backend/src/zimfarm_backend/api/routes/recipes/models.py#L53-L70

### DIFF
--- a/wp1/zimfarm.py
+++ b/wp1/zimfarm.py
@@ -298,29 +298,26 @@ def get_zim_filename_prefix(builder: Builder, selection: Selection) -> str:
     return f"{util.safe_name(builder_name)}-{selection_id_frag}"
 
 
-def _get_params(
-    builder: Builder,
-    selection: Selection,
-    title: str,
-    description: str,
-    long_description: str,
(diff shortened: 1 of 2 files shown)


benoit74 (approved):
LGTM, glad you started looking at WP1 codebase in more details ^^

audiodude:
LGTM thank you!

PR #1317, opened 2026-09-07
Create SECURITY.md for security policy

angelnava92:
Added a security policy document outlining supported versions and vulnerability reporting.


angelnava92:
Bug

PR #1319, opened 2026-09-08
Upgrade recipe definition for mwoffliner 2.0.0

benoit74:
mwoffliner 2.0.0 has been released today and this is a major with breaking "API" changes.

We hence need to update the recipe definition with this PR.

We also need to update prod wp1 configuration: it probably overrides the scraper version and offliner definition version set in this PR

We might want to update all prod recipes configs. I have tooling ready for this, but I don't think this is necessary since WP1 update the recipe each time it requests it again, correct?


audiodude:
This doesn't work, explained in: https://github.com/openzim/wp1#configuration-env

Basically the schema for env keys is defined in https://github.com/openzim/wp1/blob/main/wp1/config.py

And then the .env.example is just generated from that, it's not a static file that we just arbitrarily update (such as in this PR). I know that's somewhat annoying, but it means that the .env file can never get out of date with the actual config machinery.

The updates should be made here: https://github.com/openzim/wp1/blob/main/wp1/config.py#L359

Then the .env.example should be regenerated as explained in the CI failure:

```
Diff is 9307 characters long. Set self.maxDiff to None to see it. : .env.example is out of date. Run: pipenv run python -m wp1.config
```

audiodude (changes requested):
See comment.

audiodude (approved):

audiodude:
@benoit74 is this ready for deploy?

benoit74:
@audiodude Yes, I think you can deploy ; beware to also update prod env variables when deploying, otherwise it's going to fail.

Do you agree we do not need to update existing recipes, the WP1 logic will update them automatically next time they are scheduled?

audiodude:
That is correct for single ZIMs.

If we have any "scheduled" ZIMs laying around, those would need to be manual updated since they reuse a schedule.

EDIT: Probably best done as a yoyo migration.

benoit74:
> If we have any "scheduled" ZIMs laying around, those would need to be manual updated since they reuse a schedule.

Don't these scheduled ZIMs always update the recipe configuration before starting it again? I feel like it would be a safe choice to fix this **and** to prevent "surprises" where Zimfarm recipe configuration is not aligned with what WP1 would have expected for whatever reason

audiodude:
Oh you're right, they're scheduled on the WP1 side but use the same "Create a one time recipe" pattern as the rest of WP1, so everything should be fine. :+1:

audiodude:
Deploying now.

EDIT: Deployed!

benoit74:
I confirm recipe update worked just fine: https://farm.zimit.kiwix.org/recipes/wp1_selection_bccbf75302cf7593/history?compare=22f7f8d3-72da-490c-bdd7-6a271d7bb970...0b1b2496-1e1e-4a18-bb80-94a9b05d3fd6

Edit: to clarify, I just requested again an existing selection in WP1 and WP1 performed this update automatically on its own

benoit74:
A my bad most of you probably do not have sufficient rights on the Zimfarm to see this URL content:

image

PR #1329, opened 2026-09-15
wp1 selection tools

elfkuzco:
## Rationale
This PR brings the [wp1 selection tools](https://github.com/openzim/wp1_selection_tools) written in Perl and Bash to WP1 but purely in Python. The port is intended to be output-compatible with the Perl+Bash implementation

It doesn't port the individual scripts of the Perl implementations as they are named but where possible, functions to do the same thing have been created in a slightly different organization.

## Replacement map

| Perl / Bash | Python |
|---|---|
| `merge_lists.pl` | `fetch_page_scores_and_ratings_from_db` |
| `build_scores.pl` | `scores.py` |
| `build_projects_lists.pl` (English) | `wiki/en.py::build_projects_list` |
| `build_translated_list.pl` | `projects_list.py` |
| `compare_lists.pl` (`only1`) | `wiki/endless.py` |
| `listCategoryEntries.pl` + `Mediawiki.pm::listCategoryEntries` | `wiki/__init__.py::generate_category_entries` |
| `listLangLinks.pl` + `Mediawiki.pm::langLinks` | `wiki/__init__.py::generate_langlinks` |
| `build_en_vital_articles_list.sh` | `wiki/en.py::get_vital_articles` |
| `build_biggest_wikipedia_list.sh` | `entrypoint.py::get_wiki_languages_exceeding_count` |
| `build_custom_selections.sh` (en / fr / endless) | `wiki/en.py`, `wiki/fr.py`, `wiki/endless.py` |
| `build_selections.sh` (gather) | `entrypoint.py::build_selections` |
| `build_selections.sh` (7z / scp / cleanup) | `entrypoint.py::_compress_all`, `_upload_dir` |
| `build_all_selections.sh` | `entrypoint.py::build_all_selections` (+ `build-all`) |
| `customs/` | `selection_tools/data/`

The program can be invoked in the following ways:
```sh
python -m wp1.selection_tools.entrypoint list 100000
python -m wp1.selection_tools.entrypoint build en --data-dir /data
python -m wp1.selection_tools.entrypoint build-all --min-articles 100000 --data-dir /data
```
- `list <min_articles>`:  Wikipedias above a page count, `lang count` per line.
- `build <lang>`: one language, start to finish (gather → merge → scores → tops → projects → customs → compress → upload).
- `build-all`:  every Wikipedia, with up to four retries per language.

These three sub-parsers correspond to the original `build_biggest_wikipedia_list.sh`, `build_selections.sh` and `build_all_selections.sh` scripts


elfkuzco:
@audiodude , here's a draft of the migration of the wp1-selection-tools but I need a little more clarification on some of the things said in the Design Docs. 

The previous Perl implementation downloaded the pagecounts data from https://dumps.wikimedia.org/other/pagecounts-ez/merged/ before processing those entries and generating the `pageviews.tsv` and `pages.tsv`. However, in the docs, you mentioned `pages.tsv` source should be a render of `page_scores` and `pageviews.tsv` should be a render of `ps_views`. However, looking at this https://github.com/openzim/wp1/blob/14b6b5fc316dc067787e77b0c6c5c49493bd2f9d/wp1/scores.py#L189-L195, it appears to me that `pageviews.tsv` maps directly to the data (just page title and count) on `page_scores` table and not `pages.tsv`. Am I wrong in this deduction? 

Also, you mentioned `pageviews.tsv` (which I think should be `pages.tsv`) has it's source as a render of `ps_views` but I cannot find any table on the codebase named `ps_views`. In the notes section of this particular row, you mentioned "fresh data (see validation & acceptance)" but I can't find any such sections. Please, could you provide additional information on what these mean?

Similar thing applies to `scores.tsv` whose source is `ps_score`. Are these tables on the wikilang db instead of wp10?

audiodude:
@elfkuzco I'm not sure I understand. The current implementation only provides the schema. Your task is to populate the columns.

Is there some aspect of the .tsv files that I'm missing where these columns are not enough to materialize them?

elfkuzco:
> @elfkuzco I'm not sure I understand. The current implementation only provides the schema. Your task is to populate the columns.
> 
> Is there some aspect of the .tsv files that I'm missing where these columns are not enough to materialize them?

Okay. I think my confusion arises largely from the design docs. I will outline my understanding of specific paragraphs so you can see where my dilemma arises from.
- The initial paragraph says
> The wp1 selection tools produce every published artifact at [wp1.download.openzim.org](http://wp1.download.openzim.org/). They are written in Perl scripts and utilize flat files for data storage. They have worked well for some time, but are difficult to understand and maintain. Additionally, they rely on stale pageview data from 2020 (https://dumps.wikimedia.org/other/pagecounts-ez/)

Originally, the first commit of this draft downloaded these files and did everything 1:1 but I realized that this file https://github.com/openzim/wp1/blob/main/wp1/scores.py seems to be downloading similar files and performing similar thing as the old perl version but this time, it was writing it to a table instead of a file. Plus, it's URL seems different which made me to conclude this might be the new source of truth since the initial paragraph mentioned that they URL was showing "stale pageview data"
 https://github.com/openzim/wp1/blob/87cb8edc770dd45cb3a4cdc490fe29699bd443f6/wp1/scores.py#L41-L52
I now arrived at the conclusion that I shouldn't have replicated everything 1:1 and should instead rely on the existing code as the source of truth for the current month's `pageviews.tsv`. I might be wrong in this deduction.

This new conclusion was further strengthened by the table and it's columns (**Source** and **Notes**)
Screenshot_20260918_145816

Because the existing code has this table `page_scores`, I assumed it must have `ps_views` too but I didn't see such. Also, the notes don't offer helpful information on what it's comments mean. Where is the documentation that is referred to in the notes? What does it mean when it says "See Validation and acceptance"

Perhaps, I can revert to the original commit where everything is 1:1 and we can now work on these notes and what they mean.

> @elfkuzco I'm not sure I understand. The current implementation only provides the schema. Your task is to populate the columns.

Which columns should I populate? The screenshot mentioned `render of ps_views`. But when I do a grep of the codebase, there is no table `ps_views`. Would it be better to have a meeting on Slack instead so we can go over these confusions in a more synchronous manner?

audiodude:
Yes, the goal is to store the equivalent wp1_selection_tools information/data in the WP1 project database (wp10db).

The pageviews are already implemented in scores.py, it pulls the proper file and streams it to calculate the scores. The original idea I had was to calculate everything else as well and put it in the `pages_scores` table. This is not strictly necessary for generating the legacy [download](https://wp1.download.openzim.org/) files, but would be an enhancement so that users could create WP1 selections/ZIMs of their own based on this data.

The table you screenshotted is an explicity mapping from "WP1 Source", even though not all of the data is available in the current code.

You are free to change the schema if necessary, if I missed something or something is somewhat wrong. Please do so with a yoyo migration.

elfkuzco:
> Yes, the goal is to store the equivalent wp1_selection_tools information/data in the WP1 project database (wp10db).
> 
> The pageviews are already implemented in scores.py, it pulls the proper file and streams it to calculate the scores. The original idea I had was to calculate everything else as well and put it in the `pages_scores` table. This is not strictly necessary for generating the legacy [download](https://wp1.download.openzim.org/) files, but would be an enhancement so that users could create WP1 selections/ZIMs of their own based on this data.
> 
> The table you screenshotted is an explicity mapping from "WP1 Source", even though not all of the data is available in the current code.
> 
> You are free to change the schema if necessary, if I missed something or something is somewhat wrong. Please do so with a yoyo migration.

Okay, so, I added new columns for size, links, lang_links and score to the temp_pageviews and page_scores tables because if we want to generate `all.tsv` using `JOIN` on `page_scores` and `ratings` table, we need those columns. I also created indexes on `page_scores` and `ratings` table. During local testing for language code `en`, I found out that the query to generate `all.tsv` can be slow as there are many millions of rows.

Because of the new columns, the logic to swap out the tables is done after two operations to the `temp_pageviews` table.
- donwload pageviews data and insert lang, page_id, article and views (already existing in scores.py)
- compute metrics after fetching data from wikidb. used to add data for the new columns I mentioned above
Once these two steps are done, we swap out the tables and truncate temp_pageviews

I have run the code locally but I don't think I have bandwidth for it. There are many millions of rows and data from wikidb is in gigabytes of data. I figure the best way is to run the selection_tools and compare it with previous data on a better machine. And then, I resolve differences.

wp1/selection_tools/wiki/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+                        continue
+                    seen_titles.add(article)
+                    yield article
+
+
+def generate_langlinks(
+    lang_code: str, titles: list[str], languages: list[str]
+) -> Generator[LangLink, None, None]:
+    """Yield a LangLink for each language an article exists in."""
+    url = f"https://{lang_code}.wikipedia.org/w/api.php"
```
audiodude: Why are we querying the API to get langlinks? We have access to the wikipedia replica databases, enwiki_p, frwiki_p, eswiki_p, etc.

https://github.com/openzim/wp1_selection_tools/blob/main/build_selections.sh#L207

wp1/selection_tools/constants.py
```diff
@@ -0,0 +1,5 @@
+from pathlib import Path
+
+BASE_DIR = Path(__file__).parent.resolve()
+# where all project custom data are stored
+DATA_DIR = BASE_DIR / "data"
```
audiodude: If you need a DATA_DIR, import it from config. That way we can set it differently in dev versus prod.

Additionally, putting all this data in the source code directory tree would probably exhaust our disk quota. On Wikimedia Cloud VPS, our large data storage is on /srv/

audiodude:
I added a few "drive by" comments, but haven't done a full review of the code yet because I think there are a couple of architectural problems.

First, it is definitely correct to add `ps_size`, which was missing.

However, you should not be copying all of the data from `temp_pageviews` to `ps_scores`. It's called temp_**pageviews** for a reason, it should only be used for copying the `ps_views`. There should be parallel tables for `temp_pagelanglinks`, `temp_pagelinks`, `temp_pagesize`, etc. And they should all follow the pattern of pageviews. Store the data in the temp table while it's collected, then do an atomic swap with the live data.

This is going to be frustrating because the logic will be nearly identical, so it would seem like we should make a general function with parameters. However, this is not possible because in MySQL/MariaDB, you cannot safely use a placeholder for table name (ie `SELECT * FROM {my_table_name}...`). so the only way to do it would be with string concatenation, which we should obviously avoid.

The other main problem I see with this PR, which I commented on in only one place, is that you seem to be using the Wikipedia API to fetch the relevant data? We should 100% be using the replica dbs, like the original wp1_selection_tools does. WP1 runs in the same [Wikimedia Cloud VPS](https://wikitech.wikimedia.org/wiki/Portal:Cloud_VPS) environment as the current tools, on the same box even, and has access to the same resources.

Let me know if you have any questions!

audiodude:
You will probably need toconnect to the proper wikipedia language replica database. Luckily they are all named according to the language code: `enwiki_p`, `arwiki_p`, `frwiki_p`, `eswiki_p`, etc.

You can use the code in `wikilang_db.py` to connect to the proper replica. As mentioned, set up a SOCKS5 proxy in order to reach these from your local dev machine.

elfkuzco:
> The other main problem I see with this PR, which I commented on in only one place, is that you seem to be using the Wikipedia API to fetch the relevant data? We should 100% be using the replica dbs, like the original wp1_selection_tools does. WP1 runs in the same [Wikimedia Cloud VPS](https://wikitech.wikimedia.org/wiki/Portal:Cloud_VPS) environment as the current tools, on the same box even, and has access to the same resources.

The old selection tools uses the replica dB for `pagelinks`, `langlinks`, `redirects`. Asides from that it uses HTTP API to fetch vital articles and build custom selections with this class as the HTTP class https://github.com/openzim/wp1_selection_tools/blob/main/mediawiki/classes/Mediawiki/Mediawiki.pm

The only difference in some areas with the HTTP implementation is that where possible, it uses `format=csv` whereas I use `format=json` when interacting with the Wikipedia API. 

Is there something else I am missing?

elfkuzco:
I've tested it locally by building for language code `aa` since it's [files](https://wp1.download.openzim.org/aawiki_2026-08/) are all relatively small. I do not generate the `langlinks.tsv` and `pagelinks.tsv` as mentioned in the design docs but when i inspect the DB column for the values stored them, they match accordingly with the number of lines in those files.

The major differences when comparing those files are:
## pages.tsv
 The one listed upstream contains two entries
```sh
1269	Main_Page	0	0
1426	Main_page	23	1
```

while mine is:
```sh
1269	Main_Page	0
```
You mentioned to drop redirect rows in the design docs. The fourth column for the second entry upstream is `1` signifying it's a redirect row, hence, it's dropped

## pageviews.tsv

The one upstream is far too many to be listed but here are it's contents
[pageviews.tsv](https://github.com/user-attachments/files/32613357/pageviews.tsv)

whereas mine is just two entries:
```
Main_Page	22
Wikipedia:Sandbox	1
```

I think this discrepancy largely arises from the new data format. The old selection tools reads a completely different file compared to the recent one. 

## all.tsv
My db dump contains no entry for article `Main_Page` in `ratings` table but I suspect the prod db does. That's why I couldn't generate any ratings for mine. This also affects the `scores.tsv` file outputs. Is it possible to get a dump of the ratings table on prod?

